PR #19, opened 2011-08-31
Color

ju1ius:
Enhancements to color handling.
Improved algorithm for hex to rgb conversion.
All colors, including x11 colors are parsed to a Color object and converted to rgb/rgba internally. This is needed in order to make expandShorthands and createShorthand methods work with named (x11) colors.
New methods for color conversion to/from hsl and rgb.


ju1ius:
I guess we can now reasonably says that this won't be merged.

PR #96, opened 2015-09-14
RuleSet::addRule() insert position support.

IceReaper:
I'm currently using this lib to auto-prefix css rules for a wider browser support without the need to manually add all those rules. This requires me to add the new prefixed rules before the unprefixed one. I Made some additions to allow you to specify the exact location where the final Rule should be:

### DIFF
--- a/lib/Sabberworm/CSS/RuleSet/RuleSet.php
+++ b/lib/Sabberworm/CSS/RuleSet/RuleSet.php
@@ -17,12 +17,22 @@ abstract class RuleSet implements Renderable {
 		$this->aRules = array();
 	}
 
-	public function addRule(Rule $oRule) {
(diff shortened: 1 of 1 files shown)


sabberworm:
Cool, thanks.

Could you add a few unit tests for this feature?

Also, I think the API could be simpler: append the rule to the existing rules if `$oSibling` is `null`, prepend before `$oSibling` othewise. This way, any position can be adressed and we don’t need `$bPrepend` at all.

FMCorz:
@IceReaper Do you have that library to auto-prefix CSS lying around? I'd be interested in it.

IceReaper:
@FMCorz Sorry, left all the sources with my old employer, got a new job and have no longer access to them. Sorry :(

PR #97, opened 2015-12-25
fix php notice

sivig:


sabberworm:
I’m not sure what notice this is supposed to prevent. The array key undefined? Shouldn’t that be suppressed by the `@`?

sivig:
Yes, the array key undefined error, for some reason it's not being suppressed.

PR #98, opened 2016-02-11
Increase performance up to 15000%

ossinkine:
The basic idea is to use an array of characters instead of string processing.

**Benchmark**
For test I took the current css-file from github.com: https://assets-cdn.github.com/assets/github-a15e59d993eb0844d736a04652f63d465d909127e84cf7cd27073076df530f05.css
```php
$source = file_get_contents('https://assets-cdn.github.com/assets/github-a15e59d993eb0844d736a04652f63d465d909127e84cf7cd27073076df530f05.css');
$time = microtime(true);
$parser = new Sabberworm\CSS\Parser($source);
$css = $parser->parse();
var_dump(microtime(true) - $time);
```
On the my machine I received the following results:
before `float(180.93577218056)`
after `float(1.1720938682556)`

### DIFF
--- a/lib/Sabberworm/CSS/Parser.php
+++ b/lib/Sabberworm/CSS/Parser.php
@@ -583,8 +583,14 @@ class Parser {
 	}
 
 	private function substr($iStart, $iLength) {
+		if ($iLength < 0) {
+			$iLength = $this->iLength - $iStart + $iLength;
+		}
+		if ($iStart + $iLength > $this->iLength) {
+			$iLength = $this->iLength - $iStart;
+		}
 		$out = '';
-		while ($iLength > 0 && $iStart < $this->iLength) {
+		while ($iLength > 0) {
 			$out .= $this->aText[$iStart];
 			$iStart++;
 			$iLength--;
--- a/lib/Sabberworm/CSS/Parser.php
+++ b/lib/Sabberworm/CSS/Parser.php
@@ -490,7 +490,6 @@ class Parser {
 		if ($iOffset >= $this->iLength) {
 			return '';
 		}
-		$iLength = min($iLength, $this->iLength-$iOffset);
 		$out = $this->substr($iOffset, $iLength);
 		return $out;
 	}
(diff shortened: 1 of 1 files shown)


sabberworm:
Why 150000000% while you’re at it…

Just kidding. Great work, thanks a lot!

lib/Sabberworm/CSS/Parser.php
```diff
@@ -56,7 +59,7 @@ public function __construct($sText, Settings $oParserSettings = null) {
 
 	public function setCharset($sCharset) {
 		$this->sCharset = $sCharset;
-		$this->iLength = $this->strlen($this->sText);
+		$this->iLength = count($this->aText);
```
sabberworm: I see a problem here: splitting should only happen after the charset is known (because it depends on the charset used).

lib/Sabberworm/CSS/Parser.php
```diff
@@ -26,22 +26,25 @@
...
 
 	public function __construct($sText, Settings $oParserSettings = null) {
-		$this->sText = $sText;
 		$this->iCurrentPosition = 0;
 		if ($oParserSettings === null) {
 			$oParserSettings = Settings::create();
 		}
 		$this->oParserSettings = $oParserSettings;
+		if ($this->oParserSettings->bMultibyteSupport) {
+			$this->aText = preg_split('//u', $sText, null, PREG_SPLIT_NO_EMPTY);
```
sabberworm: I’m not sure `preg_split` is multibyte-safe. Sure it supports utf-8 (if compiled in) but I don’t think it copes well with other encodings like utf-16, shift-jis. Also there is no way to pass the encoding used. I’ve never added unit tests for other encodings, though, because the identifier parsing code wasn’t actually able to match any encoding where the identifiers weren’t ASCII. Maybe I’ll just make that my next big rework project…

lib/Sabberworm/CSS/Parser.php
sabberworm: Also, the `@charset` declaration will call this method during parsing. But it should trigger a re-parse if the count is different in the new charset, which it has never done so again, this is due for a rework that hasn’t necessarily got anything to do with your changeset.

ossinkine: If it so, why is `bMultibyteSupport` flag needed? Can I pass `bMultibyteSupport = false` and `@charset "UTF-8"`?

Issue #104, opened 2016-06-25
Missing space in @media(...) throws exception


voehl:
Unfortunately, when parsing stylesheets without the space in "@ media(" the parser throws an UnexpectedTokenException.

Example:
```
<?php
use Sabberworm\CSS\Parser;
require '../vendor/autoload.php';

// This works:
(new Parser('@media (min-width: 768px){.class{color:red}}'))->parse();

// This throws an exception: 'Identifier expected. Got “: 768”'
(new Parser('@media(min-width: 768px){.class{color:red}}'))->parse();
```

sabberworm:
Closing as duplicate of #75

Issue #85, opened 2015-01-23
Parse Exception for valid css


maraisr:
This selector wont pass, and I get a Identifier expected. Got ":80em"

This is the selector
`@media(min-width:80em){a[href^=fax],a[href^=tel]{cursor:text;pointer-events:none}a[href^=fax]:hover,a[href^=tel]:hover{text-decoration:none}}`

sabberworm:
Closing this as a duplicate of #75

Issue #75, opened 2014-01-09
Support CSS3 media queries


Jeedey:
I try to parse a last Boostrap Css but i got an error to Sabberworm/CSS/Parser.php at line 171.
Maybe it's can parse this line : `@media (min-width: 768px) {.lead {font-size: 21px;}}`
Any idea?

Thanks

rafasashi:
For your information this function parses css media queries:

https://github.com/recuweb-source/php-dom-scraper/blob/master/functions_dom_parser.php#L200

Mipme:
Since more and more css contains animation keyframes i had to update my media query replacement from above:

    $code = preg_replace("/@media\s*\(([^)]+)\)\s*\{([^\{]+\{[^\}]+\})*\}/mi", "/* @media-begin($1) */\n$2\n/* @media-end($1) */\n", $code);

This wraps the media query with comments ````/* media-begin([condition]) */```` and the corresponding end tag.

ivanjaros:
Any (partial?)progress with this 3 year old issue?

sabberworm:
@ivanjaros I’m not currently actively working on new features for the project. But I’m very much maintaining it so if you fix this and open a pull-request I’ll do my best to get it merged.

PR #105, opened 2016-06-29
Add line number support in the sabberworm CSS parser

sidkshatriya:
AMP HTML is a restricted subset of HTML (See https://www.ampproject.org/ for more details) and the AMP PHP Library provide validation and HTML to AMP HTML conversion utilities. We're using the `masterminds/html5-php` library to parse HTML5 and we're using `sabberworm/php-css-parser` to parse the CSS in the [AMP PHP Library](https://github.com/Lullabot/amp-library)

After parsing the CSS parse we report any problems in the CSS (for e.g. `!important` is disallowed in the AMP dialect of HTML, we use the `sabberworm` parser to find all the uses of `!important` for instance)

Now the CSS supplied by the user could be hundreds / thousands of lines long. So its not useful to simply say that `!important` was used and is not allowed. We should be able to say _which line_ the problem occurred.

This pull request adds support for line numbers. Here is how we're using this https://github.com/Lullabot/amp-library/pull/80#issuecomment-222695630
Notice the line numbers reported in the validation errors which coming from code submitted in this PR.

Would greatly appreciate if you had a look and merge it if you believe this feature would be useful. Hopefully the code changes have not missed anything out.

Thanks for this useful library BTW.

### DIFF
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -2,18 +2,16 @@
 
 namespace Sabberworm\CSS\CSSList;
 
+use Sabberworm\CSS\Renderable;
 use Sabberworm\CSS\RuleSet\DeclarationBlock;
 use Sabberworm\CSS\RuleSet\RuleSet;
 use Sabberworm\CSS\Property\Selector;
-use Sabberworm\CSS\Rule\Rule;
-use Sabberworm\CSS\Value\ValueList;
-use Sabberworm\CSS\Value\CSSFunction;
 
 /**
  * A CSSList is the most generic container available. Its contents include RuleSet as well as other CSSList objects.
  * Also, it may contain Import and Charset objects stemming from @-rules.
  */
-abstract class CSSList {
+abstract class CSSList implements Renderable {
 
 	protected $aContents;
 	protected $iLineNum;
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -29,6 +29,14 @@ abstract class CSSList implements Renderable {
 		return $this->iLineNum;
 	}
 
+	/**
+	 * @param int $iLineNum
+	 */
+	public function setLineNo($iLineNum = 0)
+	{
+		$this->iLineNum = $iLineNum;
+	}
+
 	public function append($oItem) {
 		$this->aContents[] = $oItem;
 	}
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -24,16 +24,14 @@ abstract class CSSList implements Renderable {
 	/**
 	 * @return int
 	 */
-	public function getLineNo()
-	{
+	public function getLineNo() {
 		return $this->iLineNo;
 	}
 
 	/**
 	 * @param int $iLineNo
 	 */
-	public function setLineNo($iLineNo = 0)
-	{
+	public function setLineNo($iLineNo = 0) {
 		$this->iLineNo = $iLineNo;
 	}
 
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -14,11 +14,11 @@ use Sabberworm\CSS\Property\Selector;
 abstract class CSSList implements Renderable {
 
 	protected $aContents;
-	protected $iLineNum;
+	protected $iLineNo;
 
-	public function __construct($iLineNum = 0) {
+	public function __construct($iLineNo = 0) {
 		$this->aContents = array();
-		$this->iLineNum = $iLineNum;
+		$this->iLineNo = $iLineNo;
 	}
 
 	/**
@@ -26,15 +26,15 @@ abstract class CSSList implements Renderable {
 	 */
 	public function getLineNo()
 	{
-		return $this->iLineNum;
+		return $this->iLineNo;
 	}
 
 	/**
-	 * @param int $iLineNum
+	 * @param int $iLineNo
 	 */
-	public function setLineNo($iLineNum = 0)
+	public function setLineNo($iLineNo = 0)
 	{
-		$this->iLineNum = $iLineNum;
+		$this->iLineNo = $iLineNo;
 	}
 
 	public function append($oItem) {
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -28,13 +28,6 @@ abstract class CSSList implements Renderable {
 		return $this->iLineNo;
 	}
 
-	/**
-	 * @param int $iLineNo
-	 */
-	public function setLineNo($iLineNo = 0) {
-		$this->iLineNo = $iLineNo;
-	}
-
 	public function append($oItem) {
 		$this->aContents[] = $oItem;
 	}
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -16,9 +16,19 @@ use Sabberworm\CSS\Value\CSSFunction;
 abstract class CSSList {
 
 	protected $aContents;
+	protected $iLineNum;
 
-	public function __construct() {
+	public function __construct($iLineNum = 0) {
 		$this->aContents = array();
+		$this->iLineNum = $iLineNum;
+	}
+
+	/**
+	 * @return int
+	 */
+	public function getLineNum()
+	{
+		return $this->iLineNum;
 	}
 
 	public function append($oItem) {
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -24,7 +24,7 @@ abstract class CSSList implements Renderable {
 	/**
 	 * @return int
 	 */
-	public function getLineNum()
+	public function getLineNo()
 	{
 		return $this->iLineNum;
 	}
--- a/lib/Sabberworm/CSS/Parser.php
+++ b/lib/Sabberworm/CSS/Parser.php
@@ -37,7 +37,7 @@ class Parser {
 	private $aSizeUnits;
 	private $iLineNo;
 
-	public function __construct($sText, Settings $oParserSettings = null, $iLineNo = 0) {
+	public function __construct($sText, Settings $oParserSettings = null, $iLineNo = 1) {
 		$this->sText = $sText;
 		$this->iCurrentPosition = 0;
 		$this->iLineNo = $iLineNo;
--- a/lib/Sabberworm/CSS/Parser.php
+++ b/lib/Sabberworm/CSS/Parser.php
@@ -34,12 +34,12 @@ class Parser {
 	private $iLength;
 	private $blockRules;
 	private $aSizeUnits;
-	private $iLineNum;
+	private $iLineNo;
 
-	public function __construct($sText, Settings $oParserSettings = null, $iLineNum = 0) {
+	public function __construct($sText, Settings $oParserSettings = null, $iLineNo = 0) {
 		$this->sText = $sText;
 		$this->iCurrentPosition = 0;
-		$this->iLineNum = $iLineNum;
+		$this->iLineNo = $iLineNo;
 		if ($oParserSettings === null) {
 			$oParserSettings = Settings::create();
 		}
@@ -68,7 +68,7 @@ class Parser {
 
 	public function parse() {
 		$this->setCharset($this->oParserSettings->sDefaultCharset);
-		$oResult = new Document($this->iLineNum);
+		$oResult = new Document($this->iLineNo);
 		$this->parseDocument($oResult);
 		return $oResult;
 	}
@@ -109,10 +109,10 @@ class Parser {
 			$oAtRule = $this->parseAtRule();
 			if($oAtRule instanceof Charset) {
 				if(!$bIsRoot) {
-					throw new UnexpectedTokenException('@charset may only occur in root document', '', 'custom', $this->iLineNum);
+					throw new UnexpectedTokenException('@charset may only occur in root document', '', 'custom', $this->iLineNo);
 				}
 				if(count($oList->getContents()) > 0) {
-					throw new UnexpectedTokenException('@charset must be the first parseable token in a document', '', 'custom', $this->iLineNum);
+					throw new UnexpectedTokenException('@charset must be the first parseable token in a document', '', 'custom', $this->iLineNo);
 				}
 				$this->setCharset($oAtRule->getCharset()->getString());
 			}
@@ -132,7 +132,7 @@ class Parser {
 	private function parseAtRule() {
 		$this->consume('@');
 		$sIdentifier = $this->parseIdentifier();
-		$iIdentifierLineNum = $this->iLineNum;
+		$iIdentifierLineNum = $this->iLineNo;
 		$this->consumeWhiteSpace();
 		if ($sIdentifier === 'import') {
 			$oLocation = $this->parseURLValue();
@@ -195,7 +195,7 @@ class Parser {
 	private function parseIdentifier($bAllowFunctions = true, $bIgnoreCase = true) {
 		$sResult = $this->parseCharacter(true);
 		if ($sResult === null) {
-			throw new UnexpectedTokenException($sResult, $this->peek(5), 'identifier', $this->iLineNum);
+			throw new UnexpectedTokenException($sResult, $this->peek(5), 'identifier', $this->iLineNo);
 		}
 		$sCharacter = null;
 		while (($sCharacter = $this->parseCharacter(true)) !== null) {
@@ -207,7 +207,7 @@ class Parser {
 		if ($bAllowFunctions && $this->comes('(')) {
 			$this->consume('(');
 			$aArguments = $this->parseValue(array('=', ' ', ','));
-			$sResult = new CSSFunction($sResult, $aArguments, ',', $this->iLineNum);
+			$sResult = new CSSFunction($sResult, $aArguments, ',', $this->iLineNo);
 			$this->consume(')');
 		}
 		return $sResult;
@@ -241,7 +241,7 @@ class Parser {
 			}
 			$this->consume($sQuote);
 		}
-		return new CSSString($sResult, $this->iLineNum);
+		return new CSSString($sResult, $this->iLineNo);
 	}
 
 	private function parseCharacter($bIsForIdentifier) {
@@ -290,7 +290,7 @@ class Parser {
 	}
 
 	private function parseSelector() {
-		$oResult = new DeclarationBlock($this->iLineNum);
+		$oResult = new DeclarationBlock($this->iLineNo);
 		$oResult->setSelector($this->consumeUntil('{', false, true));
 		$this->consumeWhiteSpace();
 		$this->parseRuleSet($oResult);
@@ -336,7 +336,7 @@ class Parser {
 	}
 
 	private function parseRule() {
-		$oRule = new Rule($this->parseIdentifier(), $this->iLineNum);
+		$oRule = new Rule($this->parseIdentifier(), $this->iLineNo);
 		$this->consumeWhiteSpace();
 		$this->consume(':');
 		$oValue = $this->parseValue(self::listDelimiterForRule($oRule->getRule()));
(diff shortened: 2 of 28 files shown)


sabberworm:
Wow, thanks for your work. I will review soon…

Maybe you could add a unit test dealing specifically with line numbers and refactor the one you changed to ignore the line numbers so it doesn’t break when we change the source?

lib/Sabberworm/CSS/Parser.php
```diff
@@ -516,13 +519,18 @@ private function peek($iLength = 1, $iOffset = 0) {
 
 	private function consume($mValue = 1) {
 		if (is_string($mValue)) {
+			$noLines = substr_count($mValue, "\n");
```
sabberworm: Call the variable `$iLineCount` for consistency.

lib/Sabberworm/CSS/Parser.php
```diff
@@ -516,13 +519,18 @@ private function peek($iLength = 1, $iOffset = 0) {
...
 			$iLength = $this->strlen($mValue);
 			if (!$this->streql($this->substr($this->iCurrentPosition, $iLength), $mValue)) {
 				throw new UnexpectedTokenException($mValue, $this->peek(max($iLength, 5)));
 			}
+			$this->iLineNum += $noLines;
 			$this->iCurrentPosition += $this->strlen($mValue);
 			return $mValue;
 		} else {
+			$substring = $this->substr($this->iCurrentPosition, $mValue);
```
sabberworm: Why get the substring here when we already have it as `$sResult` two lines down?

sabberworm:
Add the `getLineNum` method declaration to the `Renderable` interface.
And name it either `getLineNo` or `getLineNumber`.

The `CSSList` class should also implement `Renderable`, I forgot to do that.

lib/Sabberworm/CSS/Parser.php
```diff
@@ -516,13 +519,18 @@ private function peek($iLength = 1, $iOffset = 0) {
...
+			$noLines = substr_count($substring, "\n");
+			$this->iLineNum += $noLines;
 			if ($this->iCurrentPosition + $mValue > $this->iLength) {
 				throw new UnexpectedTokenException($mValue, $this->peek(5), 'count');
```
sabberworm: Add line numbers to exceptions as well for easier debugging.

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -26,11 +24,19 @@ public function __construct($iLineNum = 0) {
 	/**
 	 * @return int
 	 */
-	public function getLineNum()
+	public function getLineNo()
 	{
 		return $this->iLineNum;
```
sabberworm: Rename the variable, too.

lib/Sabberworm/CSS/Parser.php
sidkshatriya: Fixed

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -26,11 +24,19 @@ public function __construct($iLineNum = 0) {
 	/**
 	 * @return int
 	 */
-	public function getLineNum()
+	public function getLineNo()
```
sabberworm: Put the opening brace on the same line.

lib/Sabberworm/CSS/Parser.php
sidkshatriya: Fixed

lib/Sabberworm/CSS/Parser.php
sidkshatriya: Done

sabberworm:
Thanks. I do have one minor nitpick: the `Renderable` interface should not declare the setter for the line number, only the getter because in order for something to be renderable, the line-number need not be state, it could also be computed upon calling.

lib/Sabberworm/CSS/Parsing/OutputException.php
```diff
@@ -6,4 +6,13 @@
 * Thrown if the CSS parsers attempts to print something invalid
 */
 class OutputException extends \Exception {
+    private $iLineNum;
```
sabberworm: Provide a getter for the line number.

lib/Sabberworm/CSS/Parsing/OutputException.php
```diff
@@ -6,4 +6,13 @@
...
+    public function __construct($sMessage, $iLineNum = 0)
```
sabberworm: Put opening brace on same line as method declaration.

sidkshatriya:
Some notes:
- The `OutputException` and `UnexpectedTokenException` exception classes now have line number information (see the throws) but I think there are 4 places where you're just throwing `new \Exception` -- it would logical to add line numbers there too -- but I'm not sure if should create another custom exception class or re-use the `UnexpectedTokenException` (please let me know what params one must use)
- CSSList implements Renderable now. **In addition Rule also implements Renderable now**

> Thanks. I do have one minor nitpick: the Renderable interface should not declare the setter for the line number, only the getter because in order for something to be renderable, the line-number need not be state, it could also be computed upon calling.

Its a good idea to have setLineNo generally, mainly for testing purposes where you want to reset the stored line number to 0. What is your alternate suggestion?

lib/Sabberworm/CSS/CSSList/CSSList.php
sidkshatriya: Done

lib/Sabberworm/CSS/Parsing/OutputException.php
sidkshatriya: Done

lib/Sabberworm/CSS/Parsing/OutputException.php
sidkshatriya: Done

lib/Sabberworm/CSS/Parsing/OutputException.php
```diff
@@ -7,12 +7,15 @@
...
+    public function __construct($sMessage, $iLineNo = 0) {
         $this->$iLineNo = $iLineNo;
         if (!empty($iLineNo)) {
             $sMessage .= " [line no: $iLineNo]";
         }
         parent::__construct($sMessage);
     }
+
+    public function getLineNo() {
+        return $this->getLine();
```
sabberworm: I meant for the CSS line numner (`$iLineNo` member), not for the line number of the PHP code that threw the exception.

sidkshatriya: @sabberworm Good catch! I've fixed that.

Anyways, I renamed `$iLineNum` to `$iLineNo` everywhere -- is that OK? Or should I revert to the older `$iLineNum` ?

sabberworm: > On 30.06.2016, at 11:37, Sidharth Kshatriya <notifications@github.com> wrote:
> 
> In lib/Sabberworm/CSS/Parsing/OutputException.php:
> 
> >          $this->$iLineNo = $iLineNo;
> >          if (!empty($iLineNo)) {
> >              $sMessage .= " [line no: $iLineNo]";
> >          }
> >          parent::__construct($sMessage);
> >      }
> > +
> > +    public function getLineNo() {
> > +        return $this->getLine();
> 
> @sabberworm Good catch! I've fixed that.
> 
> Anyways, I renamed $iLineNum to $iLineNo everywhere -- is that OK? Or should I revert to the older $iLineNum ?

No, it’s perfect.

sabberworm:
Regarding the exceptions:

I used the generic `Exception` class when we encounter state we can’t possibly recover from. Maybe we should add a class `SourceException` (exception in source code) which has a line number and use it in place of `Exception`. We could then let `OutputException` and `UnexpectedTokenException` inherit from that. This should even be backwards-compatible unless someone explicitly uses `get_class` to get the type of an exception instead of using `instanceof`.

Regarding the line number setter: I would have suggested just removing it from the interface. I was fine with leaving it in the implementations. But seeing the setters are only used in the unit test, I’m thinking we should get rid of them altogether and change this

	$oColor->setLineNo(0);
	$this->assertEquals(array('r' => new Size(35.0, null, true), 'g' => new Size(35.0, null, true), 'b' => new Size(35.0, null, true)), $oColor->getColor());

into this:

	$this->assertEquals(array('r' => new Size(35.0, null, true, $oColor->getLineNo()), 'g' => new Size(35.0, null, true, $oColor->getLineNo()), 'b' => new Size(35.0, null, true, $oColor->getLineNo())), $oColor->getColor());

sidkshatriya:
@sabberworm 
- We pass in the line number as suggested so that the assertEqual has the correct value for the color test
- Removed `setLineNo()` from everywhere
- Added new class `SourceException`

lib/Sabberworm/CSS/Parsing/UnexpectedTokenException.php
```diff
@@ -28,11 +28,7 @@ public function __construct($sExpected, $sFound, $sMatchType = 'literal', $iLine
...
 
-		if (!empty($iLineNo)) {
-			$sMessage .= " [line no: $iLineNo]";
-		}
-
-		parent::__construct($sMessage);
+		parent::__construct($sMessage, $iLineNo);
 	}
 
 	public function getLineNo() {
```
sabberworm: Remove `getLineNo` and `private $iLineNo;`.

sidkshatriya: @sabberworm Good catch! Done!

sabberworm:
Do you want to add unit tests for line numbers or shall I? I can do it next week and then release. Does that work for you?

sidkshatriya:
@sabberworm If I add the test myself in sometime would the timelines be earlier? When might you merge to master? When might you create a release tag? (I'm asking because I would need to change some `composer.json` dependencies accordingly for the AMP PHP Library)

No hurries, of course, but I just wanted to know your plan...

sabberworm:
If you don’t add unit tests, I will merge to master now and add the unit tests next week. If you intend to add unit tests now, I will wait with merging until you’re done and release immediately afterwards.

sidkshatriya:
@sabberworm OK -- I'll add a test now... (Thanks for being super responsive on this PR -- Appreciate it)

tests/files/line-numbers.css
```diff
@@ -0,0 +1,28 @@
+@charset "utf-8"; /* line 3 */
```
sabberworm: Comment should say `1`

tests/files/line-numbers.css
```diff
@@ -0,0 +1,28 @@
...
+
+@namespace "http://toto.example.org"; /* line 1 */
```
sabberworm: Comment should say `3`

lib/Sabberworm/CSS/Parser.php
```diff
@@ -232,13 +236,13 @@ private function parseStringValue() {
 			while (!$this->comes($sQuote)) {
 				$sContent = $this->parseCharacter(false);
 				if ($sContent === null) {
-					throw new \Exception("Non-well-formed quoted string {$this->peek(3)}");
+					throw new \Exception("Non-well-formed quoted string {$this->peek(3)}", $this->iLineNo);
```
sabberworm: Change to `SourceException`

tests/files/line-numbers.css
sidkshatriya: Changed

tests/files/line-numbers.css
sidkshatriya: Changed

tests/files/line-numbers.css
```diff
@@ -0,0 +1,28 @@
...
+    from { top: 0px; } /* line 18 */
+
+    to { top: 200px; } /* line 20 */
+}
+
+@IMPORT uRL(test.css); /* line 23 */
+
+body {
+    background: #FFFFFF url("http://somesite.com/images/someimage.gif") repeat top center; /* line 25 */
+    color: rgb(233, 100, 450);
```
sabberworm: Could you maybe split up the color components into separate lines and check if they are identified correctly:

```css
color: rgb( /* line 26 */
       233, /* line 27 */
       100, /* line 28 */
       450 /* line 29 */
);
```

sabberworm:
@sidkshatriya With any luck we can drop PHP 5.3 somewhere in the next year…

lib/Sabberworm/CSS/Parsing/SourceException.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+namespace Sabberworm\CSS\Parsing;
+
+class SourceException extends \Exception {
+    private $iLineNo;
+    public function __construct($sMessage, $iLineNo = 0) {
+        $this->$iLineNo = $iLineNo;
```
sabberworm: Remove the `$`.

sabberworm:
- [ ] Add some tests for exceptions, e.g.:

```php
/**
* @expectedException Sabberworm\CSS\Parsing\UnexpectedTokenException
*/
public function testUnexpectedTokenExceptionLineNo() {
	$oParser = new Parser("\ntest: 1;", Settings::create()->beStrict());
	try {
		$oParser->parse();
	} catch(UnexpectedTokenException $e) {
		$this->assertSame($e->getLineNo(), 2);
		throw $e;
	}
}
```

Ideally there would be a test for every place in the parser where we throw an exception.

lib/Sabberworm/CSS/Parsing/SourceException.php
sidkshatriya: @sabberworm Why should we remove the `$` -- sorry I didn't understand...

sidkshatriya:
@sabberworm Hopefully things are now looking good. Let me know if you have any other suggestions (thanks for your really responsive help on this!). 

BTW My contributions on this pull request are on behalf of https://github.com/Lullabot

I'll pick up with this tomorrow (or later tonight if I get the chance)

lib/Sabberworm/CSS/Parsing/SourceException.php
sabberworm: > On 30.06.2016, at 14:45, Sidharth Kshatriya <notifications@github.com> wrote:
> 
> @sabberworm Why should we remove the $ -- sorry I didn't understand...
> 
> 

It’s just a typo. The line reads `$this->$iLineNo = $iLineNo;` but should be `$this->iLineNo = $iLineNo;`.

If $iLineNo is `1`, the former sets `$this->1` to `1`, the latter sets `$this->iLineNo` to `1`.

The newly-added test `testUnexpectedTokenExceptionLineNo` should have caught this but didn’t because you didn’t `use Sabberworm\CSS\Parsing\UnexpectedTokenException`.

sabberworm: It’s just a typo. The line reads `$this->$iLineNo = $iLineNo;` but should be `$this->iLineNo = $iLineNo;`.

If $iLineNo is `1`, the former sets `$this->1` to `1`, the latter sets `$this->iLineNo` to `1`.

The newly-added test `testUnexpectedTokenExceptionLineNo` should have caught this but didn’t because you didn’t `use Sabberworm\CSS\Parsing\UnexpectedTokenException`.

tests/Sabberworm/CSS/ParserTest.php
```diff
@@ -435,4 +438,70 @@ function parsedStructureForFile($sFileName, $oSettings = null) {
...
+
+	/**
+	 * @expectedException Sabberworm\CSS\Parsing\UnexpectedTokenException
+	 */
+	public function testUnexpectedTokenExceptionLineNo() {
+		$oParser = new Parser("\ntest: 1;", Settings::create()->beStrict());
+		try {
+			$oParser->parse();
+		} catch (UnexpectedTokenException $e) {
+			$this->assertSame($e->getLineNo(), 2);
```
sabberworm: Switch the arguments: expected should come first, actual later.

tests/Sabberworm/CSS/ParserTest.php
```diff
@@ -435,4 +438,70 @@ function parsedStructureForFile($sFileName, $oSettings = null) {
...
+
+		$this->assertEquals($aExpectedColorLines, $aActualColorLines);
+		$this->assertEquals($aUrlExpected, $aUrlActual);
+		$this->assertEquals($aExpected, $aActual);
+	}
+
+	/**
+	 * @expectedException Sabberworm\CSS\Parsing\UnexpectedTokenException
+	 */
+	public function testUnexpectedTokenExceptionLineNo() {
```
sabberworm: You can remove `public`.

tests/Sabberworm/CSS/ParserTest.php
```diff
@@ -435,4 +438,70 @@ function parsedStructureForFile($sFileName, $oSettings = null) {
...
+		$oParser = new Parser("\ntest: 1;", Settings::create()->beStrict());
+		try {
+			$oParser->parse();
+		} catch (UnexpectedTokenException $e) {
```
sabberworm: Either use the full class namespace `Sabberworm\CSS\Parsing\UnexpectedTokenException` or `use Sabberworm\CSS\Parsing\UnexpectedTokenException` at the top of the file.

tests/Sabberworm/CSS/ParserTest.php
sidkshatriya: @sabberworm done.

tests/Sabberworm/CSS/ParserTest.php
sidkshatriya: Done

lib/Sabberworm/CSS/Parsing/SourceException.php
sidkshatriya: @sabberworm This was a fantastic catch! Sorry I didn't notice the extraneous `$`

sidkshatriya:
@sabberworm Ah you merged! Great! Thanks so much!

sabberworm:
> On 30.06.2016, at 16:41, Sidharth Kshatriya <notifications@github.com> wrote:
> 
> @sabberworm Ah you merged! Great! Thanks so much!

Thank you for your great work! Now I only need someone like you to implement selector and media query parsing…

How do you prefer to be listed under contributors, as individual or as Lullabot?

sidkshatriya:
as Lullabot

**re:**  selector and media query parsing -- I could look at it in my spare time. If you make a ticket chalking out the details on whats incomplete / what needs to be done etc. it would be awesome. What happens with media queries right now? Does the parser choke or simply not get the finer structure?

sabberworm:
The selector parsing is really rudimentary, it just searches for the next `{` and splits by comma. So it has trouble parsing a selector `like a[href*=","] .complicated\{selector`.

Media query parsing is even worse. It was only designed to handle `@media print {}` and chokes on more complex stuff. See #75 and its duplicates (#85, #104) for examples.

sabberworm:
Fixing media queries should be a lot simpler than fixing selectors because there are fewer cases to consider and since the current selector parsing works well enough for real-world, non-contrived examples, it’s also the bigger pain point.

sidkshatriya:
@sabberworm 
Is your parser hand built or have you consulted something like https://drafts.csswg.org/css-syntax-3/ for your tokenizing/parsing algorithm?

I don't really need media query parsing or selector parsing for the AMP library at the moment (may need it in future) -- but I'm concerned if the parser will abort and issue an exception. Its OK for instance to not do an accurate job of parsing for things like selectors/media queries but aborting the whole parse is not so good... What happens in such cases?

sidkshatriya:
@sabberworm sorry for persisting but I just want to be aware of any parse related aborts due to syntax the parser does not understand. Like I mentioned, it's ok if you don't parse the complex selectors into its fine components or even get media query parsing correct. I just want the parser to skip over those things (or even do a bad job with them). What _is_ problematic is the whole parse getting aborted due to unrecognised (but valid) css syntax.

sabberworm:
Some well-defined exception states are recoverable. But because the parser does not know how media queries are supposed to be structures, it can’t recover and throws a `SourceException`. But I will try to implement correct media query parsing soon since it’s long been one of the major shortcomings of this library.

sidkshatriya:
@sabberworm Thanks for responding and eagerly look forward to some basic media query parsing. 

Also to my previous question: is your parsing algorithm hand built or you have consulted something like https://drafts.csswg.org/css-syntax-3/ ? IIRC that there may be general strategies in that link where you can prevent yourself from aborting a parse when you encounter syntax that you don't know about. Basically you should simply skip over to the next at-rule or declaration block that you _do_ understand.

sabberworm:
The goal of the parser was always to parse valid CSS. Handling invalid CSS and being fault-tolerant was a feature added later on and does not exactly follow the way the specification demands of browser vendors.

However, for valid CSS (for everything the parser supports, which, currently, does not include media queries) the goal was always to be compliant with the specs, even if the resulting data structure does not adhere to CSSOM.

Issue #38, opened 2012-06-11
Parse Comments?


billdami:
Would it be possible to add the ability to parse comments that are both associated with a specific declaration block (aka are within the rule's opening and closing braces), and just global comments that are at the root document level?

For instance, for a CSS snippet like this:

``` css
h1 {
  /* my comment */
  color: #000000;
  font-size: 24px;
}
```

Would be parsed to (in simplified structure):

```
    declaration block =>
        selector =>  "h1"
        comments => array
            "my comment"
        ruleset => array
            ....
```

PR #110, opened 2016-07-15
Primary support for comment parsing

FMCorz:
The following patch is adding primary support to parsing comments, it does not try to do everything and it does not include the render part (as we don't need it) but hopefully it'll be useful to others.

In short, for the most generic cases, it adds the comments preceding a statement to the statement itself. See the test for more info.

X-ref: #38


NagarLalit:
Thanks for the patch. I can get comment related to particular rule set. But I am not able to render with comments i.e. comments are not visible when I save the file.

FMCorz:
Yup, the rendering is not part of the patch.

sabberworm:
Thanks a lot!

PR #111, opened 2016-07-18
Parsing of filter rules specific to Microsoft

FMCorz:
This allows for filters typically used for gradients to be parsed. The following was taken from Bootstrap 4.

```css
filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='#80000000', endColorstr='#00000000', GradientType=1);
```

I am not sold on the Value I've introduced here, especially as a child of PrimitiveValue. Please let me know how to adapt my patch should this change interest you.

Cheers!


sabberworm:
I’m a bit biased against this as it really isn’t valid CSS and the more modern syntax (with `-ms-filter` is available to IEs dating back to IE 8). I’m not sure there is much use for the old-old syntax anymore. But maybe we can accept and silently discard it in lenient mode (on second thought, I think this may already be what we are doing but I’m not sure)? Or, add a new config that adds support for this. But then, please make it opt-in and make the class that represents the result a subclass of `CSSFunction`.

FMCorz:
Now that you mention it, I discovered that IE9 does not even support this format. Strangely it is still found in Bootstrap 4 (as of today) as you can see [here](https://github.com/twbs/bootstrap/blob/44a136107937fba8b7906f89c01894655fca5c2e/scss/mixins/_gradients.scss). I did change to only allow the old-style filters in lenient mode though, and tested that string quoting had no impact on IE8 using this:

```html
<html>
    <head>
        <style>
            div {
                height: 100px;;
                width: 100%;
                margin-bottom: 10px;
            }
            .a {
                filter: progid:DXImageTransform.Microsoft.gradient(startColorstr='#FF000000', endColorstr='#00000000', GradientType=1);
            }
            .b {
                filter: progid:DXImageTransform.Microsoft.gradient(startColorstr="#FF000000", endColorstr="#00000000", GradientType=1);
            }
            .c {
                filter: progid:DXImageTransform.Microsoft.gradient(startColorstr=#FF000000, endColorstr=#00000000, GradientType=1);
            }
        </style>
    </head>
    <body>
        <div class="a">Here</div>
        <div class="b">There</div>
        <div class="c">Over there</div>
    </body>
</html>
```

Thanks for suggesting to use a CSSFunction, I did not realise I could.

Cheers.
Fred

sabberworm:
Great, thanks a lot!

PR #112, opened 2016-07-18
Support for IE hack parsing in lenient mode

FMCorz:
For your consideration. This allows for \0 and \9 to remain as they are intended. And will throw an exception when they are found in strict mode. Again, that to be able to parse Bootstrap 4.


sabberworm:
This library has never supported CSS hacks (the slash hack or the star hack come to mind – and for good reason: these are all but historical footnotes now and barely relevant) but I’m willing to accept it. I will, however, back it out when I find the usage of these hacks in the wild to be too low to warrant having this code around.

PR #115, opened 2016-08-01
Support for inserting an item in a CSSList

FMCorz:
For your consideration.

### DIFF
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -36,6 +36,20 @@ abstract class CSSList implements Renderable, Commentable {
 	}
 
 	/**
+	 * Insert an item before its sibling.
+	 *
+	 * @param mixed $oItem The item.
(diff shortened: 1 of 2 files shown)


raxbg:
I like the idea, but the method name `insert` suggests that it must have a parameter for position: before or after. Also what if we just map this to the existing `replace`? Suggestion:
```PHP
	public function insert($oItem, $oSibling, $sPosition = 'before') {
		$iIndex = array_search($oSibling, $this->aContents);
		if ($iIndex === false) {
			return $this->append($oItem);
		} else if ($sPosition === 'before') {
			$this->replace($oItem, array($oSibling, $oItem));
		} else {
			$this->replace($oItem, array($oItem, $oSibling));
		}
	}
```

ziegenberg:
As there was no progress on this for the last five years I created #545.

ziegenberg:
@oliverklee I think this can now be closed.

oliverklee:
Indeed it can. Thanks for implementing/finishing this, @ziegenberg! :heart:

Implemented in #545.

PR #116, opened 2016-08-20
Fix encoding issues and improve parsing performance

skodak:
Hello,

I was wondering if you would be interested in a patch that tries to solve some charset related issues and improves parsing performance. I was trying to parse a 500kb CSS file (Bootstrap 3, Font Awesome and some custom stuff) and it took more than 70 seconds, with this patch I managed to get it down to 3.5-7 seconds. While working on the patch I have noticed a few other problems.

List of changes:
- internally all strings are now stored as UTF-8
- BOM is used for utf-8 detection and removed if found
- setCharset() was removed completely - all charset conversions are done in constructor
- the rendered text is always returned as utf-8 with all CSS strings \encoded as ASCII chars
- the @charset atRule is switched to 'utf-8' if present
- mbstring extension is not required any more
- iconv is required - previously it was required only if there were \encoded unicode chars
- all non-ascii unicode chars are \encoded in CSS strings
- null character is completely ignored in CSS strings for security reasons

Ciao,
Petr


lib/Sabberworm/CSS/RuleSet/DeclarationBlock.php
```diff
@@ -226,7 +226,11 @@ public function expandFontShorthand() {
 		}
 		foreach ($aValues as $mValue) {
 			if (!$mValue instanceof Value) {
-				$mValue = mb_strtolower($mValue);
+				if (function_exists('mb_strtolower')) {
+					$mValue = mb_strtolower($mValue, 'utf-8');
+				} else {
+					$mValue = strtolower($mValue);
```
sabberworm: Same here…

lib/Sabberworm/CSS/Parser.php
```diff
@@ -688,40 +762,12 @@ private function streql($sString1, $sString2, $bCaseInsensitive = true) {
 	}
 
 	private function strtolower($sString) {
-		if ($this->oParserSettings->bMultibyteSupport) {
-			return mb_strtolower($sString, $this->sCharset);
+		if ($this->sTextLibrary === 'mb') {
+			return mb_strtolower($sString, 'utf-8');
```
sabberworm: Since the `strtolower` method is only used for identifiers, which, after your change, are always ASCII, just calling `strtolower` should suffice.

lib/Sabberworm/CSS/RuleSet/DeclarationBlock.php
```diff
@@ -369,7 +377,11 @@ public function expandListStyleShorthand() {
...
```
sabberworm: And here…

lib/Sabberworm/CSS/Parser.php
```diff
@@ -64,20 +64,75 @@ public function __construct($sText, Settings $oParserSettings = null, $iLineNo =
...
 	}
 
-	public function setCharset($sCharset) {
-		$this->sCharset = $sCharset;
-		$this->aText = $this->strsplit($this->sText);
-		$this->iLength = count($this->aText);
+	private function fixCharset() {
+		// We need to know the charset before the parsing starts.
+		$sCharset = strtolower($this->oParserSettings->sDefaultCharset);
+		if (strpos($this->sText, "\xef\xbb\xbf") === 0) {
```
sabberworm: Also remove the UTF-16 and -32 BOMs (0xFFFE and 0xFEFF). Throw an exception if the `@charset` declaration does not match the byte order from the DOM.

sabberworm:
Thanks for this pull-request!

I like the approach of moving the whole internal processing to a well-defined charset (UTF-8).

But I really think the parser should, by default and unless the user explicitly requests something else, output the document in its original encoding. This means using iconv at the end to re-convert the output string back to its original encoding and preserving original value of the `@charset` rule.

Also, converting any non-ASCII chars to unicode escapes seems wasteful too me in a day and age where most files are served using UTF-8 and do contain some non-ASCII chars. In my opinion, this should be an option of `OutputFormat` that can be configured to: a) escape all non-ASCII chars, or, b) only escape the characters not representable in the output charset (this should be the default).

Ideally we could add a third option: escape all the chars that were escaped in the input and leave the chars unescaped that appeared verbatim in the input. But since we don’t store that information, this should prove difficult to implement. Maybe we could add a sort of „blacklist“ where the user can configure ranges of characters to always escape even if they would be representable in the output charset.

Also, one of the open issues of the parser is handling UTF-16 files (either with or without BOM). This means the heuristic that searches for the `@charset` rule should also test, as a fallback, if it can find `0'@'0'c'0'h'0'a'0'r'0's'0'e'0't'` (for BE) or `'@'0'c'0'h'0'a'0'r'0's'0'e'0't'0` (for LE) and parse its value. UTF-8, 16-LE, 16-BE should all be covered by at least one test case, each with and without BOM.

skodak:
The problem with returning text in different encoding is that some characters may not be present in other encodings. I have converted only the CSS strings back to escaped form, those are the strings in double quotes (such as the content in font icon definitions - that was where I needed it). If I understand it correctly the identifiers are not touched which means they might be still in unicode. Also UTF-8 is now the default and recommended encoding in PHP, the same goes for CSS.

I am not sure if I ever saw anything encoded in UTF-16, I agree that detecting it and converting it to UTF-8 is possible and easy, on the other had I do not think UTF-16 has any use for web because it is not compatible with ascii - which means no developer or designer should ever try to create CSS in UTF-16. I was more worried about the GB2312 encoding which was required to be used in China, but luckily it appears not to be used much these days.

Anyway I will be very busy with our company project in the next few weeks and then I can work a bit more on this - adding the UTF-16, optional output encoding witch @charset change and some more tests. I'll need to study the CSS spec a bit more I guess.

Thanks a lot for having a detailed look at my patch!

skodak:
hey @FMCorz! I guess you might be interested in this patch...

FMCorz:
Thanks @skodak! Looks like we're solving similar problems, how surprising :). Cheers!

lib/Sabberworm/CSS/Parser.php
skodak: Looks like the charset should be just ignored if there is BOM, see  https://www.w3.org/International/questions/qa-css-charset

skodak: https://www.w3.org/International/questions/qa-html-encoding-declarations.en#utf16 seems to be pretty clear about discouraging UTF-16 and if it is used it must have appropriate BOM header

sabberworm: Let’s throw an exception in strict mode and just ignore the charset (or set appropriately) in lenient mode.

skodak:
Hi, I have reworked the patch, I got much deeper than I wanted originally - the problem was that the strict parsing did not work much for me when I started writing tests for the new UTF stuff - I had to fix some bugs first. When I started stepping through the code in debugger my hand started to hurt from the repeated clicking, so I rewrote some parts to stop calling peek() a million times.

I am not finished yet, I would like to add more test coverage to make sure there are no regressions.

Ciao

lib/Sabberworm/CSS/Parser.php
```diff
@@ -64,20 +71,91 @@ public function __construct($sText, Settings $oParserSettings = null, $iLineNo =
...
 		ksort($this->aSizeUnits, SORT_NUMERIC);
-	}
+		$this->fixCharset();
+	}
+
+	private function fixCharset() {
+		// We need to know the charset before the parsing starts,
+		// UTF BOMs have the highest precedence and must ve removed before other processing.
+		$this->sOriginalCharset = strtolower($this->oParserSettings->sDefaultCharset);
+		if (strpos($this->sText, self::BOM8) === 0) {
```
sabberworm: `strpos` searches the whole string for the BOM. I think we should use `strpos(substr($this->sText, 0, strlen(self::BOM8)), self::BOM8) === 0` for performance. Same for all other checks.

lib/Sabberworm/CSS/Parser.php
```diff
@@ -64,20 +71,91 @@ public function __construct($sText, Settings $oParserSettings = null, $iLineNo =
...
+			$this->aText = preg_split('//u', $this->sText, null, PREG_SPLIT_NO_EMPTY);
+			if (!is_array($this->aText) || count($this->aText) !== $this->iLength) {
+				$this->aText = null;
+			}
+		}
 	}
 
 	public function getCharset() {
-		return $this->sCharset;
+		return 'utf-8';
```
sabberworm: I think we should go the other route here: leave `getCharset` and `setCharset` backwards-compatible by getting and setting the `sOriginalCharset` and add a new method `getProcessingCharset` that returns `'utf-8'`.

This way one can also input a file as `iso-8859-15` and then later use `->setCharset('utf-8')` to declare that no re-encoding should be done when rendered.

`setCharset` should also change the document’s `@charset` rule (if present).

lib/Sabberworm/CSS/Parser.php
```diff
@@ -64,20 +71,91 @@ public function __construct($sText, Settings $oParserSettings = null, $iLineNo =
...
+		} else if (strpos($this->sText, self::BOM32LE) === 0) {
+			$this->sText = substr($this->sText, strlen(self::BOM32LE));
+			$this->sOriginalCharset = 'utf-32le';
+		} else if (strpos($this->sText, self::BOM16BE) === 0) {
+			$this->sText = substr($this->sText, strlen(self::BOM16BE));
+			$this->sOriginalCharset = 'utf-16be';
+		} else if (strpos($this->sText, self::BOM16LE) === 0) {
+			$this->sText = substr($this->sText, strlen(self::BOM16LE));
+			$this->sOriginalCharset = 'utf-16le';
+		} else if (preg_match('/(.*)@charset\s+["\']([a-z0-9-]+)["\']\s*;/ims', $this->sText, $aMatches)) {
```
sabberworm: We should remember whether a BOM was used so we can output it again in `Document::render()`. (Unless `setCharset` was used to set the output charset to a non-UTF charset).

sabberworm:
Thanks.

lib/Sabberworm/CSS/Parser.php
```diff
@@ -64,20 +71,91 @@ public function __construct($sText, Settings $oParserSettings = null, $iLineNo =
...
+			} else {
+				$this->sTextLibrary = 'iconv';
+			}
+		}
+		unset($asciiText);
+		$this->iLength = $this->strlen($this->sText);
+
+		// Substring operations are slower with unicode, aText array is used for faster emulation.
+		if ($this->sTextLibrary !== 'ascii') {
+			$this->aText = preg_split('//u', $this->sText, null, PREG_SPLIT_NO_EMPTY);
```
sabberworm: Shouldn’t we just use the `aText` array for all cases instead of distinguishing between an ASCII and non-ASCII case? Splitting an ASCII string into chars should be pretty cheap and the memory overhead does IMO not warrant having separate code paths (double the code paths means double the number of tests).

lib/Sabberworm/CSS/Parser.php
skodak: ok, makes sense

lib/Sabberworm/CSS/Parser.php
skodak: I disagree, there is no way to convert from unicode to single byte encodings without a data loss, also utf-8 is the recommended encoding for everything on the web, I am not going to implement it, sorry

lib/Sabberworm/CSS/Parser.php
skodak: I disagree here, the presence of BOM is causing problems when processing files in PHP because quite often people forget to ignore it, it is mandatory for utf-16 and utf-32 only

lib/Sabberworm/CSS/Parser.php
skodak: if I understand it correctly the strings in PHP are already stored as arrays of bytes

I did not do much testing after the peek refactoring, also the substrings are now much smaller - so I guess this is not critical for performance any more

I'll try to do more perf testing to see which way is faster

lib/Sabberworm/CSS/Parser.php
skodak: I kept the getChartset() for backwards compatibility because it was telling you in which encoding are the data structures - now it is always utf-8, so all code should work fine as before, I think it is better to keep it this way in case somebody relies on it, alternatively it could be removed completely or at least deprecated

jkrzefski:
Although I don't know so much about this matter, I'd like to see this PR get merged (or refined so far that it becomes mergeable). I had some encoding troubles myself that I could fix using kind of a hack. Also I could really use the performance boost. In the current state I can only use this in combination with a file cache. Any update on this PR?

matzeeable:
This patch looks good. @skodak what exact changes did improve the performance that lot? I am using https://github.com/cweagans/composer-patches and want to add a patch to improve parsing performance. :-)

ThemeMetro:
Is there any update OR solution for this bug?

oliverklee:
@ThemeMetro Someone (TM) would need to take this PR and split it into smaller, focused PRs (with good test coverage). Would you be willing to do this?

JoshuaBehrens:
@skodak is the fork not worth anymore ? we've been waiting for this almost a decade :D

JakeQZ:
Hi @JoshuaBehrens.  As I understand it, the PR provided various performance improvements as well as resolving various issues with charset and BOM handling.  Is there a specific issue you are encountering?  Also, @ThemeMetro, same question.

The changes at the last review can still be viewed on GitHub, so it would be possible to cherry-pick those that are valid and still applicable.  @oliverklee, will the changes always be viewable, or might they disappear?

oliverklee:
> will the changes always be viewable, or might they disappear?

AFAIK, they'll always be viewable in the GitHub UI, and we even can check them out locally (I've just tried) using the `gh` command-line client:

```bash
gh pr checkout 116
```

JoshuaBehrens:
@JakeQZ This post basically summarizes my issue https://github.com/MyIntervals/PHP-CSS-Parser/pull/116#issuecomment-277658481 I can only use this with a caching layer. We once tried to convert CSS to Google AMP compatible CSS with this and things were unbearable slow.

JakeQZ:
Thanks @JoshuaBehrens.  I have some follow-up questions.  There seem to be many small performance improvements in the original PR, so I'd like to establish which ones give the most benefit in real-world use cases.
- Is the slowdown (bottleneck) in the parsing, manipulation, or rendering phase?
- What is your caching layer caching, and how?
- Would you be able to provide sample CSS (which I could use when running a profiler; thus I would need the full lengthy CSS to get meaningful results; can be done under NDA if it's proprietary)?

JoshuaBehrens:
@JakeQZ so the idea was to parse a Shopware 5 compiled theme css. So visit any Shopware 5 shop or host one yourself with dockware https://hub.docker.com/layers/dockware/play/5.5.10/images/sha256-6c8588e9f3d1a2f73daed121092e76a1b677a07453422fb2a9773df68b07f38b and download the compiled css file. Caching layer was eventually file hash of that file. Back then likely not xxhash. Bottleneck likely was parsing IIRC.

We stopped this effort and approach so I have no reference to test against.

PR #124, opened 2017-12-28
Add option to keep comments in output

eternoendless:
At @PrestaShop we want to use this library to help us programmatically transform a bunch of CSS files into right-to-left. Our problem is, the license block included in our CSS files was being lost after the process.

Unless I'm mistaken, the Parser is able to read comments into the syntax tree, but they are always ignored when rendering the output CSS.

The main goal of this change is to make it possible to keep comments in the output CSS, at least the ones located at root level of the original CSS.

With that in mind, this PR adds an option `setKeepComments` in `OutputFormat` which allows toggling this feature on and off (off by default)/

Code is mostly whitespace and comments. I followed PSR-2 on some classes, sorry. Feel free to change it back to your standards if you decide to merge it.

For a clutter-free review I suggest looking at https://github.com/sabberworm/PHP-CSS-Parser/commit/9e23f77d2131fd2f11897ca10c806cf6287eb121, which contains the main change.

### DIFF
--- a/lib/Sabberworm/CSS/Rule/Rule.php
+++ b/lib/Sabberworm/CSS/Rule/Rule.php
@@ -154,16 +154,16 @@ class Rule implements Renderable, Commentable {
 		return $this->bIsImportant;
 	}
 
-    /**
-     * {@inheritdoc}
-     */
+	/**
+	 * {@inheritdoc}
+	 */
 	public function __toString() {
 		return $this->render(new OutputFormat());
 	}
 
-    /**
-     * {@inheritdoc}
-     */
+	/**
+	 * {@inheritdoc}
+	 */
 	public function render(OutputFormat $oOutputFormat) {
 		$sResult = "{$this->sRule}:{$oOutputFormat->spaceAfterRuleName()}";
 		if ($this->mValue instanceof Value) { //Can also be a ValueList
--- a/lib/Sabberworm/CSS/Rule/Rule.php
+++ b/lib/Sabberworm/CSS/Rule/Rule.php
@@ -2,6 +2,8 @@
 
 namespace Sabberworm\CSS\Rule;
 
+use Sabberworm\CSS\Comment\Comment;
+use Sabberworm\CSS\OutputFormat;
 use Sabberworm\CSS\Renderable;
 use Sabberworm\CSS\Value\RuleValueList;
 use Sabberworm\CSS\Value\Value;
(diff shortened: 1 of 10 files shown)


lib/Sabberworm/CSS/RuleSet/DeclarationBlock.php
```diff
@@ -591,15 +593,40 @@ public function createFontShorthand() {
...
+
+		// render comments
+		if ($oOutputFormat->getKeepComments()) {
+			$comments = $this->getCommentsBefore();
+			if (!empty($comments)) {
+				$sResult .= implode(
+					'',
+					array_map(
+						function (Comment $comment) use ($oOutputFormat) {
+							return '/*' . $comment->getComment() . "*/\n";
```
sabberworm: There is already a `render` method on `Comment` that does this. I think you can change that to respect the `KeepComments` option and then you could replace this whole block with

```php
$sResult .= $oOutputFormat->implode('', $this->getCommentsBefore());
```

lib/Sabberworm/CSS/RuleSet/RuleSet.php
```diff
@@ -170,4 +199,21 @@ public function setComments(array $aComments) {
 		$this->aComments = $aComments;
 	}
 
+	/**
+	 * Returns all comments that were declared before this Rule
```
sabberworm: replace “Rule” with “RuleSet”

lib/Sabberworm/CSS/RuleSet/RuleSet.php
```diff
@@ -170,4 +199,21 @@ public function setComments(array $aComments) {
...
+	 * @return Comment[]
+	 */
+	public function getCommentsBefore()
```
sabberworm: I’m not up-to-speed with the comment parsing routines but the way I believe this should work is that the comments associated with this RuleSet would always appear before. If a comment appeared inside, it would be associated with a `Rule`. And if it were to appear below the RuleSet, it would be parsed with the next RuleSet.

I’m not sure, though, how comments are associated that appear in an otherwise empty document or after the last RuleSet. IMHO, they should be stored on the `Document`, not simply discarded.

I guess my question is: why is this method needed, can’t we simply output all comments?

lib/Sabberworm/CSS/Rule/Rule.php
```diff
@@ -152,11 +154,17 @@ public function getIsImportant() {
...
 	public function __toString() {
-		return $this->render(new \Sabberworm\CSS\OutputFormat());
+		return $this->render(new OutputFormat());
 	}
 
-	public function render(\Sabberworm\CSS\OutputFormat $oOutputFormat) {
+	/**
+	 * {@inheritdoc}
+	 */
+	public function render(OutputFormat $oOutputFormat) {
```
sabberworm: A Rule should also output its comments. Maybe we should be able to configure in `OutputFormat` which comments we want to output: only document-level, document-and-ruleset-level, document,-ruleset-and-rule-level.

Issue #79, opened 2014-07-03
Support CSS3 calc()


mikkorantalainen:
Following document throws `Sabberworm\CSS\Parsing\UnexpectedTokenException` with message `Identifier expected. Got “+ 20p”`. This is expected to be valid CSS document:

``` css
body
{
    border-top: solid green 0em;
    border-top-width: calc(1em + 20px);
}
```

The parser is initialized like this:

``` php
$settings = Sabberworm\CSS\Settings::create();
$settings->beStrict();
$parser = new Sabberworm\CSS\Parser($css, $settings);
```

I think `Parser::parseValue()` would need to be modified but I have no idea how to fix this.

bestplay9384:
hey, is there any change calc() handle will be added to the parser?

ivanraz:
Hi there,
We found an issue when using multiplication inside the calc function.

For example, we got this line in our .less file:
right: calc(~"-1*(100% - 54px)");

The parser hits an exception when it goes over the * character.

axute:
is there any news?
we have the same issues with calc()

sabberworm:
I don’t have time to work on this now or in the near future but pull-requests are always welcome.

raxbg:
Can this be closed now?

axute:
we don't use it anymore :-\

sabberworm:
I’ll close this as soon as I make a release.

PR #128, opened 2018-06-28
Add support for CSS3 calc

raxbg:
Fixes issue #79

### DIFF
--- a/lib/Sabberworm/CSS/OutputFormat.php
+++ b/lib/Sabberworm/CSS/OutputFormat.php
@@ -41,6 +41,10 @@ class OutputFormat {
 	// This is what’s printed after the comma of value lists
 	public $sSpaceBeforeListArgumentSeparator = '';
 	public $sSpaceAfterListArgumentSeparator = '';
+
+	// This is what’s printed after the operators in a calc function
+	public $sSpaceBeforeCalcListArgumentSeparator = ' ';
+	public $sSpaceAfterCalcListArgumentSeparator = ' ';
 	
 	public $sSpaceBeforeOpeningBrace = ' ';
 	
@@ -217,6 +221,14 @@ class OutputFormatter {
 
 	public function spaceAfterListArgumentSeparator($sSeparator) {
 		return $this->space('AfterListArgumentSeparator', $sSeparator);
+	}
+
+	public function spaceBeforeCalcListArgumentSeparator($sSeparator) {
+		return $this->space('BeforeCalcListArgumentSeparator', $sSeparator);
+	}
+
+	public function spaceAfterCalcListArgumentSeparator($sSeparator) {
+		return $this->space('AfterCalcListArgumentSeparator', $sSeparator);
 	}
 
 	public function spaceBeforeOpeningBrace() {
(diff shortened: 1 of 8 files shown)


lib/Sabberworm/CSS/Parser.php
```diff
@@ -520,6 +524,17 @@ private function parseURLValue() {
 		return $oResult;
 	}
 
+	private function parseCalcValue() {
+		$func = $this->comes('calc', true) ? 'calc' : '-webkit-calc';
```
sabberworm: You could use `$func = trim(consumeUntil('('))` to streamline this.

lib/Sabberworm/CSS/Parser.php
```diff
@@ -434,6 +436,8 @@ private function parsePrimitiveValue() {
 			$oValue = $this->parseColorValue();
 		} else if ($this->comes('url', true)) {
 			$oValue = $this->parseURLValue();
+		} else if ($this->comes('calc', true) || $this->comes('-webkit-calc', true)) {
```
sabberworm: Support `-moz-calc` as well. Also, is there an `-o-calc`?

lib/Sabberworm/CSS/Parser.php
```diff
@@ -520,6 +524,17 @@ private function parseURLValue() {
 		return $oResult;
 	}
 
+	private function parseCalcValue() {
+		$func = $this->comes('calc', true) ? 'calc' : '-webkit-calc';
+		$this->consume($func);
+		$this->consumeWhiteSpace();
+		$this->consume('(');
+		$aArguments = $this->parseValue(array('+', '-', '*', '/', ' '), '\Sabberworm\CSS\Value\CalcRuleValueList');
```
sabberworm: Parsing the arguments using `parseValue` probably means an expression like `calc(50% -8px)` would be treated as `calc(50% - 8px)`, not as an error (like it should).

sabberworm: Am I right in assuming that this does not support parenthesized subexpressions like `calc( ( 100px / 2) / 2)`?

lib/Sabberworm/CSS/OutputFormat.php
```diff
@@ -41,6 +41,10 @@ class OutputFormat {
 	// This is what’s printed after the comma of value lists
 	public $sSpaceBeforeListArgumentSeparator = '';
 	public $sSpaceAfterListArgumentSeparator = '';
+
+	// This is what’s printed after the operators in a calc function
+	public $sSpaceBeforeCalcListArgumentSeparator = ' ';
```
sabberworm: Since the spaces around `-` and `+` operators are mandatory, their output should not be affected by `OutputFormat`.

lib/Sabberworm/CSS/Parser.php
raxbg: Done for `-moz-calc`, but `-o-calc` does not seem to exist https://www.w3schools.com/cssref/func_calc.asp

lib/Sabberworm/CSS/Parser.php
raxbg: Correct for both. Should be fixed now :)

Issue #131, opened 2018-07-11
Support for grid line names


raxbg:
These rule values are wrapped in `[]` which is currently not supported.
Example: 
```CSS
grid-template-columns: [linename1] 100px [linename2 linename3];
```
MDN Doc: https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-columns

PR #132, opened 2018-07-11
Add support for grid line names

raxbg:
Fix proposal for #131

### DIFF
--- a/lib/Sabberworm/CSS/Parser.php
+++ b/lib/Sabberworm/CSS/Parser.php
@@ -485,19 +485,20 @@ class Parser {
 
 	private function parseLineNameValue() {
 		$this->consume('[');
-		$sName = '';
-		try {
-			do {
-				$this->consumeWhiteSpace();
-				$sName .= $this->parseIdentifier(false, true) . ' ';
-			} while (!$this->comes(']'));
-		} catch (UnexpectedTokenException $e) {
-			if(!$this->oParserSettings->bLenientParsing && (!$sName || !$this->comes(']'))) {// This handles constructs like this [ linename ] in non lenient mode
-				throw $e;
+		$this->consumeWhiteSpace();
+		$aNames = array();
+		do {
+			if($this->oParserSettings->bLenientParsing) {
+				try {
+					$aNames[] = $this->parseIdentifier(false, true);
+				} catch(UnexpectedTokenException $e) {}
+			} else {
+				$aNames[] = $this->parseIdentifier(false, true);
 			}
-		}
+			$this->consumeWhiteSpace();
+		} while (!$this->comes(']'));
 		$this->consume(']');
-		return new LineName(rtrim($sName), $this->iLineNo);
+		return new LineName($aNames, $this->iLineNo);
 	}
 
 	private function parseColorValue() {
--- a/lib/Sabberworm/CSS/Parser.php
+++ b/lib/Sabberworm/CSS/Parser.php
@@ -21,6 +21,7 @@ use Sabberworm\CSS\Value\Size;
 use Sabberworm\CSS\Value\Color;
 use Sabberworm\CSS\Value\URL;
 use Sabberworm\CSS\Value\CSSString;
+use Sabberworm\CSS\Value\LineName;
 use Sabberworm\CSS\Rule\Rule;
 use Sabberworm\CSS\Parsing\UnexpectedTokenException;
 use Sabberworm\CSS\Comment\Comment;
@@ -447,6 +448,8 @@ class Parser {
 			$oValue = $this->parseStringValue();
 		} else if ($this->comes("progid:") && $this->oParserSettings->bLenientParsing) {
 			$oValue = $this->parseMicrosoftFilter();
+		} else if ($this->comes("[")) {
+			$oValue = $this->parseLineNameValue();
 		} else {
 			$oValue = $this->parseIdentifier(true, false);
 		}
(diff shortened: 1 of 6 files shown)


lib/Sabberworm/CSS/Parser.php
```diff
@@ -480,6 +483,16 @@ private function parseNumericValue($bForColor = false) {
 		return new Size(floatval($sSize), $sUnit, $bForColor, $this->iLineNo);
 	}
 
+	private function parseLineNameValue() {
+		$this->consume('[');
+		$sName = '';
```
sabberworm: Use `consumeWhitespace` after `[` and before `]`.

lib/Sabberworm/CSS/Parser.php
```diff
@@ -480,6 +483,16 @@ private function parseNumericValue($bForColor = false) {
...
+		while(!$this->comes(']')) {
+			$sName .= $this->consume(1);
```
sabberworm: Reading https://developer.mozilla.org/en-US/docs/Web/CSS/custom-ident it seems we should support certain escape sequences. We could use `parseCharacter(true)` for this.

lib/Sabberworm/CSS/Parser.php
raxbg: Should okay now. Please check.

lib/Sabberworm/CSS/Parser.php
raxbg: Should okay now. Please check.

lib/Sabberworm/CSS/Parser.php
```diff
@@ -480,6 +483,23 @@ private function parseNumericValue($bForColor = false) {
...
+	private function parseLineNameValue() {
+		$this->consume('[');
+		$sName = '';
+		try {
+			do {
+				$this->consumeWhiteSpace();
+				$sName .= $this->parseIdentifier(false, true) . ' ';
+			} while (!$this->comes(']'));
+		} catch (UnexpectedTokenException $e) {
+			if(!$this->oParserSettings->bLenientParsing && (!$sName || !$this->comes(']'))) {// This handles constructs like this [ linename ] in non lenient mode
```
sabberworm: Lenient mode has always tried to mimic what the browser does when parsing. I think if the browser encounters an unparseable value, it throws away the whole rule so we should do the same. Since there is a handler for `UnexpectedTokenException` already in place that does that, we don’t need to catch it in lenient mode.

Then again I’m not sure what the other half of the `if` (`!$sName || !$this->comes(']')`) is for. The comment does not enlighten me much. Shouldn’t `[ linename ]` parse correctly without throwing an exception? There’s something I don’t understand.

raxbg:
> Lenient mode has always tried to mimic what the browser does when parsing. I think if the browser encounters an unparseable value, it throws away the whole rule so we should do the same.

`[]` is parsed as `[]` in Chrome. If the `UnexpectedTokenException` is not handled here the parser will drop the value for the rule.

>Then again I’m not sure what the other half of the if (!$sName || !$this->comes(']')) is for.

On the second iteration of the loop, `parseIdentifier()` will try to parse the `]`, so it will throw an exception. However this should not be considered an error because `$sName` will not be empty (it will have the value of '**linename** '). The exception should be forwarded only if `$sName` is empty and lenient mode is disabled. This will be the case if parsing `[]`.

lib/Sabberworm/CSS/Parser.php
raxbg: > Lenient mode has always tried to mimic what the browser does when parsing. I think if the browser encounters an unparseable value, it throws away the whole rule so we should do the same.

`[]` is parsed as `[]` in Chrome. If the `UnexpectedTokenException` is not handled here the parser will drop the value for the rule.

>Then again I’m not sure what the other half of the if (!$sName || !$this->comes(']')) is for.

On the second iteration of the loop, `parseIdentifier()` will try to parse the `]`, so it will throw an exception. However this should not be considered an error because `$sName` will not be empty (it will have the value of '**linename** '). The exception should be forwarded only if `$sName` is empty and lenient mode is disabled. This will be the case if parsing `[]`.

sabberworm: I still think using exception handling for control flow should be avoided. Having read your comments, shouldn’t the following also work?:

```php
private function parseLineNameValue() {
	$this->consume('[');
	$this->consumeWhiteSpace();
	$sName = '';
	do {
		if($this->oParserSettings->bLenientParsing) {
			try {
				$sName .= $this->parseIdentifier(false, true) . ' ';
			} catch(UnexpectedTokenException $e) {}
		} else {
			$sName .= $this->parseIdentifier(false, true) . ' ';
		}
		$this->consumeWhiteSpace();
	} while (!$this->comes(']'));
	$this->consume(']');
	return new LineName(rtrim($sName), $this->iLineNo);
}
```

Having said that: now that I know there can be multiple identifiers in line name value brackets, I think `LineName` should extend `ValueList` with `$sSeparator = ' '` and thus the code would become:

```php
private function parseLineNameValue() {
	$this->consume('[');
	$this->consumeWhiteSpace();
	$aNames = array();
	do {
		if($this->oParserSettings->bLenientParsing) {
			try {
				$aNames[] = $this->parseIdentifier(false, true);
			} catch(UnexpectedTokenException $e) {}
		} else {
			$aNames[] = $this->parseIdentifier(false, true);
		}
		$this->consumeWhiteSpace();
	} while (!$this->comes(']'));
	$this->consume(']');
	return new LineName($aNames, $this->iLineNo);
}
```

which would allow you to shorten `LineName` to the following:

```php
class LineName extends ValueList {
	public function __construct($aComponents = array(), $iLineNo = 0) {
		parent::__construct($aComponents, ' ', $iLineNo);
	}

	public function __toString() {
		return $this->render(new \Sabberworm\CSS\OutputFormat());
	}

	public function render(\Sabberworm\CSS\OutputFormat $oOutputFormat) {
		return '['.parent::render(\Sabberworm\CSS\OutputFormat::createCompact()).']';
	}
}
```

Optionally, you could create OutputFormat options on whether or not white-space should be output after `[` and before `]`.

raxbg: Love it ❤️

sabberworm:
Thanks a lot @raxbg !

PR #135, opened 2018-07-13
how to add rule set in rule

papnejavicky:
Please give me full  documentation of css parser.
Also, How to use addrule set function

Please mail me vpspapneja@gmail.com/papnejavicky@gmail.com


sabberworm:
@papnejavicky That’s not how pull-requests work.

I agree with you the documetation is sparse but I don’t plan on actively improving it because I don’t have the resources. But if someone else wants to do that, great.

Doxygen documentation is available at https://sabberworm.github.io/PHP-CSS-Parser/

PR #138, opened 2018-07-25
Fix parsing CSS selectors which contain commas

westonruter:
I found an issue where a CSS selectors would fail to parse correctly when they contain commas. For example:

```css
.widget:not(.widget_text, .jetpack_widget_social_icons) ul {
    list-style: none;
    margin: 0;
    padding: 0;
}
```

There should only be one selector here: `.widget:not(.widget_text, .jetpack_widget_social_icons) ul`. Instead, the parser incorrectly extracts two:

* `.widget:not(.widget_text`
* `.jetpack_widget_social_icons) ul`

The fix is to add placeholders for expressions contained in parentheses or brackets prior to splitting, and then replace the placeholders with their original values on the split selectors.

### DIFF
--- a/lib/Sabberworm/CSS/RuleSet/DeclarationBlock.php
+++ b/lib/Sabberworm/CSS/RuleSet/DeclarationBlock.php
@@ -28,13 +28,71 @@ class DeclarationBlock extends RuleSet {
 		if (is_array($mSelector)) {
 			$this->aSelectors = $mSelector;
 		} else {
-			$this->aSelectors = explode(',', $mSelector);
+			list( $sSelectors, $aPlaceholders ) = $this->addSelectorExpressionPlaceholders( $mSelector );
+			if ( empty( $aPlaceholders ) ) {
+				$this->aSelectors = explode(',', $sSelectors);
+			} else {
+				$aSearches = array_keys( $aPlaceholders );
+				$aReplaces = array_values( $aPlaceholders );
+				$this->aSelectors = array_map(
+					function( $sSelector ) use ( $aSearches, $aReplaces ) {
+						return str_replace( $aSearches, $aReplaces, $sSelector );
+					},
+					explode(',', $sSelectors)
+				);
+			}
 		}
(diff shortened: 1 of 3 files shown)


westonruter:
@sabberworm Anything else you'd like to see with this PR?

sabberworm:
Thanks for the input @westonruter. Sorry for the long response time, I had been away on holiday.

While this does solve one of the long-standing gripes I’ve had with selector parsing, it does not solve all of them and I don’t think it’s a particularly elegant solution and it’s not the direction I want to pursue.

What I would like to see is to see the `Selector` class expanded so becomes an interface with a subclass for every type of (combining) selector there is, so that every selector is stored as an instance instead of a string.

So the selector `.a + :not(#b) + [c] ~ foo|.d span > *:root`, would decompose into:

```php
new AdjacentSiblingCombinator(
	new ClassSelector('a'),
	new Negation(
		new IDSelector('b')
	),
	new GeneralSiblingCombinator(
		new AttributeSelector('c'),
		new DescendantCombinator(
			new NamespaceSelector(
				'foo',
				new ClassSelector('d')
			),
			new ChildCombinator(
				new TypeSelector('span'),
				new Combinator(
					new UniversalSelector(),
					new RootSelector()
				)
			)
		)
	)
);
```

Note: this is just a suggestion. Maybe the `Combinator` could be replaced with a simple array and maybe it’s not a good idea to let the same combinator have more than two elements (like the `AdjacentSiblingCombinator`). And maybe we don’t need to nest them, just store combinator and simple selector in an array side-by-side.

However what this requires is actually parsing and understanding the selector. And I think string substitution, even when done right, is a poor substitute (pun intended) for that.

Actually parsing selectors would finally also fix the following:

```css
.some[selectors-may='contain-a-{'] {}
```

which parses neither with nor without your fix.

westonruter:
Thanks. You're right. My fix is more of a bandaid than a permanent solution. I wasn't sure that you intended to actually have the selector themselves parsed as well. That being the case, then indeed my fix isn't ideal.

PR #140, opened 2018-08-21
Bug fix in calc parsing when the first operand is a negative value

raxbg:
These changes fix issues with rules like `calc(-120% - 4px)` and `calc(8px + -50%)`

### DIFF
--- a/lib/Sabberworm/CSS/Parser.php
+++ b/lib/Sabberworm/CSS/Parser.php
@@ -555,7 +555,7 @@ class Parser {
 	}
 
 	private function parseCalcValue() {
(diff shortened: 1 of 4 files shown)


sabberworm:
Thanks. Can you also test for nested `calc`s? Both with braces or `calc` functions?

```css
.test {
  font-size: calc((3 * 4px) + -2px);
  top: calc(200px - calc(20 * 3px));
}
```

raxbg:
I remember that I tested these cases while I was working on the changes, but I must have forgotten to add unit tests for them. Should be ok now.

PR #142, opened 2018-09-04
Suppress "Unexpected end of document" in lenient mode

raxbg:
Browsers seem to be okay with incomplete documents, they just parse up to the last valid rule. I guess we should do the same in lenient mode.


sabberworm:
LGTM

PR #144, opened 2018-09-24
Unicode range

raxbg:
The unicode-range property (https://developer.mozilla.org/en-US/docs/Web/CSS/@font-face/unicode-range) is currently not supported. This PR adds support for that.


sabberworm:
I’m not quite on board with parsing these as literals. I guess a custom `Value` subclass `UnicodeRange` would be more appropriate. It could then also provide methods to check if a code point is included in the range…

raxbg:
Hm, I thought these were supposed to always be literals. Having a separate `Value` subclass sounds good, but honestly I am not sure when will I have time to do it. I was able to update this PR since the old code was reusable. As of now, this PR at least helps to keep the unicode ranges in the rendered CSS, otherwise these rules get removed.

It would be great if you(or anybody) can add the `UnicodeRange` type and complete the implementation, if not I will ask you to at least merge this PR as is and open an issue for the suggested improvements.

Thanks!

sabberworm:
I’ll merge it as-is, thanks.

PR #149, opened 2018-11-19
Add prepend and replace methods

raxbg:
This PR adds `prepend` and `replace` methods.

Possible use case for `prepend` is to allow adding `@import` rules, since they must come first. 

Possible use case for `replace` is changing the charset.


sabberworm:
LGTM, thanks!

PR #154, opened 2019-01-18
Fix undefined index notice when property lacks value

westonruter:
Given this CSS:

```css
div {
	display: inline-block;
	display:
}
```

I'm getting a PHP notice:

```
PHP Notice:  Undefined offset: 0 in .../sabberworm/php-css-parser/lib/Sabberworm/CSS/Value/Value.php on line 58
```

This PR fixes that problem by improving the resiliency to bad CSS.

### DIFF
--- a/lib/Sabberworm/CSS/Value/Value.php
+++ b/lib/Sabberworm/CSS/Value/Value.php
@@ -3,6 +3,7 @@
 namespace Sabberworm\CSS\Value;
 
 use Sabberworm\CSS\Parsing\ParserState;
+use Sabberworm\CSS\Parsing\UnexpectedTokenException;
 use Sabberworm\CSS\Renderable;
 
 abstract class Value implements Renderable {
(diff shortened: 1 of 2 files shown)


lib/Sabberworm/CSS/Value/Value.php
```diff
@@ -55,6 +55,9 @@ public static function parseValue(ParserState $oParserState, $aListDelimiters =
 				array_splice($aStack, $iStartPosition - 1, $iLength * 2 - 1, array($oList));
 			}
 		}
+		if (!isset($aStack[0])) {
+			return null;
```
sabberworm: Could you try to instead to throw a custom `UnexpectedTokenException` which, in lenient mode, should be [caught](https://github.com/sabberworm/PHP-CSS-Parser/blob/master/lib/Sabberworm/CSS/RuleSet/RuleSet.php#L35) already.

westonruter:
@sabberworm thanks for the review. I've added the throwing of `UnexpectedTokenException` in 2c61b73 along with tests.

PR #165, opened 2019-06-27
Fix/size render

raxbg:
There are cases where people add very big numbers for z-index (e.g 9999999999999999999999), which renders in scientific notation and becomes invalid to the browsers. This PR fixes how large numbers are rendered. If a number is not going to be rendered using scientific notation, the library behaves as before.


lib/Sabberworm/CSS/Value/Size.php
```diff
@@ -116,7 +116,8 @@ public function __toString() {
 	public function render(\Sabberworm\CSS\OutputFormat $oOutputFormat) {
 		$l = localeconv();
 		$sPoint = preg_quote($l['decimal_point'], '/');
-		return preg_replace(array("/$sPoint/", "/^(-?)0\./"), array('.', '$1.'), $this->fSize) . ($this->sUnit === null ? '' : $this->sUnit);
+		$sSize = preg_match("/[\d\.]+e[+-]?\d+/i", (string)$this->fSize) ? preg_replace("/$sPoint?0+$/", "", sprintf("%f", $this->fSize)) : $this->fSize;
```
sabberworm: Could we maybe simplify this to always use the sprintf code path?

raxbg: We could, but it gets more complicated. As far as I remember when I initially tried it I had an issue with keeping the original number values in the output. The problem was that we have to at least know the precision of the original number in order to try to output the same thing. Otherwise 0.1 might turn out as 0.099999999. I am also not sure what are the implications of that (other than the tests start to fail), but I'd rather keep the current behavior for everything that is not printed in scientific notation, and only convert the rest of the numbers.

sabberworm: OK, I simply worried about the `preg_match` not covering all cases and stuff becoming more difficult when we have two code paths.

PR #174, opened 2019-11-12
Parse escaped URL and Calc functions correctly

### LINKED ISSUE #170: Function name with escaping not properly recognized
e.g.

`body {
  background: ur\l("//example.org/picture.jpg");
}`

the url-function is not parsed into a URL object but into a CSSFunction object, even though the function name is normalized to 'url'

raxbg:
Resolves #170. Thanks to @mtsvetkov for this PR.

### DIFF
--- a/lib/Sabberworm/CSS/Value/Value.php
+++ b/lib/Sabberworm/CSS/Value/Value.php
@@ -68,7 +68,15 @@ abstract class Value implements Renderable {
 		if ($oParserState->comes('(')) {
 			$oParserState->consume('(');
 			$aArguments = Value::parseValue($oParserState, array('=', ' ', ','));
-			$sResult = new CSSFunction($sResult, $aArguments, ',', $oParserState->currentLine());
+
+			if ($sResult == 'url') {
+				$sResult = new URL($aArguments, $oParserState->currentLine());
+			} else if ($sResult == 'calc') {
+				$sResult = new CalcFunction($sResult, $aArguments, ',', $oParserState->currentLine());
+			} else {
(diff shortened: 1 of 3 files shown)


lib/Sabberworm/CSS/Value/Value.php
```diff
@@ -68,7 +68,15 @@ public static function parseIdentifierOrFunction(ParserState $oParserState, $bIg
 		if ($oParserState->comes('(')) {
 			$oParserState->consume('(');
 			$aArguments = Value::parseValue($oParserState, array('=', ' ', ','));
-			$sResult = new CSSFunction($sResult, $aArguments, ',', $oParserState->currentLine());
+
+			if ($sResult == 'url') {
+				$sResult = new URL($aArguments, $oParserState->currentLine());
+			} else if ($sResult == 'calc') {
+				$sResult = new CalcFunction($sResult, $aArguments, ',', $oParserState->currentLine());
```
sabberworm: Make sure we support `-webkit-calc`, `-moz-calc`, etc here too.

lib/Sabberworm/CSS/Value/Value.php
```diff
@@ -68,7 +68,15 @@ public static function parseIdentifierOrFunction(ParserState $oParserState, $bIg
...
+			} else {
+				$sResult = new CSSFunction($sResult, $aArguments, ',', $oParserState->currentLine());
+			}
+
```
sabberworm: Can we unify this with the logic from `parsePrimitiveValue` (i.e. drop the duplicate code and `URL::parse` and `CalcFunction::parse`)?

PR #176, opened 2019-11-12
Improve EOF handling + handle lonely import statements

### LINKED ISSUE #175: Missing ';' breaks @import parsing
`@import` parsing breaks if defined as follows:

```CSS
@import "example.css"
```

raxbg:
Resolves #175


lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -110,15 +111,34 @@ private static function parseAtRule(ParserState $oParserState) {
...
+				}
+				$oParserState->consume(';');
+			} catch (UnexpectedEOFException $e) {
+				// Save the media query if present
+				$sMediaQuery = '';
+				try {
+					while (!$oParserState->isEnd()) {
+						$sMediaQuery .= $oParserState->consume(1);
+					}
+				} catch (UnexpectedEOFException $e) {}
```
sabberworm: Can this exception ever be thrown? I would assume not since you’re checking for `!$oParserState->isEnd()`.

sabberworm:
Codacy Here is an overview of what got changed by this pull request:


```diff

Complexity increasing per file
==============================
- tests/Sabberworm/CSS/ParserTest.php  14
- lib/Sabberworm/CSS/Parsing/ParserState.php  11
- lib/Sabberworm/CSS/CSSList/CSSList.php  22
         
```


See the complete overview on [Codacy](https://app.codacy.com/app/sabberworm/PHP-CSS-Parser/pullRequest?prid=4484965&bid=15131748)

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -110,15 +111,34 @@ private static function parseAtRule(ParserState $oParserState) {
...
+
+				$sMediaQuery = trim($sMediaQuery);
```
sabberworm: I prefer calling `consumeWhiteSpace` before and after reading the media query instead of trimming afterwards.

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -110,15 +111,34 @@ private static function parseAtRule(ParserState $oParserState) {
 			$oLocation = URL::parse($oParserState);
 			$oParserState->consumeWhiteSpace();
 			$sMediaQuery = null;
-			if (!$oParserState->comes(';')) {
-				$sMediaQuery = $oParserState->consumeUntil(';');
+			try {
+				if (!$oParserState->comes(';')) {
+					$sMediaQuery = $oParserState->consumeUntil(';');
```
sabberworm: Maybe we can add an optional parameter to `consumeUntil` that allows reading till either the desired char or EOF, whichever comes first. Would clean up the code somewhat and also resolve using exceptions for control flow.

lib/Sabberworm/CSS/CSSList/CSSList.php
sabberworm: Can this exception ever be thrown? I would assume not since you’re checking for `!$oParserState->isEnd()`.

lib/Sabberworm/CSS/Parsing/ParserState.php
```diff
@@ -212,19 +213,25 @@ public function consumeUntil($aEnd, $bIncludeEnd = false, $consumeEnd = false, a
...
 		$start = $this->iCurrentPosition;
 
-		while (($char = $this->consume(1)) !== '') {
-			if (in_array($char, $aEnd)) {
-				if ($bIncludeEnd) {
-					$out .= $char;
-				} elseif (!$consumeEnd) {
-					$this->iCurrentPosition -= $this->strlen($char);
+		try {
+			while (($char = $this->consume(1)) !== '') {
```
sabberworm: Is there any situation where `consume(1)` would return the empty string? This condition seems like it’s left over from a previous refactoring…

lib/Sabberworm/CSS/CSSList/CSSList.php
sabberworm: Or simply add a sentinel value (e.g. `ParserState::EOL = 'EOL'`) that can be used like a regular token:

```php
$sMediaQuery = $oParserState->consumeUntil(array(';', ParserState::EOL));
```

raxbg: I agree, however `consumeUntil` takes a lot of parameters already, so I added a new constant `ParserState::EOF` which can be passed as the first argument of `consumeUntil`. Also helps with readability.

lib/Sabberworm/CSS/Parsing/ParserState.php
```diff
@@ -216,7 +217,8 @@ public function consumeUntil($aEnd, $bIncludeEnd = false, $consumeEnd = false, a
 		$start = $this->iCurrentPosition;
 
 		try {
-			while (($char = $this->consume(1)) !== '') {
+			while (true) {
+				$char = $this->consume(1);
```
sabberworm: Maybe test for `isEnd` and handle the `in_array(self::EOF, $aEnd)` case gracefully instead of relying on the exception handling here.

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -111,14 +112,20 @@ private static function parseAtRule(ParserState $oParserState) {
...
 			}
-			$oParserState->consume(';');
 			return new Import($oLocation, $sMediaQuery, $iIdentifierLineNum);
 		} else if ($sIdentifier === 'charset') {
 			$sCharset = CSSString::parse($oParserState);
-			$oParserState->consumeWhiteSpace();
-			$oParserState->consume(';');
+			try {
+				$oParserState->consumeWhiteSpace();
+				$oParserState->consume(';');
```
sabberworm: Use the `EOL` marker here too.

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -136,7 +143,11 @@ private static function parseAtRule(ParserState $oParserState) {
 				$sPrefix = $mUrl;
 				$mUrl = Value::parsePrimitiveValue($oParserState);
 			}
-			$oParserState->consume(';');
+			try {
+				$oParserState->consume(';');
```
sabberworm: As well as here.

lib/Sabberworm/CSS/CSSList/CSSList.php
raxbg: LoL, sorry for the repeat, I posted my reply without realizing you had a second post. I set the value of `ParserState::EOF` to null though. Do you think this can cause issues?

sabberworm: It seems like we had the same idea…

I think any value that isn’t a valid return value from `consume(1)` is fine so `null` should work.

lib/Sabberworm/CSS/CSSList/CSSList.php
sabberworm: Sorry… when using `consumeUntil(';')` this will of course include trailing whitespace if any so you do need `trim` after all…

lib/Sabberworm/CSS/Parsing/ParserState.php
```diff
@@ -229,6 +232,10 @@ public function consumeUntil($aEnd, $bIncludeEnd = false, $consumeEnd = false, a
 			}
 		}
 
+		if (in_array(self::EOF, $aEnd)) {
+			return $out;
+		}
+
 		$this->iCurrentPosition = $start;
 		throw new UnexpectedTokenException('One of ("'.implode('","', $aEnd).'")', $this->peek(5), 'search', $this->iLineNo);
```
sabberworm: Maybe use `UnexpectedEOFException` here too…

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -111,14 +112,18 @@ private static function parseAtRule(ParserState $oParserState) {
 			$oParserState->consumeWhiteSpace();
 			$sMediaQuery = null;
 			if (!$oParserState->comes(';')) {
-				$sMediaQuery = $oParserState->consumeUntil(';');
+				$sMediaQuery = trim($oParserState->consumeUntil(array(';', ParserState::EOF)));
+			}
+			if (!$oParserState->isEnd()) {
+				$oParserState->consume(';');
 			}
```
sabberworm: I think you could remove these lines if you set `consumeEnd` to `true` in the call to `consumeUntil` (at least it’s supposed to work this way).

PR #180, opened 2019-12-09
Fix comments being omitted in the middle of a rule

scq:
Comments in the middle of a rule (i.e. between two properties) don't make it to the parsed AST.

This used to work but stopped working in c468e71


scq:
Closing in favour of #339

PR #187, opened 2020-02-19
Removed vm unit from ABSOLUTE_SIZE_UNITS

hkarlstrom:
Currently e.g. `1vmax` is rendered as `1vm ax`. Removing `vm` from ABSOLUTE_SIZE_UNITS fixes that issue. Afaik there is no such unit as `vm`.


raxbg:
Hi,

This is already fixed in master via #167

PR #192, opened 2020-04-04
Test enhancement

peter279k:
# Changed log
- Add `php-7.4` version test during Travis CI build.
- Let `composer.lock` be removed on Git version control and it can avoid caching dependencies.
- Using the `PHPUnit ^4.8.36` version to support future PHPUnit versions.
- Using the `\PHPUnit\Framework\TestCase` namespace to be compatible with future `PHPUnit` versions.
- Using the `assertCount` to assert expected count is same as result count.
- Using `assertTrue` or `assertFalse` to assert that expected value is `true` or `false`.


.gitignore
```diff
@@ -1 +1,2 @@
 vendor/
+composer.lock
```
sabberworm, line 2: I think the lock file should remain in the repo, otherwise we’ll get different versions on each install.

peter279k: Hi @sabberworm, thanks for your reply.

Yes. You're right. And I've reverted this :).

PR #193, opened 2020-05-06
Add additional validation for size unit

pierlon:
This PR adds additional validation when parsing a size unit.

With this change, when parsing in strict mode an `UnexpectedTokenException` error will be thrown when an invalid size unit occurs after a number. When in lenient mode, however, if a valid size unit identifier is detected, it will be kept and the proceeding set of characters will be stripped. For example:

  | Dimension value | Strict parsing | Lenient parsing
---|---|---|---
**Before** | `1radsss` | `1 rad sss` | `1 rad sss`
**After** | `1radsss` | `UnexpectedTokenException` | Value is removed

Also, there is no such unit named `turns`, but there is one called `turn` ([ref](https://drafts.csswg.org/css-values-3/#turn)), which I suppose is what was meant here :smile:.

### DIFF
--- a/lib/Sabberworm/CSS/Value/Size.php
+++ b/lib/Sabberworm/CSS/Value/Size.php
@@ -38,26 +38,21 @@ class Size extends PrimitiveValue {
 		}
 
 		$sUnit = null;
-		$sParsedUnit = null;
 		$aSizeUnits = self::getSizeUnits();
-
 		$iMaxSizeUnitLength = max(array_keys($aSizeUnits));
-		if ( preg_match( '/^[a-zA-Z0-9%]+/', $oParserState->peek($iMaxSizeUnitLength), $matches ) ) {
-			$sParsedUnit = $matches[0];
-		}
 
-		foreach($aSizeUnits as $iLength => $aValues) {
-			$sKey = strtolower($oParserState->peek($iLength));
-			if(array_key_exists($sKey, $aValues)) {
-				if (strtolower($sParsedUnit) !== $sKey) {
-					throw new UnexpectedTokenException('Unit', $sParsedUnit, 'identifier', $oParserState->currentLine());
-				}
-				if (($sUnit = $aValues[$sKey]) !== null) {
-					$oParserState->consume($iLength);
-					break;
-				}
+		if ( preg_match( '/^[a-zA-Z0-9%]+/', $oParserState->peek($iMaxSizeUnitLength), $matches ) ) {
+			$sUnit = strtolower($matches[0]);
+			$iUnitLength = strlen($sUnit);
+
+			if (isset($aSizeUnits[$iUnitLength][$sUnit])) {
+				$sUnit = $aSizeUnits[$iUnitLength][$sUnit];
+				$oParserState->consume($iUnitLength);
+			} else {
+				throw new UnexpectedTokenException('Unit', $sUnit, 'identifier', $oParserState->currentLine());
 			}
 		}
+
 		return new Size(floatval($sSize), $sUnit, $bIsColorComponent, $oParserState->currentLine());
 	}
 
--- a/lib/Sabberworm/CSS/Value/Size.php
+++ b/lib/Sabberworm/CSS/Value/Size.php
@@ -41,7 +41,7 @@ class Size extends PrimitiveValue {
 		$aSizeUnits = self::getSizeUnits();
 		$iMaxSizeUnitLength = max(array_keys($aSizeUnits));
 
-		if ( preg_match( '/^[a-zA-Z0-9%]+/', $oParserState->peek($iMaxSizeUnitLength), $matches ) ) {
+		if ( preg_match( '/^(%|[a-zA-Z0-9]+)/', $oParserState->peek($iMaxSizeUnitLength), $matches ) ) {
 			$sUnit = strtolower($matches[0]);
 			$iUnitLength = strlen($sUnit);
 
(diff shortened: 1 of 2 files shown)


pierlon:
That failing test seems to be unrelated to the change here.

westonruter:
Yeah, if I try [validating](https://jigsaw.w3.org/css-validator/validator):

```css
@keyframes spin {
  to {
    transform: rotate(1turns);
  }
}
```

Then it fails with `Unknown dimension: 1turns`

lib/Sabberworm/CSS/Value/Size.php
```diff
@@ -36,13 +37,39 @@ public static function parse(ParserState $oParserState, $bIsColorComponent = fal
...
+		foreach($aSizeUnits as $iLength => $aValues) {
+		    $iConsumeLength = $iLength;
 			$sKey = strtolower($oParserState->peek($iLength));
 			if(array_key_exists($sKey, $aValues)) {
+			    if (strtolower($sParsedUnit) !== $sKey) {
+                    if (!$oParserState->getSettings()->bLenientParsing) {
+                        throw new UnexpectedTokenException('Unit', $sParsedUnit, 'identifier', $oParserState->currentLine());
+                    }
+                    $iConsumeLength = strlen($sParsedUnit);
+                }
```
westonruter, line 70: This appears to be using spaces for indentation rather than tabs.

tests/Sabberworm/CSS/ParserTest.php
```diff
@@ -766,4 +766,22 @@ function testLonelyImport() {
 		$sExpected = "@import url(\"example.css\") only screen and (max-width: 600px);";
 		$this->assertSame($sExpected, $oDoc->render());
 	}
+
+    function testTurnUnitLenient() {
+        $sText = ".foo {transform: rotate(1turn);}\n.bar {transform: rotate(1turns);}";
+        $sExpected = ".foo {transform: rotate(1turn);}\n.bar {transform: rotate(1turn);}";
```
westonruter, line 772: Question: Rather than truncating the unit to the part that is invalid, should not the property just be removed entirely?

For example, If I try using the `turns` unit, it is rejected entirely in Chrome:

image

So I think this should actually rather be:

```suggestion
        $sExpected = ".foo {}\n.bar {}";
```

lib/Sabberworm/CSS/Value/Size.php
pierlon: Addressed in https://github.com/sabberworm/PHP-CSS-Parser/pull/193/commits/54ae4fe617dafc6771db717cf85d986d4300275a.

tests/Sabberworm/CSS/ParserTest.php
pierlon: Ah right, that makes sense. I've addressed this in https://github.com/sabberworm/PHP-CSS-Parser/pull/193/commits/54ae4fe617dafc6771db717cf85d986d4300275a.

lib/Sabberworm/CSS/Value/Size.php
```diff
@@ -36,11 +37,33 @@ public static function parse(ParserState $oParserState, $bIsColorComponent = fal
...
+			if (($iPeek >= 97 && $iPeek <= 122) ||
+				($iPeek >= 65 && $iPeek <= 90) ||
+				($iPeek >= 48 && $iPeek <= 57) ||
+				($iPeek === 37)) {
+				$sParsedUnit .= $sChar;
+				$iOffset++;
+			} else {
+				break;
+			}
+		}
```
pierlon, line 56: I'm not entirely comfortable with this approach of getting the unit; better solutions are welcomed.

westonruter: Instead of a `while` loop, another option would be to peek 10 chars and then use `preg_match()` to match the range, and if matched then capture that as the `$sParsedUnit` and increase the `$iOffset` by the length.

```
if ( preg_match( '/^[a-zA-Z0-9%]+/', $oParserState->peek(10, $iOffset), $matches ) ) {
    $sParsedUnit = $matches[0];
    $iOffset += strlen( $matches[0] );
}
```

pierlon: Any reason behind the 10 characters? Or is it that a unit length of 10 characters or more would be unlikely?

westonruter: Good point. It could be set to the max length of the unit chars.

westonruter:
@sabberworm Anything else you'd like to see here?

sabberworm:
I’m just wondering if we really need to hard-code all possible units. To be future-proof, couldn’t we just parse everything as a unit that looks like a unit?

I don’t know if there are contexts where it’s ambiguous whether a token is a unit or not but I’d think everything that starts with a number is a CSSSize and every token that follows such a number without white-space in between is a unit. Or am I missing something?

I know this implies a less strict validation but the parser was always meant as a parser that checks syntax, not semantics.

pierlon:
> I don’t know if there are contexts where it’s ambiguous whether a token is a unit or not but I’d think everything that starts with a number is a CSSSize and every token that follows such a number without white-space in between is a unit.

Yes, you're correct. In this case the parser would be looking for either a [dimension](https://drafts.csswg.org/css-values-3/#dimensions) (number immediately followed by a unit identifier) or a [percentage](https://drafts.csswg.org/css-values-3/#percentages) ( number immediately followed by a percent sign).

Removing the hard-coded units along with the validation of the size unit does seem like the best option here, and if that's the case, it will be a breaking change and a major release would be warranted.

pierlon:
:warning: Rebasing with latest changes from `master`.

Issue #186, opened 2020-01-24
Bug in parsing CSS vars


EvanHerman:
Currently working on a project that uses PHP-CSS-Parser. We've defined some CSS vars in the following manner:
```css
:root {
   --theme-header--bg: 236,47%,46%;
}
```

And then using them throughout our CSS in the following way:
```css
.header__navigation {
   background-color: hsl(var(--theme-header--bg));
}
```

However, when `:root { --theme-header--bg:236,47%,46%; }` is run through `PHP-CSS-Parser` it is stripped and the our CSS declarations where we are using the `--theme-header--bg` var is being dropped.

Original discussion: https://wordpress.org/support/topic/css-hsl-stripped-not-working/

westonruter:
It's actually the `hsl(var(--theme-header--bg))` which is getting dropped.

So given:

```css
:root {
    --theme-header--bg: 236,47%,46%;
}
.header__navigation {
    color: black;
    background-color: hsl(var(--theme-header--bg));
    border: solid 1px red;
}
```

The PHP-CSS-Parser is only able to parse this as: 

```css
:root {
    --theme-header--bg:236,47%,46%; 
}
.header__navigation {
    color:black;
    border:solid 1px red; 
}
```

Notice how `background-color: hsl(var(--theme-header--bg))` is dropped.

A workaround is to move the `hsa()` function to the variable, like so:

```css
:root {
    --theme-header--bg: hsl(236,47%,46%);
}
.header__navigation {
    color: black;
    background-color: var(--theme-header--bg);
    border: solid 1px red;
}
```

But since the browsers work either way, PHP-CSS-Parser should also work.

EvanHerman:
@sabberworm Any insights you might be able to provide here?

sabberworm:
@EvanHerman Hm, I agree it’s a bug but I don’t currently have any time to look into this myself.

I’ll gladly accept PRs though…

sabberworm:
According to the [specs](https://www.w3.org/TR/css-variables/#using-variables), “[t]he var() function can be used in place of any part of a value in any property on an element”. I think we currently expect it only as single value (as opposed to “any part of a value”).

I don’t know how we could reconcile this with the way the parsed content is structured currently. We obviously cannot resolve the value(s) since we don’t know the run-time cascade but we also can’t parse a reasonable representation of the `hsl` function… In theory this would probably mean we can lex – but never fully parse – any expression if it contains `var(…)`s.

[...]

PR #197, opened 2020-07-10
Fix var function inside color functions

### LINKED ISSUE #196: var function inside color functions cannot be parsed
Thank you for this library. I just started playing around with it and it works really good so far.

But I just noticed that parsing a rule like the following fails:

```css
a {
    background: rgba(var(--some-rgb) , 0.1);
}
```

The reason is that the var is inside the rgba function. If strict mode is enabled, this error is given: `Token “,” (1) not found. Got “var(-”.`

But this is valid CSS, see here for example: https://stackoverflow.com/a/41265350/1044527

UPDATE: fails also for `rgb(var(--some-rgb))`

digilist:
I tried to fix #196. Not sure if it is good solution, but at least it works ;)

### DIFF
--- a/lib/Sabberworm/CSS/Value/Color.php
+++ b/lib/Sabberworm/CSS/Value/Color.php
@@ -52,15 +52,7 @@ class Color extends CSSFunction {
 				}
 
 				if ($bContainsVar && $oParserState->comes(')')) {
-					// With a var argument the function can have fewer arguments and the assignment to the color
-					// channels does not make sense
-					if ($i < $iLength - 1) {
-						// Add remaining characters to last item
-						$remainingChars = substr($sColorMode, $i + 1);
-						$aColor[$sColorMode[$i] . $remainingChars] = $aColor[$sColorMode[$i]];
-						unset($aColor[$sColorMode[$i]]);
-					}
-
+					// With a var argument the function can have fewer arguments
 					break;
 				}
 
@@ -70,6 +62,10 @@ class Color extends CSSFunction {
 				}
 			}
 			$oParserState->consume(')');
+
+			if ($bContainsVar) {
+				return new CSSFunction($sColorMode, array_values($aColor), ',', $oParserState->currentLine());
+			}
 		}
 		return new Color($aColor, $oParserState->currentLine());
 	}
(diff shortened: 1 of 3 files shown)


lib/Sabberworm/CSS/Value/Color.php
```diff
@@ -39,10 +39,31 @@ public static function parse(ParserState $oParserState) {
...
+				}
+
+				if ($bContainsVar && $oParserState->comes(')')) {
+					// With a var argument the function can have fewer arguments and the assignment to the color
+					// channels does not make sense
+					if ($i < $iLength - 1) {
+						// Add remaining characters to last item
+						$remainingChars = substr($sColorMode, $i + 1);
+						$aColor[$sColorMode[$i] . $remainingChars] = $aColor[$sColorMode[$i]];
+						unset($aColor[$sColorMode[$i]]);
```
digilist, line 61: I do not really like this solution but, but it's necessary as the array keys are used in the render function below to define the color function. And because of the nature of `var()` it's not possible to easily tell which arguments are defined through the variable.

If it's not clear what I mean: `rgba(var(--rg), 255)` is valid CSS code and the variable value can contain a comma, so it's valid CSS and in the parser it's hard to tell whether it's for red and green. It's somehow possible to identify it, but this will result in quite some code to identify all combinations and I am not sure if its necessary.

sabberworm: What if we parse colors that contain variables not as a `Color` instance but just an instance of a generic `CSSFunction`? I don’t see the additional value (pun intended) of using a specialized class when we can’t use it for any of the specializations we built it for…

westonruter:
This should also fix #186.

lib/Sabberworm/CSS/Value/Color.php
digilist: Good idea, I updated it as suggested.

westonruter:
@sabberworm Anything else you'd like to see from this? We'd like to include it in the v2.0 release of the AMP plugin for WordPress.

sabberworm:
@westonruter No, I just haven’t had time to do a proper review yet. I only wonder if it would be possible to solve this case generically as opposed to special-casing the color parsing. Next we’ll have someone wanting to use variables in `url()` and we’ll have to add a special case there.

PR #199, opened 2020-10-28
Remove tests and others from distribution

viezel:


sabberworm:
Can you elaborate on the benefits? Smaller file size?
We discussed removing the lock file from the repo already and concluded it should stay in.

viezel:
Yes, it’s a best practice to only ship production needed files. Php is used in many different environments where size matters like serverless. 

More here https://php.watch/articles/composer-gitattributes

sabberworm:
I didn’t know that composer respects git attributes so wohoo for that. But this will also affect ZIP file downloads from GitHub, won’t it. I’m not sure people expect test files to be missing when downloading a ZIP file of the repository.

viezel:
Only a release zip. If people download the code using the standard clone to zip feature. They will get it all.

sabberworm (approved):

PR #204, opened 2020-11-12
Fixes #203

### LINKED ISSUE #203: Ordering of properties after expanding shorthands is wrong
```css
.some-rule{
	padding:5px;
	padding-top: 20px
}
```
leads to
```
array(5) {
  [0]=>
  string(18) "padding-top: 20px;"
  [1]=>
  string(17) "padding-top: 5px;"
  [2]=>
  string(19) "padding-right: 5px;"
  [3]=>
  string(20) "padding-bottom: 5px;"
  [4]=>
  string(18) "padding-left: 5px;"
}
```
Giving padding-top a 5px instead of the expected 20px when getting rules as assoc. Most likely related to that creating shorthands leads to the original input instead of merging the padding rules.

Idrinth:
Order of rules will now be preserved in most cases when expanding shorthands

### DIFF
--- a/lib/Sabberworm/CSS/Parsing/ParserState.php
+++ b/lib/Sabberworm/CSS/Parsing/ParserState.php
@@ -44,6 +44,10 @@ class ParserState {
 		return $this->iLineNo;
 	}
 
+	public function currentColumn() {
+		return $this->iCurrentPosition;
+	}
+
 	public function getSettings() {
 		return $this->oParserSettings;
 	}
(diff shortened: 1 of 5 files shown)


sabberworm:
Thanks! Could you add your example from #203 as a unit test case?

Idrinth:
will have to clean that up after work tonight regarding cleanups and fixing tests

Idrinth:
@sabberworm let me know if there's anything left to change, it looks ok to me now.

sabberworm (approved):

PR #206, opened 2021-04-18
Add some PHPDoc type annotations

oliverklee:
This will help static type analysis tools.

### DIFF
--- a/lib/Sabberworm/CSS/Property/Charset.php
+++ b/lib/Sabberworm/CSS/Property/Charset.php
@@ -19,6 +19,10 @@ class Charset implements AtRule {
 	 * @var int
 	 */
 	protected $iLineNo;
+
+	/**
+	 * @var array
+	 */
 	protected $aComment;
 
 	/**
--- a/lib/Sabberworm/CSS/Property/Charset.php
+++ b/lib/Sabberworm/CSS/Property/Charset.php
@@ -10,11 +10,21 @@ namespace Sabberworm\CSS\Property;
  * • Must not appear more than once.
  */
 class Charset implements AtRule {
-
+	/**
+	 * @var string
+	 */
 	private $sCharset;
+
+	/**
+	 * @var int
+	 */
 	protected $iLineNo;
 	protected $aComment;
 
+	/**
+	 * @param string $sCharset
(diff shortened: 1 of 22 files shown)


sabberworm:
Really cool, thanks. Is there some check we could install into the CI build that ensures everything remains annotated?

oliverklee:
> Is there some check we could install into the CI build that ensures everything remains annotated?

Yes, I will add PHPStan checks later to make sure the types are correct.

We can also add PHP_CodeSniffer checks for checking for the presence of type annotations, but I'd like to avoid that as we probably won't need many of them once we've updated the PHP requirement and we can use more of the native type declarations.

westonruter:
> We can also add PHP_CodeSniffer checks for checking for the presence of type annotations, but I'd like to avoid that as we probably won't need many of them once we've updated the PHP requirement and we can use more of the native type declarations.

Updating the PHP requirement detrimental for WordPress which still supports PHP 5.6. If PHPDoc can be used until WordPress [increases the minimum](https://make.wordpress.org/core/2020/08/24/proposal-dropping-support-for-old-php-versions-via-a-fixed-schedule/), this would be ideal. Otherwise, the [dozens of WordPress plugins](https://wpdirectory.net/search/01F3NM1ZE9PHFQJ957XEXBJJ7C) that use PHP-CSS-Parser would be forced to fork the package.

oliverklee:
@westonruter @sabberworm I'd really appreciate this PR getting reviewed and merged as it blocks other things I'd like to work on that also touch the code. Thanks! :heart:

lib/Sabberworm/CSS/Property/Charset.php
```diff
@@ -10,11 +10,21 @@
...
+	/**
+	 * @var string
+	 */
 	private $sCharset;
+
+	/**
+	 * @var int
+	 */
 	protected $iLineNo;
 	protected $aComment;
```
westonruter, line 22:
```suggestion

	/**
	 * @var array
	 */
	protected $aComment;
```

westonruter (approved):
Looks good! Just one additional phpdoc tag that I saw could be added as well.

sabberworm (approved):

PR #209, opened 2021-04-18
Stop versioning `composer.lock`

oliverklee:
`composer.lock` should only be versioned for projects, not for libraries.


oliverklee:
I will update this PR once #208 is merged.

PR #210, opened 2021-04-18
Add a LICENSE file

oliverklee:


LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
```
sabberworm, line 1: Do you think having the license in the README isn’t legally enforcible? In any case, I would assume having it only in one place is sufficient so if we have a separate file, please remove it from the README.

oliverklee: I don't know about legals here. It's mostly about following established conventions, and about help GitHub (which can process LICENSE files, but not license sections in the README).

Will change.

oliverklee: Updated and repushed.

sabberworm (approved):

PR #212, opened 2021-04-21
Add an `.editorconfig` file

oliverklee:
This helps contributors get the basic formatting options in their IDEs
automatically set right. (The setting for PHP files complies to the
PSR-12 standard.)

This change does not autoformat any files with these settings, though.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# top-most EditorConfig file
+root = true
+
+# Unix-style newlines with a newline ending every file, and with same defaults
+[*]
+charset = utf-8
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+indent_style = space
+indent_size = 4
+max_line_length = 120
+
+[*.md]
+max_line_length = 80
+# GitHub-flavored markdown uses two spaces and the end of a line to indicate a linebreak.
+trim_trailing_whitespace = false
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,4 @@
+/.editorconfig export-ignore
 /.gitattributes export-ignore
 /.github/ export-ignore
 /.gitignore export-ignore


.editorconfig
```diff
@@ -0,0 +1,34 @@
...
+
+# PHP files
+[*.php]
+indent_style = space
+indent_size = 4
+
+# MD files
+[*.md]
+indent_style = space
+indent_size = 4
```
westonruter, line 19: GitHub-flavored markdown uses two spaces and the end of a line to indicate a linebreak, so `trim_trailing_whitespace` should be disabled for markdown files.

```suggestion
indent_size = 4
trim_trailing_whitespace = false
```

oliverklee:
Hold on, I think Markdown should have lines no longer than 80 characters. I'll do some more research on best practices there.

.editorconfig
```diff
@@ -0,0 +1,20 @@
...
+max_line_length = 120
+
+[*.md]
+max_line_length = 80
+# GitHub-flavored markdown uses two spaces and the end of a line to indicate a linebreak.
+trim_trailing_whitespace = false
+
+[*.css]
+indent_style = tab
```
oliverklee, line 20: I'm not quite sure about this. Currently, the CSS test files are indented with tabs, and so this setting would keep this tabs. Or we could (in a separate PR) convert the CSS files to also use spaces for consistency. I'd prefer the second option and would be okay with both. :-)

@westonruter @sabberworm What do you think?

oliverklee: And should we allow (semi-)unlimited line lengths for the CSS files (as some files have compressed CSS)?

oliverklee: I've looked through some more CSS files: Their indentation is are a colorful mixture of tabs, 2 spaces and 4 spaces. :-)

sabberworm: Well, adding an editorconfig doesn’t mean the files need to be reformatted, it’s just a hint to the IDE on how such a file _should_ be formatted.
I think it’s OK for fixtures to differ in indentation since the parser needs to be able to handle all styles found in the wild. Some fixtures are just real-life cases of files that the parser couldn’t handle correctly, which is why they’re committed as-is.
Personally I vastly prefer tab indentation over spaces but I prefer consistency even more so I don’t think we need to special-case `*.css` here.

oliverklee: Okay, I've remove the CSS special case now and repushed.

PR #214, opened 2021-04-21
Add PHP linting to the CI build

oliverklee:
This ensures that the PHP files have valid syntax, and blocks the unit
tests if there are any errors found.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,10 +9,32 @@ on:
 name: CI
 
 jobs:
+  php-lint:
+    name: PHP Lint
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -9,11 +9,33 @@ on:
 name: CI
 
 jobs:
+  php-lint:
+    name: PHP Lint
+    runs-on: ubuntu-20.04
+    strategy:
+      matrix:
+        php-version: [ '5.3', '5.4', '5.5', '5.6', '7.0', '7.1', '7.2', '7.3', '7.4' ]
```
sabberworm, line 17: Is there some way to declare the php version list only once in the file?

oliverklee: No, none that I'm currently aware of. (And unfortunately, GitHub Actions do not support YAML anchors yet.)

sabberworm (approved):

PR #215, opened 2021-04-25
Require PHP >= 5.6

### LINKED ISSUE #207: Drop support for older PHP versions
I propose this library drops support for older PHP versions so it can use newer PHP features.

According to https://www.php.net/supported-versions.php, PHP 7.2 recently reached its end of life. So I propose to drop support for all PHP versions < 7.2.

(We then can use Rector to upgrade the code accordingly.)

What do you think?

oliverklee:
Now the PHP version requirement is on par with WordPress:
https://wordpress.org/support/article/requirements/

Fixes #207

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ## x.y.z
 
-* Require PHP >= 5.6
+* Require PHP ≥ 5.6
 
 ## 8.0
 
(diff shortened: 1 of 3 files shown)


westonruter (approved):

westonruter (approved):

CHANGELOG.md
```diff
@@ -1,5 +1,11 @@
 # Revision History
 
+## x.y
+
+## x.y.z
+
+* Require PHP >= 5.6
```
sabberworm, line 7:
```suggestion
* Require PHP ≥ 5.6
```

PR #216, opened 2021-04-25
Update most URLs to HTTPS

oliverklee:


phpunit.xml
```diff
@@ -1,6 +1,6 @@
 <phpunit
-    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-    xsi:noNamespaceSchemaLocation="http://schema.phpunit.de/4.5/phpunit.xsd"
+    xmlns:xsi="https://www.w3.org/2001/XMLSchema-instance"
```
westonruter, line 2: I don't believe this namespace change is correct. I don't see an HTTPS version for the xsi namespace in the spec: https://www.w3.org/TR/xmlschema-1/

Regardless, a namespace is not intended to be fetched.

See also https://github.com/sebastianbergmann/phpunit/blob/b25cb0dc1f77f4187f13c295784fec2b8b6060c6/phpunit.xml#L2

```suggestion
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
```

westonruter (changes requested):

phpunit.xml
oliverklee: Fixed, rebased and repushed.

westonruter (approved):

oliverklee:
Rebased and repushed.

phpunit.xml
sabberworm: I think this also applies to most namespaces used in CSS `@namespace` examples.

oliverklee: Changed and repushed.

PR #218, opened 2021-05-02
Add Composer scripts for code sniffing and autofixing

oliverklee:
There now are the following new Composer scripts:

- `ci`: run all CI-related Composer scripts
- `ci:static` run the static code analysis
- `ci:php:sniff` run PHP_CodeSniffer
- `fix:php` fix the autofixable PHP code warnings
- `fix:php:sniff` fix the autofixable PHP_CodeSniffer warnings

For starters, we have a sniff that checks for the PSR-12 standard.
(The CI builds do not run this check as the files are not
PSR-12 compliant yet at this point.)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,6 +1,8 @@
 /.gitattributes export-ignore
 /.github/ export-ignore
 /.gitignore export-ignore
+/.phive/
 /Doxyfile export-ignore
+/phpcs.xml export-ignore
 /phpunit.xml export-ignore
 /tests export-ignore
--- a/phpcs.xml
+++ b/phpcs.xml
@@ -1,0 +1,15 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<ruleset name="phpList Coding Standard">
+    <description>
+        This standard requires PHP_CodeSniffer >= 3.6.0.
+    </description>
+
+    <arg name="colors"/>
+    <arg name="extensions" value="php"/>
+
+    <!-- The complete PSR-12 ruleset -->
+    <rule ref="PSR12">
+        <!-- Exclude a rule that requires PHP >= 7.1. -->
+        <exclude name="PSR12.Properties.ConstantVisibility"/>
+    </rule>
+</ruleset>
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,5 @@
+/.phive/*
+/.php_cs.cache
 /composer.lock
 /vendor/
+!/.phive/phars.xml
--- a/.phive/phars.xml
+++ b/.phive/phars.xml
@@ -1,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phive xmlns="https://phar.io/phive">
+  <phar name="phpcbf" version="^3.6.0" location="./.phive/phpcbf.phar" copy="false" installed="3.6.0"/>
+  <phar name="phpcs" version="^3.6.0" location="./.phive/phpcs.phar" copy="false" installed="3.6.0"/>
+</phive>
--- a/composer.json
+++ b/composer.json
@@ -17,5 +17,18 @@
     },
     "autoload": {
         "psr-4": { "Sabberworm\\CSS\\": "lib/Sabberworm/CSS/" }
+    },
+    "scripts": {
+        "ci": [
+            "@ci:static"
+        ],
+        "ci:php:sniff": "@php ./.phive/phpcs.phar lib tests",
+        "ci:static": [
+            "@ci:php:sniff"
+        ],
+        "fix:php": [
+            "@fix:php:sniff"
+        ],
+        "fix:php:sniff": "@php ./.phive/phpcbf.phar lib tests"
     }
 }


phive.xml
```diff
@@ -0,0 +1,5 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<phive xmlns="https://phar.io/phive">
+  <phar name="phpcbf" version="^3.6.0" location="./tools/phpcbf.phar" copy="false" installed="3.6.0"/>
+  <phar name="phpcs" version="^3.6.0" location="./tools/phpcs.phar" copy="false" installed="3.6.0"/>
+</phive>
```
westonruter, line 5: I'm not familiar with PHIVE. What's the benefit over installing PHPCS via Composer? Will Dependabot be able to update PHIVE dependencies?

phpcs.xml
```diff
@@ -0,0 +1,15 @@
...
+<ruleset name="phpList Coding Standard">
+    <description>
+        This standard requires PHP_CodeSniffer >= 3.6.0.
+    </description>
+
+    <arg name="colors"/>
+    <arg name="extensions" value="php"/>
+
+    <!-- The complete PSR-12 ruleset -->
+    <rule ref="PSR12">
```
westonruter, line 11: It would also be useful to add the [`PHPCompatibility`](https://github.com/PHPCompatibility/PHPCompatibility) sniffs to make sure that PHP 5.6 features are used.

phive.xml
oliverklee: [PHIVE](https://phar.io/) is a package manager for PHP development tools that installs the tools as stand-alone, GPG-signed PHARs (i.e., a PHAR contains the corresponding tool with all its dependencies). This helps avoiding dependency hell - both inter-tools as well as between a tool and the project in which it is used. (Dependency hell is e.g., when package A requires package X in version < 3, while package B requires package X in version >= 3.) (This was a real problem in some of my projects.)

In addition, this allows tools like [PHPStan](https://packagist.org/packages/phpstan/phpstan) to be used in their current version (which requires PHP >= 7.1) without breaking Composer-installability of the project in lower PHP versions. (This also is a real problem in some of my projects.)

PHIVE is intended to be used for development tools only, and the PHIVE-installed tools are not expected to be shipped with the library package (hence the entries in the `.gitattributes), and they are also not expected to be deployed to production systems.

Dependabot currently does not update the PHIVE-installed tools yet.

phpcs.xml
oliverklee: This probably will be a bit more work to get those to run with a PHIVE-based installation. Please let's do this in a separate PR. (I can do that.)

phive.xml
westonruter: Got it. Makes sense to me, although we've managed to work around this issue in [amp-wp](https://github.com/ampproject/amp-wp) which uses PHPCS and PHPStan and supports PHP 5.6+.

phpcs.xml
westonruter, line 11: To enforce phpdoc, what about also including `Squiz.Commenting`? See also how it is [configured for WordPress-Docs](https://github.com/WordPress/WordPress-Coding-Standards/blob/develop/WordPress-Docs/ruleset.xml).

oliverklee: I, too, have lots of ideas which sniffs to add. :-)

Please let's start with PSR-12, get the included sniffs green, and then add other sniffs one by one (with the necessary code changes). Deal?

westonruter: Deal :)

westonruter (approved):

.gitattributes
```diff
@@ -2,5 +2,8 @@
 /.github/ export-ignore
 /.gitignore export-ignore
 /Doxyfile export-ignore
+/phive.xml
+/phpcs.xml export-ignore
```
sabberworm, line 6: Do we have to have these files in the root dir? I would like to keep the root dir as tidy as possible.

oliverklee: Concerning PHIVE: There is a second default location for the configuration file: `.phive/phars.xml`, and we could also move the tools from `tools/` to `.phive/` as well (like we do in the [Emogrifer project](https://github.com/MyIntervals/emogrifier)).

We can also move the PHP_Codesniffer configuration to file and that reference the configuration ("code standard") file in the Composer commands, e.g., to `config/phpcs.xml`. What do you think?

sabberworm: I like it. Maybe phpunit.xml could also be moved to the config folder? And Doxyfile. Which reminds me: we should have a task that updates gh-pages after releases to publish https://sabberworm.github.io/PHP-CSS-Parser/.

westonruter: > We can also move the PHP_Codesniffer configuration to file and that reference the configuration ("code standard") file in the Composer commands, e.g., to `config/phpcs.xml`. What do you think?

A downside to that is that tools/editors will fail to discover the configuration files automatically.

oliverklee: I'll move the PHIVE configuration file (and the PHAR symlinks) in this PR. We can take care of the other configuration files later then (in separate issues/PRs as needed).

oliverklee: I have moved the PHARs and the PHIVE configuration file now and repushed.

PR #220, opened 2021-05-03
Stop trying to upload the coverage if there is no auth token

oliverklee:
This fixes build failures on pull requests from forks (where
usually no authentication token for the code coverage is set).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -72,9 +72,9 @@ jobs:
       - name: Run Tests
         run: ./vendor/bin/phpunit --coverage-clover build/coverage/xml
 
-      - name: Upload coverage results to Coveralls
-        if: "${{ matrix.coverage != 'none' }}"
+      - name: Upload coverage results to Codacy
         env:
           CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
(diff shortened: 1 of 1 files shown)


oliverklee:
Note to myself: I need to do this a different way.
https://github.com/actions/runner/issues/520

.github/workflows/ci.yml
```diff
@@ -72,9 +72,9 @@ jobs:
       - name: Run Tests
         run: ./vendor/bin/phpunit --coverage-clover build/coverage/xml
 
-      - name: Upload coverage results to Coveralls
-        if: "${{ matrix.coverage != 'none' }}"
+      - name: Upload coverage results to Codacy
         env:
           CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
+        if: "${{ matrix.coverage != 'none' && env.CODACY_PROJECT_TOKEN != '' }}"
```
pierlon, line 78: > This fixes build failures on pull requests from forks...

You could instead check whether the PR is not from a fork:

```suggestion
        if: github.event.pull_request.head.repo.fork == false
```

oliverklee: I'd prefer it this way as this allows forks to still upload coverage to their own coverage server by setting the auth key.

pierlon: Ah I see. That would include the person also using Codacy.

oliverklee: Indeed.

PR #221, opened 2021-05-03
Autofix all autofixable PSR-12 code style issues

oliverklee:
These changes were all automatically done by PHP_CodeSniffer,
and there are no manual changes.


sabberworm:
Personally I find PSR-12 styles ugly. But, then again, that fits in nicely with the rest of PHP, one might argue…

PR #222, opened 2021-05-07
Add missing method visibilties

oliverklee:
This gets the code base one more step towards PSR-12 compliance.


westonruter (approved):
I suppose we don't need to add `PSR2.Methods.MethodDeclarationSniff` to PHPCS if the `PSR2` standard is being used already.

oliverklee:
> I suppose we don't need to add `PSR2.Methods.MethodDeclarationSniff` to PHPCS if the `PSR2` standard is being used already.

That's correct.

PR #223, opened 2021-05-07
Use PHP 5.6 language features

oliverklee:


westonruter:
WIth the switch from `array()` to `[]`, should the PHPCS config be updated to add `Generic.Arrays.DisallowLongArraySyntax`?

westonruter (approved):

oliverklee:
> WIth the switch from `array()` to `[]`, should the PHPCS config be updated to add `Generic.Arrays.DisallowLongArraySyntax`?

Good catch, yes! I will update the PR.

sabberworm (approved):

PR #224, opened 2021-05-16
Remove from CSSList via key

bytestream:
If you're looping over `getContents()` and want to remove items from the `CSSList` it would be more performant to remove items by the key given it's already known. The current method uses `array_search` which is wasteful in the specified scenario.

### DIFF
--- a/lib/Sabberworm/CSS/CSSList/CSSList.php
+++ b/lib/Sabberworm/CSS/CSSList/CSSList.php
@@ -246,11 +246,18 @@ abstract class CSSList implements Renderable, Commentable
 
     /**
      * Removes an item from the CSS list.
-     * @param RuleSet|Import|Charset|CSSList $oItemToRemove May be a RuleSet (most likely a DeclarationBlock), a Import, a Charset or another CSSList (most likely a MediaQuery)
+     *
+     * @param RuleSet|Import|Charset|CSSList|int $oItemToRemove May be a RuleSet (most likely a DeclarationBlock), a Import, a Charset or another CSSList (most likely a MediaQuery)
      * @return bool Whether the item was removed.
      */
     public function remove($oItemToRemove)
     {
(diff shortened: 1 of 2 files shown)


westonruter:
Shouldn't you rather than use the `splice()` method? The `remove()` method wasn't intended to accept an array index.

Aside: it seems `remove()` needs to re-index the numeric keys to make them all sequential.

bytestream:
> Shouldn't you rather than use the splice() method?

`unset` is marginally faster than `array_splice` on large arrays. It's also easier to work with, `array_splice` keeps reindexing so it's difficult to keep track of the offset to remove in a loop

> it seems remove() needs to re-index the numeric keys to make them all sequential.

How come? It seems to work OK as it is.

lib/Sabberworm/CSS/CSSList/CSSList.php
```diff
@@ -246,11 +246,18 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
 
     /**
      * Removes an item from the CSS list.
-     * @param RuleSet|Import|Charset|CSSList $oItemToRemove May be a RuleSet (most likely a DeclarationBlock), a Import, a Charset or another CSSList (most likely a MediaQuery)
+     *
+     * @param RuleSet|Import|Charset|CSSList|int $oItemToRemove May be a RuleSet (most likely a DeclarationBlock), a Import, a Charset or another CSSList (most likely a MediaQuery)
```
oliverklee, line 250: I strongly recommend not mixing types likes this in a parameter. Instead, I propose a new, separate method `removeByKey(int $key)`.

westonruter: Agreed.

bytestream: Happy to make the change but are the points raised in https://github.com/sabberworm/PHP-CSS-Parser/pull/224#issuecomment-841852668 still a concern with this PR?

PR #227, opened 2021-05-27
Add static type analysis with PHPStan

oliverklee:
Add Composer scripts for checking the code with PHPStan (while using
a baseline) and for updating the baseline.

When we fix PHPStan-reported issues (or update the PHPStan configuration),
we can update the baseline bit by bit until all issues are fixed.

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -1,0 +1,27 @@
+parameters:
+	ignoreErrors:
+		-
+			message: "#^Class Sabberworm\\\\CSS\\\\Parsing\\\\ParserState constructor invoked with 1 parameter, 2\\-3 required\\.$#"
+			count: 1
+			path: lib/Sabberworm/CSS/CSSList/CSSList.php
+
+		-
+			message: "#^Access to an undefined property Sabberworm\\\\CSS\\\\Parsing\\\\ParserState\\:\\:\\$oParserHelper\\.$#"
+			count: 1
+			path: lib/Sabberworm/CSS/Parsing/ParserState.php
+
+		-
+			message: "#^Access to an undefined property Sabberworm\\\\CSS\\\\Property\\\\Charset\\:\\:\\$aComments\\.$#"
+			count: 5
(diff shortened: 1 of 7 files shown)


oliverklee:
Marking as draft again. I think we should start with level 0 first (so that the baseline file is more manageable), and work our way up through the levels.

I'd also like to add this to the CI build.

sabberworm:
> I'd also like to add this to the CI build.

Feel free to do this. Let the PR CI fail until all code is up to standards.

oliverklee:
> Let the PR CI fail until all code is up to standards.

Even better: With the baseline file, the build will pass with the code as-is, while we still prevent the introduction of new typing issues. Neat, isn't it?

phpstan-baseline.neon
```diff
@@ -0,0 +1,27 @@
...
+		-
+			message: "#^Class PHPUnit\\\\Framework\\\\TestCase referenced with incorrect case\\: PHPunit\\\\Framework\\\\TestCase\\.$#"
+			count: 1
+			path: tests/Sabberworm/CSS/OutputFormatTest.php
+
+		-
+			message: "#^Class PHPUnit\\\\Framework\\\\TestCase referenced with incorrect case\\: PHPunit\\\\Framework\\\\TestCase\\.$#"
+			count: 1
+			path: tests/Sabberworm/CSS/ParserTest.php
+
```
sabberworm, line 27: Cool, I thought there’d be more

phpstan.neon
```diff
@@ -0,0 +1,16 @@
+includes:
```
sabberworm, line 1: Should we name this `phpstan.neon.dist` as per https://phpstan.org/config-reference#config-file?

sabberworm (approved):

phpstan-baseline.neon
oliverklee: That's also because we're starting at level 0. (It goes up to level 8 …)

phpstan.neon
oliverklee: Thanks! Changed, rebased and repushed.

PR #236, opened 2021-05-30
Move the production code from `lib/` to `src/`

### LINKED ISSUE #230: Flatten the production code directory structure
To be more consistent with most other PHP projects, I'd like to change the directory structure like this (without changing the namespaces of the PHP classes):

`lib/Sabberworm/CSS/*.php` -> `src/*.php`

@sabberworm @westonruter Would this be okay for you?

oliverklee:
This gets the directory structure more in line with what most
other projects (including Symfony) are using. (The PHP namespaces
are unchanged, though.)

Also flatten the directory structure.

Fixes #230


oliverklee:
Rebased and repushed.

PR #239, opened 2021-06-01
Remove access to inexistent property in `ParserState`

oliverklee:


oliverklee:
Rebased and repushed.

oliverklee:

PR #241, opened 2021-06-01
Raise PHPStan checks to level 1

oliverklee:
Now that there are fixes for all level 0 warnings, we can tackle
the next level of warnings.


oliverklee:
Rebased and repushed.

oliverklee:
Thanks for the merge, @sabberworm! I'll probably need to rebase a bunch of other PRs now. :-)

PR #242, opened 2021-06-02
Move the `OutputFormatter` class to a separate file

oliverklee:
This resolves a PSR-12 violation.


oliverklee:

PR #244, opened 2021-06-02
Autoformat all PHP files with PhpStorm

oliverklee:
This allows contributors to autoformat their changed files without
risking to also affect the unchanged code.

This change also fixes a bunch of PSR-12 violations, and it hopefully
also makes the code a bit nicer to read.


oliverklee:

PR #245, opened 2021-06-02
Adapt the PHPStan baseline to the new directory structure

oliverklee:
This should make the build green again.


sabberworm:
You’re really quick but I was also working on that…

PR #247, opened 2021-06-02
Drop the quickdump tool

oliverklee:
This tool does not seem to be referenced or documented anywhere.


sabberworm:
It’s the tool I used to generate the examples in the readme… I think that could be automated though.

oliverklee:
@sabberworm So should we keep the tool (and maybe move it to `bin/`), or should it go?

sabberworm:
I think it should be refactored into a small script that can be called from a github action to automatically update the README.

oliverklee:
I have now created #247 for moving the file out of the `tests/` directory. (I propose we do the refactoring in a separate step.)

PR #259, opened 2021-06-08
Allow calc() function inside rgba() and hsla()

thomashigginbotham:
This fix allows the alpha value of rgba() and hsla() functions to accept
a calc() function.

### DIFF
--- a/tests/ParserTest.php
+++ b/tests/ParserTest.php
@@ -120,7 +120,10 @@ class ParserTest extends \PHPUnit\Framework\TestCase
             . 'background-color: rgb(255,var(--rg));background-color: hsl(var(--some-hsl));}'
             . "\n"
             . '#variables-alpha {background-color: rgba(var(--some-rgb),.1);'
-            . 'background-color: rgba(var(--some-rg),255,.1);background-color: hsla(var(--some-hsl),.1);}',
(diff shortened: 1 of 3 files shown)


tests/ParserTest.php
```diff
@@ -120,7 +120,10 @@ public function testColorParsing()
             . 'background-color: rgb(255,var(--rg));background-color: hsl(var(--some-hsl));}'
             . "\n"
             . '#variables-alpha {background-color: rgba(var(--some-rgb),.1);'
-            . 'background-color: rgba(var(--some-rg),255,.1);background-color: hsla(var(--some-hsl),.1);}',
+            . 'background-color: rgba(var(--some-rg),255,.1);background-color: hsla(var(--some-hsl),.1);}'
```
oliverklee, line 123: Soon, I'd like to split up the tests so that each test method is only about one piece of behavior, not multiple things mashed together. You can help my work by moving the test for the new behavior into a new test. Thanks!

thomashigginbotham: Not sure if I'll have time for awhile. I needed a quick fix for the `calc()` issue, and I saw the necessary changes were simple enough. I didn't dig too far into the unit testing, but I agree that it would be better to break them up into separate tests.

oliverklee: @thomashigginbotham If you don't have time now, then let's keep this PR as it is, and then I will split up the test later. Kudos to you for covering your change with a test, by the way! :heart:

PR #264, opened 2021-06-14
Run the code sniffer in the CI pipeline

oliverklee:
Now that the code is PSR-12-compliant, we should run the code sniffer
in the CI pipeline to keep pull request from introducing new
code style issues.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -79,12 +79,23 @@ jobs:
         run: |
           ./vendor/bin/codacycoverage clover build/coverage/xml
 
-  phpstan:
-    name: PHPStan
+  static-analysis:
+    name: Static Analysis
 
     runs-on: ubuntu-20.04
 
     needs: [ php-lint ]
+
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - command: sniff
+            tool: phpcs
+            php-version: 7.4
+          - command: stan
+            tool: phpstan
+            php-version: 7.4
 
     steps:
       - name: Checkout
@@ -93,7 +104,7 @@ jobs:
       - name: Install PHP
         uses: shivammathur/setup-php@v2
         with:
-          php-version: 7.4
+          php-version: ${{ matrix.php-version }}
           tools: "composer:v2, phive"
           coverage: none
 
@@ -112,7 +123,8 @@ jobs:
 
       - name: Install development tools
         run: |
-          echo y | phive --no-progress update phpstan;
+          # These workarounds are needed until PHIVE can do installs in a non-interactive way
+          echo y | phive --no-progress update ${{ matrix.tool }};
 
-      - name: Run PHPStan
-        run: "composer ci:php:stan"
+      - name: Run Command
+        run: composer ci:php:${{ matrix.command }}


.github/workflows/ci.yml
```diff
@@ -79,6 +79,44 @@ jobs:
...
+        with:
+          php-version: 7.4
+          tools: "composer:v2, phive"
+          coverage: none
+
+      - name: Cache dependencies installed with composer
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/composer
+          key: php${{ matrix.php-version }}-composer-${{ hashFiles('**/composer.json') }}
```
sabberworm, line 104: I don’t know too much about github actions still but it seems counter-intuitive to me that `matrix.php-version` would be defined in this action that doesn’t even use matrix.

.github/workflows/ci.yml
```diff
@@ -79,6 +79,44 @@ jobs:
         run: |
           ./vendor/bin/codacycoverage clover build/coverage/xml
 
+  phpcs:
+    name: PHP_CodeSniffer
```
sabberworm, line 83: I would prefer we only have one action for static analysis and run all checks there.
I realize the output for failures is arguably nicer but there’s just an uncomfortable amount of repetition between `phpcs` and `phpstan`.
It would probably also mean that if both checks fail, you would only see one of the errors. I think we can live with that.
Linting can be kept separate since we want to run it against multiple versions of PHP.

sabberworm (changes requested):

.github/workflows/ci.yml
oliverklee: I'll refactor this to use a matrix.

.github/workflows/ci.yml
oliverklee: Oh, that's a copy'n'paste error. Good catch!

.github/workflows/ci.yml
sabberworm: Ah, yes, matrix would also work. Thanks.

.github/workflows/ci.yml
oliverklee: I have now moved the PHP version to a (single-element) matrix so we have exactly one place for the PHP version (which is used both in the PHP install step as well as the caching step) and so that it's consistent with the structure of the other jobs.

.github/workflows/ci.yml
```diff
@@ -112,7 +121,9 @@ jobs:
 
       - name: Install development tools
         run: |
+          # These workarounds are needed until PHIVE can do installs in a non-interactive way
           echo y | phive --no-progress update phpstan;
+          echo y | phive --no-progress update phpcs;
```
sabberworm, line 126:
```suggestion
          echo y | phive --no-progress update ${{ matrix.command }};
```
?

sabberworm: Ah, the script name doesn’t equal the command name.

sabberworm: Maybe unify the two?

oliverklee: You mean, like merge into one command? Or parametrize the command?

sabberworm: I meant: name the script the same as the command it’s calling, i.e. phpstan and phpcs instead of sniff and stan.

oliverklee: I have now moved the tool name into the matrix as well so we can have keep the nice command names in the Composer scripts.

sabberworm: Can matrix variables only be primitives? If not, the following could also work:

```yaml
matrix:
		command:
			- name: sniff
			  tool: phpcs
			- name: stan
			  tool: phpstan
```

and then use `${{ matrix.command.name }}` and `${{ matrix.command.tool }}`?

oliverklee: Thanks! This would probably work, but I'm not really sure. I'll give it a try.

oliverklee: No, this does not seem to work. I'll undo the corresponding change and repush.

sabberworm: OK, thanks for trying, though.

PR #272, opened 2021-06-19
Add `@covers` annotations to the tests

oliverklee:
These make sure that only the lines of the classes the testcases are
about are marked as covered, hence reducing false coverage.

### DIFF
--- a/tests/OutputFormatTest.php
+++ b/tests/OutputFormatTest.php
@@ -6,6 +6,9 @@ use PHPUnit\Framework\TestCase;
 use Sabberworm\CSS\OutputFormat;
 use Sabberworm\CSS\Parser;
 
+/**
+ * @covers \Sabberworm\CSS\OutputFormat
+ */
 class OutputFormatTest extends TestCase
 {
     /**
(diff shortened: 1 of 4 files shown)


oliverklee:
Note: This change _will_ reduce the measured code coverage score. On the other hand, the new number will reflect reality much better.

oliverklee:
Review note: I'll appreciate a critical eye on whether the annotations are complete, or whether we need to add annotations for more covered classes to the test cases.

sabberworm:
> Review note: I'll appreciate a critical eye on whether the annotations are complete, or whether we need to add annotations for more covered classes to the test cases.

I think the ones you marked as covering `\Sabberworm\CSS\Parser` are also supposed to cover all of the static `parse` methods like `Color::parse`, `CSSString::parse`, etc.

oliverklee:
@sabberworm Thanks! I will update the annotations accordingly (and rebase).

oliverklee:
Rebased, updated according to the review comments, and repushed.

sabberworm:
> Rebased, updated according to the review comments, and repushed.

Thanks

PR #287, opened 2021-06-27
Add PHP CS Fixer to the toolchain

oliverklee:
This tool will help detect and fix more PSR-12 issues, and it will also
help modernize the tests so that it's easier to cover issues found by
PHPStan with tests.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -90,8 +90,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          - command: sniff
+          - command: sniffer
             tool: phpcs
+            php-version: 7.4
+          - command: fixer
+            tool: php-cs-fixer
             php-version: 7.4
           - command: stan
             tool: phpstan
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /.phive/*
+/.php-cs-fixer.cache
 /.php_cs.cache
 /composer.lock
 /phpstan.neon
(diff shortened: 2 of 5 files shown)


.gitattributes
```diff
@@ -3,6 +3,7 @@
 /.github/ export-ignore
 /.gitignore export-ignore
 /.phive/ export-ignore
+/.php-cs-fixer.dist.php export-ignore
```
sabberworm, line 6: Now would maybe be an appropriate time to group all the config files for checking tools into a common folder so that the root folder doesn’t become a mess.

oliverklee: Done: #288

I'll update this PR once the other PR is merged.

Should we move `phpunit.xml` as well?

oliverklee: And should we also move the `Doxyfile`?

sabberworm: Yes, Doxyfile would be great. I’m on the fence about phpunit.xml. I kinda like the fact that you can just run phpunit without args in the root dir. But I think I like clean dirs even more… WDYT?

oliverklee: Okay, then I'll move `Doxyfile` as well. Concerning `phpunit.xml`: I'd also prefer to keep it in the project root as  this makes tool integration a lot easier (e.g., being able to easily run the tests in PhpStorm).

oliverklee: Concerning the config file move: Done, rebased and repushed.

PR #294, opened 2021-07-01
Add type annotations for `Commentable`

oliverklee:
Also copy these type annotations to the implementing classes.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -32,7 +32,7 @@ use Sabberworm\CSS\Value\Value;
 abstract class CSSList implements Renderable, Commentable
 {
     /**
-     * @var array<array-key, Comment> $aComments
+     * @var array<array-key, Comment>
      */
     protected $aComments;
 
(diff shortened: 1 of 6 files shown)


sabberworm:
Is it really necessary to repeat type annotations in overrides? Isn’t there some tag (like `@inheritDoc` in Java) that can be used instead?

oliverklee:
Technically, we need to have _something_ - be it either a copy of the iterface's method, or be it `@inheritdoc` (which should be recognized by tools like PHPStan, Psalm and PhpStorm's static analysis).

However, for the human reader who scrolls through the source code, I very strongly prefer copying the type annotations. This is because the human reader would need to click on the `@inheritdoc` annotation (or hover, depending on which plugins are installed) to see the types, which makes quick reading quite impossible.

sabberworm:
> However, for the human reader who scrolls through the source code, I very strongly prefer copying the type annotations. This is because the human reader would need to click on the `@inheritdoc` annotation (or hover, depending on which plugins are installed) to see the types, which makes quick reading quite impossible.

On the other hand, maintenance is going to be a pain when we want to change the signature (unless we have a tool that enforces consistent type annotations).

oliverklee:
> On the other hand, maintenance is going to be a pain when we want to change the signature (unless we have a tool that enforces consistent type annotations).

This indeed is true. PHPStan (on higher levels) will at least warn us if the type annotations get inconsistent.

PR #296, opened 2021-07-02
Streamline the tests code style

oliverklee:
- calls the assert methods statically
- use `@test` annotations instead of the `test` method name prefix

Also add PHP CS Fixer rules to automate these changes.

### DIFF
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -11,5 +11,17 @@ return (new \PhpCsFixer\Config())
             '@PSR12' => true,
             // Disable constant visibility from the PSR12 rule set as this would break compatibility with PHP < 7.1.
             'visibility_required' => ['elements' => ['property', 'method']],
+
+            'php_unit_construct' => true,
+            'php_unit_dedicate_assert' => ['target' => '5.0'],
+            'php_unit_expectation' => ['target' => '5.6'],
+            'php_unit_fqcn_annotation' => true,
+            'php_unit_method_casing' => true,
+            'php_unit_mock' => ['target' => '5.5'],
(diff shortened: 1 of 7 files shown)


config/php-cs-fixer.php
```diff
@@ -11,5 +11,17 @@
...
+            'php_unit_dedicate_assert' => ['target' => '5.0'],
+            'php_unit_expectation' => ['target' => '5.6'],
+            'php_unit_fqcn_annotation' => true,
+            'php_unit_method_casing' => true,
+            'php_unit_mock' => ['target' => '5.5'],
+            'php_unit_mock_short_will_return' => true,
+            'php_unit_namespaced' => ['target' => '5.7'],
+            'php_unit_set_up_tear_down_visibility' => true,
+            'php_unit_test_annotation' => ['style' => 'annotation'],
+            'php_unit_test_case_static_method_calls' => ['call_type' => 'self'],
```
sabberworm, line 25: Why not use
```suggestion
            'php_unit_test_case_static_method_calls' => ['call_type' => 'this'],
```
and leave the assertions as-is?

I think most people are used to calling assertions using `$this` and it’s also what the phpunit documentation does.

oliverklee: I talked to Sebastian Bergmann about this. He says calling the assert methods statically and non-statically both is valid (as PHP allows calling static methods non-statically), and it comes down to a matter of preference. He himself prefers the non-static calls because he likes doing calls statically only when this is technically necessary. I, on the other hand, prefer to keep things consistent: Always call static methods statically, and call non-static methods in a non-static way.

PR #303, opened 2021-07-02
Polish the README

oliverklee:
- fix some typos
- clean up the CSS and PHP examples
- clean up the Markdown formatting
- recommend `composer install` for installing

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,4 @@
-PHP CSS Parser
---------------
+# PHP CSS Parser
 
 [![Build Status](https://github.com/sabberworm/PHP-CSS-Parser/workflows/CI/badge.svg?branch=master)](https://github.com/sabberworm/PHP-CSS-Parser/actions/)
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -611,20 +661,30 @@ class Sabberworm\CSS\CSSList\Document#4 (2) {
 
 ## Contributors/Thanks to
```
sabberworm, line 662: I don’t like breaking up list items over multiple lines. Makes it hard to move them up or down in the editor (which is something I tend to do a lot more often with list items than with other paragraphs).

README.md
```diff
@@ -611,20 +661,30 @@ class Sabberworm\CSS\CSSList\Document#4 (2) {
...
```
sabberworm, line 663: You could add yourself here:

```suggestion

 * [oliverklee](https://github.com/oliverklee) for lots of refactorings, code modernizations and CI integrations
```

README.md
sabberworm: Come to think of it, I also prefer to rely on the editor’s soft-wrap behaviour for paragraphs. But my opinions on that are not as strong as on list items.

sabberworm (changes requested):

README.md
oliverklee: Rebased, changed and repushed.

sabberworm (approved):

PR #308, opened 2021-07-03
Sort the CHANGELOG entries chronologically

oliverklee:
There should be only one level of sorting, not two levels of sorting
with different directions (which made finding out what has changed
latest a real brain-twister).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,41 +1,10 @@
 # Revision History
 
-## x.y
-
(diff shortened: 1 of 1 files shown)


sabberworm:
The changelog was formatted like this because it provides a compromise between the desire to read changelogs chronologically and the desire not to have to scroll too much to find the starting point. But I agree, it can be confusing, so let’s change it.

PR #336, opened 2021-08-01
Install the development tools in the CI build the proper way

oliverklee:
- install all tools in one step instead of only one tool at a time
- install the exact registered version instead of updating
  (which avoid version-dependent differences between CI and the
  local environment)
- provide the fingerprints of the trusted GPG keys with which the
  PHARs are signed so that the faked pseudo-interactive "yes"
  confirmation is not needed anymore


sabberworm:
LGTM, thanks

PR #339, opened 2021-09-13
Fix comments being omitted in the middle of a rule

derschatta:
This is an updated version of #180 

We do rely on this patch, it would be great if it could be merged into the main repository.


scq:
This test now passes without any changes to the code on `main`, so I think this can be closed.

JakeQZ:
> This test now passes without any changes to the code on `main`, so I think this can be closed.

Thanks for checking up on this.

I recall a few issues with comment parsing being fixed, but don't know their issue/PR numbers offhand.

Rather than closing this immediately, I'd prefer to verify that the issue is ckecked with unit tests.

PR #350, opened 2021-12-21
Replace invalid `turns` unit with `turn`

westonruter:
This fixes merge conflicts for the key change in https://github.com/sabberworm/PHP-CSS-Parser/pull/193 which replaces the invalid `turns` init with `turn`, as identified by @pierlon:

> Also, there is no such unit named `turns`, but there is one called `turn` ([ref](https://drafts.csswg.org/css-values-3/#turn)), which I suppose is what was meant here 😄.

The use of `turns` in this codebase goes back [10 years](https://github.com/sabberworm/PHP-CSS-Parser/blame/7675640abec375a6c3c99590326bd78c3a7ec45d/lib/Sabberworm/CSS/Value/Size.php#L42).

### DIFF
--- a/src/Value/Size.php
+++ b/src/Value/Size.php
@@ -24,7 +24,7 @@ class Size extends PrimitiveValue
     /**
      * @var array<int, string>
      */
-    const NON_SIZE_UNITS = ['deg', 'grad', 'rad', 's', 'ms', 'turns', 'Hz', 'kHz'];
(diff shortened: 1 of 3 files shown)


oliverklee:
Can we cover this with a regression test?

westonruter:
@oliverklee I'm not sure what you mean. I added a test to ensure that `turn` is handled properly.

oliverklee:
@westonruter I'm sorry, I totally missed that the PR indeed includes a regression test. Thanks!

sabberworm:
Thanks!

PR #351, opened 2021-12-23
retain CSSList and Rule comments when rendering CSS

Ruud68:
Hi,
was banging my head against the wall as my css comments would vanish when using a autoprefixer repo that utilizes this repo.
Found out that the issue was that comments are only partly implemented in this repo: they are recognized, but when rendering the parsed css code they are not added back.

So I rolled up my sleeves and did this PR.

This PR does two things:
1. fix the comments implementation as only the first comments where taken and not the subsequent comments in a CSSList
2. adds the comments back into the rendered CSS (as also requested in #38 )

So this is my first PR in this repo, would love to have some feedback / discussion / tests :)

Thanks in advance,
regards, Ruud

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -128,7 +128,7 @@ abstract class CSSList implements Renderable, Commentable
                         $oParserState->currentLine()
                     );
                 }
-                $oParserState->setCharset($oAtRule->getCharset()->getString());
+                $oParserState->setCharset($oAtRule->getCharset());
             }
             return $oAtRule;
         } elseif ($oParserState->comes('}')) {
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -69,11 +69,9 @@ abstract class CSSList implements Renderable, Commentable
             $oParserState = new ParserState($oParserState, Settings::create());
         }
         $bLenientParsing = $oParserState->getSettings()->bLenientParsing;
-        $comments = [];
+        $aComments = [];
         while (!$oParserState->isEnd()) {
-            if (empty($comments)) {
-                $comments = $oParserState->consumeWhiteSpace();
-            }
+            $aComments = array_merge($aComments, $oParserState->consumeWhiteSpace());
             $oListItem = null;
             if ($bLenientParsing) {
                 try {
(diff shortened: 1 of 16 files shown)


oliverklee:
@Ruud68 Thanks for contributing! :heart:

Two things:
- If this PR fixes two problems, is there a way we could separate it into two PRs?
- I'd really like to have this covered with tests (that fail without the change and pass with it). Would you be willing to add those?

src/CSSList/CSSList.php
```diff
@@ -89,7 +89,6 @@ public static function parseList(ParserState $oParserState, CSSList $oList)
                 $oListItem->setComments($comments);
                 $oList->append($oListItem);
             }
-            $oParserState->consumeWhiteSpace();
```
Ruud68, line 92: Fix: remove this otherwise 'next' comment will be not set

src/Rule/Rule.php
```diff
@@ -106,7 +106,6 @@ public static function parse(ParserState $oParserState)
         while ($oParserState->comes(';')) {
             $oParserState->consume(';');
         }
-        $oParserState->consumeWhiteSpace();
```
Ruud68, line 109: Fix: remove this otherwise 'next' comment will be not set

Ruud68:
@oliverklee the fix is actually a very small one, I have added it as a comment to the code in the PR. 
As for the tests: I have no idea how to do that and how to test if the changes to the tests are correct. I need to check how that is done

oliverklee:
@Ruud68 I'll check what would be a good place for you to start with the tests.

Ruud68:
okay thanks @oliverklee I spent greater part of the day reverse engineering the code for these changes (not documented) and had a quick look at the unit tests. I do not think this will be a quick add for me (the tests) :(

Ruud68:
@oliverklee learned a lot :) did some small changes and fixed unittesting.... I hope :)

sabberworm:
Thanks @Ruud68 for the great work! I took the liberty of making rendering comments optional (by way of a `OutputFormat` setting), defaulting to off, for backwards-compatibility.

Ruud68:
> Thanks @Ruud68 for the great work! I took the liberty of making rendering comments optional (by way of a `OutputFormat` setting), defaulting to off, for backwards-compatibility.

My pleasure, glad I could contribute. making it optional is definitely a good option :)

PR #353, opened 2022-01-07
Do not allow string values for rules anymore

oliverklee:
String values had not been allowed for rules, and should not
be. (Passing string values was a bug in the Emogrifier
library.)

@see https://github.com/MyIntervals/emogrifier/pull/1144

This reverts commit 67a6e95f8ad2c9c3d8d745f1943eda8da81383a4.

### DIFF
--- a/src/Rule/Rule.php
+++ b/src/Rule/Rule.php
@@ -24,7 +24,7 @@ class Rule implements Renderable, Commentable
     private $sRule;
 
     /**
-     * @var RuleValueList|string|null
+     * @var RuleValueList|null
      */
     private $mValue;
 
@@ -171,7 +171,7 @@ class Rule implements Renderable, Commentable
     }
 
     /**
-     * @return RuleValueList|string|null
+     * @return RuleValueList|null
      */
     public function getValue()
     {
@@ -179,7 +179,7 @@ class Rule implements Renderable, Commentable
     }
 
     /**
-     * @param RuleValueList|string|null $mValue
+     * @param RuleValueList|null $mValue
      *
      * @return void
      */


sabberworm:
I’m not sure this is true. When the rule value is an identifier, it will be a string stored in `mValue`:

```css
.test {
white-space: nowrap;
}
```

will parse into the following `Rule`:

```
          object(Sabberworm\CSS\Rule\Rule)#8 (7) {
            ["sRule":"Sabberworm\CSS\Rule\Rule":private]=>
            string(11) "white-space"
            ["mValue":"Sabberworm\CSS\Rule\Rule":private]=>
            string(6) "nowrap"
            ["bIsImportant":"Sabberworm\CSS\Rule\Rule":private]=>
            bool(false)
            ["aIeHack":"Sabberworm\CSS\Rule\Rule":private]=>
            array(0) {
            }
            ["iLineNo":protected]=>
            int(2)
            ["iColNo":protected]=>
            int(19)
            ["aComments":protected]=>
            array(0) {
            }
          }
```

sabberworm:
AFAICT the type for `mValue` should be `Value|string`. IMHO we should also get rid of `null` and just use `""` for the default value.

JakeQZ:
> When the rule value is an identifier, it will be a string stored in `mValue`:

Would introducing an `Identifier` (or `Keyword`) subclass of `Value` help resolve this?  Such a class would just wrap a `string`, but also allow the unique `Value` superclass type to be used whenever a value is expected or provided (e.g. see #507 which results in lots of type specifications of `Value|string`).

> IMHO we should also get rid of `null` and just use `""` for the default value.

It seems that the default, with no value having been set, does not represent a valid property declaration.  This would not arise from parsing, since `Value::parseValue` would throw an exception, but may arise from manipulation by user code instantiating additional `Rule`s and failing to set their values.  `Rule`s with a missing value would currently be rendered as `key:;`, and discarded upon reparsing.

PR #359, opened 2022-03-22
Trying to reproduce a bug on php 8+ related to lenient parsing of invalid colors (at least)

mathroc:
We are getting warnings with some CSS code even on lenient mode on php 8.1 (but the same also happens on php 8.0)

> Warning: Uninitialized string offset 5 in /app/src/Value/Color.php on line 56


mathroc:
Hi! I wanted to showcase an issue but the tests aren't running automatically here. Does someone need to start them manually?

if you want to reproduce locally, this script causes the warning on php 8+:

```php
<?php

use Sabberworm\CSS\Parser;

include 'vendor/autoload.php';

$oParser = new Parser("a { color: #fffff;}", Settings::create());
$oParser->parse();
```

JakeQZ:
This should be fixed by #485, though I'd like to include your additional tests...

JakeQZ:
> the tests aren't running automatically here. Does someone need to start them manually?

GitHub doesn't allow them for first-time contributors until manually approved, as they have reportedly been used as a crypto-mining vector.  I have a Hotmail account for which most genuine email is dunked in the 'Spam' folder, whereas actual spam is not.  Draw your own conclusions.

mathroc:
Hi! I'm not on my laptop so I sync'ed the branch with the GitHub UI (hence the merge commit) I can rebase it properly later if needed.

still need a maintainer to approve running the unit tests

JakeQZ (changes requested):
> I sync'ed the branch with the GitHub UI (hence the merge commit) I can rebase it properly later if needed.

Thanks.  The GitHub UI sync seems to have been sufficient.  (I still don't know why I couldn't get the fork to rebase.)

I've now been able to confirm that these new tests would fail without the fix in #485, and pass with the fix in place.

I like the completeness of the test file, containing invalid color values with 1, 2, 4 and 5 characters.

Could you rename the test methods (to avoid the second one simply having the number `2` appeneded to the name)?  And resolve the PHP-CS-Fixer error?

Thanks again.

JakeQZ (approved):
All looks good now.  Thanks for contributing <3

PR #378, opened 2022-07-17
[TASK] Use PHP 8.0 for unit tests and static analysis on CI

oliverklee:


oliverklee:
It looks like this has to wait until we are using PHPUnit 8.

steamboatid:
please take a look at https://github.com/sabberworm/PHP-CSS-Parser/discussions/380

PR #385, opened 2022-09-17
More accurate parsing of invalid identifiers

raxbg:


raxbg:
Hey @oliverklee @JakeQZ I've caught this up with the latest main. Do you think this can be merged?

JakeQZ:
> Hey @oliverklee @JakeQZ I've caught this up with the latest main. Do you think this can be merged?

Thanks.  It's on the radar.  Review will require a bit of thinking.

JakeQZ:
I don't think this change is valid.  See [CSS nesting on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting/Using_CSS_nesting).

PR #386, opened 2022-09-17
More accurate parsing of invalid selectors

raxbg:


raxbg:
@oliverklee @JakeQZ Do you think this can be merged?

oliverklee (changes requested):
Hi @raxbg, thanks for all the hard work!

There seem to be some style checker warnings. Could you please have a look and fix them?

PR #387, opened 2022-09-17
Handle malformed RGB HEX values

raxbg:


JakeQZ (commented):
Hi @raxbg,

This has been fixed by #485 with an almost identacal code change.  Additional tests were added with #359.

I'm going through the open PRs starting with those more recent, so got to #485 before this.  It's reassuring to see that the change you proposed matches the change agreed upon there.

I think this can be closed as a duplicate, unless you think including the additional test here would add extra value - ?

Thanks.

raxbg:
> Hi @raxbg,
> 
> This has been fixed by #485 with an almost identacal code change. Additional tests were added with #359.
> 
> I'm going through the open PRs starting with those more recent, so got to #485 before this. It's reassuring to see that the change you proposed matches the change agreed upon there.
> 
> I think this can be closed as a duplicate, unless you think including the additional test here would add extra value - ?
> 
> Thanks.

Awesome! Closing this one.

PR #389, opened 2022-09-19
Parse simple expressions

### LINKED ISSUE #388: Expressions are not being parsed
Simple expressions do not seem to be parsed. Example:

```CSS
div {
    height: (vh - 10);
}
```

This ends up being treated as:
```CSS
div {}
```

raxbg:
Fixes #388 
Builds on top of #390

### DIFF
--- a/src/Value/CSSFunction.php
+++ b/src/Value/CSSFunction.php
@@ -3,6 +3,7 @@
 namespace Sabberworm\CSS\Value;
 
 use Sabberworm\CSS\OutputFormat;
+use Sabberworm\CSS\Parsing\ParserState;
 
 /**
  * A `CSSFunction` represents a special kind of value that also contains a function name and where the values are the
@@ -30,6 +31,55 @@ class CSSFunction extends ValueList
         $this->sName = $sName;
         $this->iLineNo = $iLineNo;
         parent::__construct($aArguments, $sSeparator, $iLineNo);
+    }
+
+    /**
+     * @param ParserState $oParserState
+     * @param bool $bIgnoreCase
+     *
+     * @return string
+     *
+     * @throws SourceException
+     * @throws UnexpectedEOFException
+     * @throws UnexpectedTokenException
+     */
+    public static function parseName(ParserState $oParserState, $bIgnoreCase = false)
+    {
+        return $oParserState->parseIdentifier($bIgnoreCase);
+    }
+
+    /**
+     * @param ParserState $oParserState
+     *
+     * @return array
+     *
+     * @throws SourceException
+     * @throws UnexpectedEOFException
+     * @throws UnexpectedTokenException
+     */
+    public static function parseArgs(ParserState $oParserState)
+    {
+        return Value::parseValue($oParserState, ['=', ' ', ',', '/', '*', '+', '-']);
+    }
+
+    /**
+     * @param ParserState $oParserState
+     * @param bool $bIgnoreCase
+     *
+     * @return CSSFunction
+     *
+     * @throws SourceException
+     * @throws UnexpectedEOFException
+     * @throws UnexpectedTokenException
+     */
+    public static function parse(ParserState $oParserState, $bIgnoreCase = false)
+    {
+        $mResult = self::parseName($oParserState, $bIgnoreCase);
+        $oParserState->consume('(');
+        $aArguments = self::parseArgs($oParserState);
+        $mResult = new CSSFunction($mResult, $aArguments, ',', $oParserState->currentLine());
+        $oParserState->consume(')');
+        return $mResult;
     }
 
     /**
--- a/src/Value/CSSFunction.php
+++ b/src/Value/CSSFunction.php
@@ -59,7 +59,7 @@ class CSSFunction extends ValueList
      */
     public static function parseArgs(ParserState $oParserState)
     {
-        return Value::parseValue($oParserState, ['=', ' ', ',', '/', '*', '+', '-']);
+        return Value::parseValue($oParserState, ['=', ' ', ',']);
     }
 
     /**
--- a/src/Value/Expression.php
+++ b/src/Value/Expression.php
@@ -24,7 +24,7 @@ class Expression extends CSSFunction
     public static function parse(ParserState $oParserState, $bIgnoreCase = false)
     {
         $oParserState->consume('(');
-        $aArguments = Value::parseValue($oParserState, ['=', ' ', ',']);
+        $aArguments = self::parseArgs($oParserState);
         $mResult = new Expression("", $aArguments, ',', $oParserState->currentLine());
         $oParserState->consume(')');
         return $mResult;
--- a/src/Value/Expression.php
+++ b/src/Value/Expression.php
@@ -1,0 +1,32 @@
+<?php
+
+namespace Sabberworm\CSS\Value;
+
+use Sabberworm\CSS\OutputFormat;
+use Sabberworm\CSS\Parsing\ParserState;
+
+/**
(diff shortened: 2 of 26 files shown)


tests/ParserTest.php
```diff
@@ -1180,4 +1238,28 @@ public function lonelyImport()
...
         self::assertSame($sExpected, $oDoc->render());
     }
+
+    /**
+     * @test
+     */
+    public function functionArithmeticInFile()
+    {
+        $oDoc = self::parsedStructureForFile('function-arithmetic', Settings::create()->withMultibyteSupport(true));
+        $sExpected = 'div {height: max(300,vh + 10);}
```
sabberworm, line 1248: AFAICT, vh without a number in front is still not valid and I’d prefer it would throw an error in strict mode.

raxbg: Indeed, it is invalid, however this seems to be outside of the scope of these changes. I guess we need a separate issue for this. Also is it a concern of the parser to validate the semantic meaning of identifiers it encounters?

westonruter:
For what it's worth, I'm including this patch in https://github.com/ampproject/amp-wp/pull/7487 and it is passing all tests in our project.

westonruter:
@oliverklee was this closed as well as other PRs (e.g. #185 and #193) just because the `master` branch was deleted?

oliverklee:
@westonruter Oh no! It indeed was. ;-(

oliverklee:
I have now reinstated the legacy `master` branch and will reopen the PRs in a minute. So sorry for this.

sabberworm:
> I have now reinstated the legacy `master` branch and will reopen the PRs in a minute. So sorry for this.

I would prefer we delete the `master` branch and just change all the PRs to point to `main` instead.

oliverklee:
I have now moved all PRs over to `main` and dropped the `master` branch again.

tests/ParserTest.php
JakeQZ: > Also is it a concern of the parser to validate the semantic meaning of identifiers it encounters?

I agree it's not.

Moving forwards, we've agreed to replace strict mode parsing with a logging mechanism.  And try to replicate what most browsers do when encountering invalid syntax.  A unit with no number in front should therefore probably be parsed as zero of that unit.  And, as a bonus, a separate change would log the lack of a number.

So, for now, we don't need to worry about strict mode, but should create issues for things that could be logged in future.

src/Value/CSSFunction.php
```diff
@@ -33,6 +33,35 @@ public function __construct($sName, $aArguments, $sSeparator = ',', $iLineNo = 0
...
+     *
+     * @throws SourceException
+     * @throws UnexpectedEOFException
+     * @throws UnexpectedTokenException
+     */
+    public static function parseArgs(ParserState $oParserState)
+    {
+        return Value::parseValue($oParserState, ['=', ' ', ',']);
+    }
+
```
JakeQZ, line 64: Nice refactoring.  Could this be put through as a separate PR?

JakeQZ (commented):
I like this.  I'd be happy for this to be merged.  @oliverklee, @sabberworm WDYT?

JakeQZ (changes requested):
We need a changelog entry, though.

src/Value/Expression.php
```diff
@@ -0,0 +1,32 @@
...
+     */
+    public static function parse(ParserState $oParserState, $bIgnoreCase = false)
+    {
+        $oParserState->consume('(');
+        $aArguments = self::parseArgs($oParserState);
+        $mResult = new Expression("", $aArguments, ',', $oParserState->currentLine());
+        $oParserState->consume(')');
+        return $mResult;
+    }
+}
```
JakeQZ, line 32: The should be a class-specific unit test for this class.  The overall parser tests are good, but we also want to make sure each individual class behaves as it should.

JakeQZ (changes requested):

src/Value/Expression.php
JakeQZ: @raxbg, would you be willing to create a corresponding `TestExpression` class to test the new class as a standalone?

JakeQZ:
> I like this. I'd be happy for this to be merged. @oliverklee, @sabberworm WDYT?

On further looking, a class-specific unit test is also needed for the new class (inherited from the terribly misnomered `TestCase` - Rename That!).

oliverklee:
Also, according to the original commit message, this builds on on top of #390. So we should get that in first and then rebase.

oliverklee (changes requested):
Let's get #390 in first.

src/Value/Expression.php
raxbg: > There should be a class-specific unit test for this new class. The overall parser tests are good, but we also want to make sure each individual class behaves as it should.

Makes sense. How does that look like: 14118a20cee6273a33087b305f026aefcea77989

raxbg:
#390 is closed now and seems like it's no longer needed. There shouldn't be anything else blocking the merge of this PR :)

Let me know if you have any concerns

milindmore22:
Please, can anyone with permission take care of conflicts?

JakeQZ:
> #390 is closed now and seems like it's no longer needed. There shouldn't be anything else blocking the merge of this PR :)
> 
> Let me know if you have any concerns

In response to [this comment](https://github.com/MyIntervals/PHP-CSS-Parser/pull/389#issuecomment-2220353997) and the latest commits, I gave permission for the continuous integration checks to run for this PR (sadly GitHub does not automatically run them for external contributors).  This found some issues that would initially need resolving prior to review.

I assumed @raxbg would be notified about this, and fix the failing tests, so did not comment further.

> Please, can anyone with permission take care of conflicts?

Now some time has passed, and many other small changes have been made, so rebasing/merging to the latest `main` may not be straightforward.

@milindmore22, normally we would ask the PR author to merge/rebase if needed, but that's in the normal workflow of a few days to review a PR.

@raxbg, do you have time available to rebase/merge and resolve the CI build failures?  Don't worry if not; we can take what you've done here and resolve the issues internally (there are two additional collaborators on the project since this PR was originally created).  Let us know either way.

tests/ParserTest.php
JakeQZ: > Also is it a concern of the parser to validate the semantic meaning of identifiers it encounters?

Looking again, this is syntax, which is a concern of the parser.  A unit without a length is syntactically invalid, and should be discarded (along with the entire property declaration).

JakeQZ:
> Please, can anyone with permission take care of conflicts?

I've rebased this at #1437 to move this along a bit.  Don't know if that actually works yet.  (It wasn't straightfoward with the branch including many commits, some of which were already merged, nor with the variable name changes away from Hungarian notation.)

JakeQZ:
I can't find any specification reference that indicates this should be allowed or possible, though I'm sure I've seen it before.

Unless anyone can find it, I suggest this PR be closed as invalid.

oliverklee:
>  (It wasn't straightfoward with the branch including many commits, some of which were already merged, nor with the variable name changes away from Hungarian notation.)

In my experience, rebasing it a lot less of a hassle when the branch that is to be rebased has fewer commits, i.e., squashing before rebasing usually is helpful.

oliverklee:
> I can't find any specification reference that indicates this should be allowed or possible, though I'm sure I've seen it before.
>
> Unless anyone can find it, I suggest this PR be closed as invalid.

I also can only find the [`calc()` syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc).

I agree that we should close this as invalid.

Closing it now.

PR #169, opened 2019-09-18
Fix/invalid calc parsing

raxbg:
This PR adds better handling of invalid `calc` definitions.

PR #390, opened 2022-09-19
Fix/arithmetic in functions

raxbg:
Builds on top of #169


tests/fixtures/infinite-loop.css
```diff
@@ -0,0 +1,3 @@
+div {
+    height: ///!;
```
JakeQZ, line 2: Nice.

JakeQZ (changes requested):
Functional changes are all good.  A few minor side issues though.

@oliverklee, would you like to see if there is anything else?

I'm aware of two things that I think are beyond the scope of this PR, and part of broader changes we want to make moving forwards:
- We want to move the test data from files to be inlined with the relevant test code.
- We want to have class-specific tests as well as general parsing tests (but as yet there is no `ValueTest` `TestCase`).

oliverklee (changes requested):
Please let's move the refactoring to a separate change (either as a pre-patch, or as a post-PR cleanup). This allows us to stick to the "only one kind of change per PR/commit" rule of thumb, and also to selectively backport or revert commits.

nickfmc:
do we have an ETA on this one? not having the calc and clamp functions work is a huge issue in 2024

oliverklee:
@nickfmc How fast we progress with this PR depends on whether someone (TM) would be willing to invest time into it.

Would you be willing to create a PR (or multiple PRs) for the refactoring contained in this PR here so we get them out of the way, and then take over this PR?

JakeQZ:
I looked at rebasing this.  I don't think it's possible.  There have been too many changes since.

I'd like to be able to see the full diff on GitHub without the review comments, in order to re-apply the changes.  But I can't find how to do that.

JakeQZ:
> I'd like to be able to see the full diff on GitHub without the review comments, in order to re-apply the changes. But I can't find how to do that.

I found the '[Don't] Show Comments' option, it's on the '•••` menu, on a per-file basis.  I should be able to take this on as fresh PRs, unless @nickfmc you would like to?!  Thanks for the nudge :)

JakeQZ:
> do we have an ETA on this one? not having the calc and clamp functions work is a huge issue in 2024

Upon checking, I find that `calc` is already parsed correctly.  `clamp` I'm not so sure about - `calc` is specifically handled including with vendor prefixes.  Constructs like `max(300px, 50vw + 10px)` definitely fail at present.

oliverklee:
This PR now needs a rebase.

JakeQZ:
> This PR now needs a rebase.

I already looked at rebasing, but it seemed a bit of a nightmare with the number of changes since, particularly given I'd be working on someone else's fork, and that in any case we wanted to split this up into three more atomic changes.  So I am reimplementing the changes from here as fresh PRs: #599, #607, and the infinite loop fix TBD.

nickfmc:
Does this ticket address the lack of CSS feature clamp working correctly? Or do we need a new ticket for that?
`font-size: clamp(2.19rem,0.5vw + 2.06rem,2.5rem);`

JakeQZ:
> Does this ticket address the lack of CSS feature clamp working correctly? Or do we need a new ticket for that? `font-size: clamp(2.19rem,0.5vw + 2.06rem,2.5rem);`

I'll add a test for that and get back to you...

JakeQZ:
> > Does this ticket address the lack of CSS feature clamp working correctly? Or do we need a new ticket for that? `font-size: clamp(2.19rem,0.5vw + 2.06rem,2.5rem);`
> 
> I'll add a test for that and get back to you...

Good news.  Yes, it does.  Or at least I'm 99.99% sure - I extended the test previously added to include `clamp`; the exact expression tested in the 2nd argument doesn't quite match yours (and doesn't include an `rem` value), but I can't think of any reason why your specific example could fail whilst the expression used in the test works.  See #620.

JakeQZ:
> Good news. Yes, it does. Or at least I'm 99.99% sure - I extended the test previously added to include `clamp`; the exact expression tested in the 2nd argument doesn't quite match yours (and doesn't include an `rem` value), but I can't think of any reason why your specific example could fail whilst the expression used in the test works. See #620.

PS. What may still not work is expressions involving brackets, like `(1rem + 1vw) * 2`, though I don't honestly know at time of writing.

@nickfmc, would you be able to try the [9.0.x-dev](https://packagist.org/packages/sabberworm/php-css-parser#dev-main) branch to confirm this fixes the `clamp` issue for you, and let us know if you find any other issues?  Thanks.

JakeQZ:
> Reopening, as there's still the infinite loop bugfix from this to be applied.

Upon investigation, I find the infinite loop bug was subsequently fixed by #413.  Since we want to move away from having test data in separate files, I am not going to include the additional test for this as it is.  I looked at adding a more targeted test, and found a curious behaviour, which I logged as #622.  Thus, I am re-closing this in favour of that.

PR #395, opened 2022-10-04
Run the tests on CI with all warnings enabled

oliverklee:
This allows us to catch things that will break in future PHP versions, but still kind-of-work, e.g., deprecation warnings or incorrect types.


oliverklee:
AFAIK, the warnings will indeed cause the tests to fail.

sabberworm:
> AFAIK, the warnings will indeed cause the tests to fail.

Ah, OK, good

PR #409, opened 2023-02-03
Fix PHP8.2 str_split function returns empty arrays for empty strings

HuongNV13:
In PHP 8.2, the str_split function will returns empty arrays for empty strings.
See: https://php.watch/versions/8.2/str_split-empty-string-empty-array

We can use mb_str_split() instead

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -520,7 +520,7 @@ class ParserState
             if ($sString === '') {
                 return [];
             } else {
-                return str_split($sString);
+                return mb_str_split($sString);
             }
         }
     }


src/Parsing/ParserState.php
```diff
@@ -520,7 +520,7 @@ private function strsplit($sString)
             if ($sString === '') {
                 return [];
             } else {
-                return str_split($sString);
+                return mb_str_split($sString);
```
oliverklee, line 523: The check in line 520 ensures that this `str_split` call will only happen with a none-empty string. So I don't understand why the behavior with an empty string would be a problem here. Am I missing something?

sabberworm:
The linked article also mentions that `mb_str_split` has always behaved this way (returning an empty array), so if the behavior in `strsplit` were indeed broken in PHP 8.2, it would now be broken in all versions.

src/Parsing/ParserState.php
HuongNV13: Yes, you're correct @oliverklee.
Sorry for the noise.
I'm going to close this PR

Thanks,

PR #413, opened 2023-06-10
[TASK] Avoid poor scaling of `array_search()` with very long arrays

bartbutler:
When there are many many elements in $aStack, starting the delimiter search from the beginning for each loop is very very slow. I addressed this by building a new array rather than modifying $aStack in place and iterating over it in a single pass. The particular 1.6M style string that was giving me trouble went from taking 4 minutes to parse to 5 seconds.

### DIFF
--- a/src/Value/Value.php
+++ b/src/Value/Value.php
@@ -70,20 +70,26 @@ abstract class Value implements Renderable
             if (count($aStack) === 1) {
                 return $aStack[0];
             }
-            $iStartPosition = null;
-            while (($iStartPosition = array_search($sDelimiter, $aStack, true)) !== false) {
+            $aNewStack = [];
+            for ($iStartPosition = 0; $iStartPosition < count($aStack); $iStartPosition += 1) {
+                if ($iStartPosition === (count($aStack) - 1) || $sDelimiter !== $aStack[$iStartPosition + 1]) {
+                    $aNewStack[] = $aStack[$iStartPosition];
+                    continue;
+                }
                 $iLength = 2; //Number of elements to be joined
-                for ($i = $iStartPosition + 2; $i < count($aStack); $i += 2, ++$iLength) {
+                for ($i = $iStartPosition + 3; $i < count($aStack); $i += 2, ++$iLength) {
                     if ($sDelimiter !== $aStack[$i]) {
                         break;
                     }
                 }
                 $oList = new RuleValueList($sDelimiter, $oParserState->currentLine());
-                for ($i = $iStartPosition - 1; $i - $iStartPosition + 1 < $iLength * 2; $i += 2) {
+                for ($i = $iStartPosition; $i - $iStartPosition < $iLength * 2; $i += 2) {
                     $oList->addListComponent($aStack[$i]);
                 }
-                array_splice($aStack, $iStartPosition - 1, $iLength * 2 - 1, [$oList]);
+                $aNewStack[] = $oList;
+                $iStartPosition += $iLength * 2 - 2;
             }
+            $aStack = $aNewStack;
         }
         if (!isset($aStack[0])) {
             throw new UnexpectedTokenException(
--- a/src/Value/Value.php
+++ b/src/Value/Value.php
@@ -66,8 +66,8 @@ abstract class Value implements Renderable
             $oParserState->consumeWhiteSpace();
         }
         // Convert the list to list objects
-        $iStackLength = count($aStack);
         foreach ($aListDelimiters as $sDelimiter) {
+            $iStackLength = count($aStack);
             if ($iStackLength === 1) {
                 return $aStack[0];
             }
--- a/src/Value/Value.php
+++ b/src/Value/Value.php
@@ -71,13 +71,14 @@ abstract class Value implements Renderable
                 return $aStack[0];
             }
             $aNewStack = [];
-            for ($iStartPosition = 0; $iStartPosition < count($aStack); ++$iStartPosition) {
-                if ($iStartPosition === (count($aStack) - 1) || $sDelimiter !== $aStack[$iStartPosition + 1]) {
+            $aStackLength = count($aStack);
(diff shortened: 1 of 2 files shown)


src/Value/Value.php
```diff
@@ -70,20 +70,26 @@ public static function parseValue(ParserState $oParserState, array $aListDelimit
             if (count($aStack) === 1) {
                 return $aStack[0];
             }
-            $iStartPosition = null;
-            while (($iStartPosition = array_search($sDelimiter, $aStack, true)) !== false) {
+            $aNewStack = [];
+            for ($iStartPosition = 0; $iStartPosition < count($aStack); $iStartPosition += 1) {
```
JakeQZ, line 74:
```suggestion
            for ($iStartPosition = 0; $iStartPosition < count($aStack); ++$iStartPosition) {
```

JakeQZ (changes requested):
Quite a complicated bit of code to understand, but I think I'm satisfied that it's functionally equivalent before and after this change, and that it indeed should now run in more like O(n) time than O(n2).

`$aStack` is an odd-length array containing an alternating series of values and delimiters (perhaps operators).  The order in which the delimiters are processed is important, as it represents precedence, so the outer loop is necessary.

I wondered whether the inner loop could start at position 1 and step by 2, so it only looked at delimiters.  In that case it would need to copy the previous value and the delimiter to the new stack, and special-case the final delimiter to also copy the final value (if the delimiter doesn't match).  Might be slightly more optimal and marginally clearer.

Otherwise, `++` can be used instead of `+= 1`.

Also, I think this change is worthy of a changelog entry.

src/Value/Value.php
```diff
@@ -70,20 +70,26 @@ public static function parseValue(ParserState $oParserState, array $aListDelimit
             if (count($aStack) === 1) {
                 return $aStack[0];
             }
-            $iStartPosition = null;
-            while (($iStartPosition = array_search($sDelimiter, $aStack, true)) !== false) {
+            $aNewStack = [];
+            for ($iStartPosition = 0; $iStartPosition < count($aStack); ++$iStartPosition) {
```
oliverklee, line 74: `count` is relatively expensive. We should move it out of the loop and store the number of elements in a local variable to avoid calling `count` on each loop iteration.

bartbutler: Good point, fixed.

src/Value/Value.php
```diff
@@ -70,20 +70,27 @@ public static function parseValue(ParserState $oParserState, array $aListDelimit
             if (count($aStack) === 1) {
                 return $aStack[0];
             }
-            $iStartPosition = null;
-            while (($iStartPosition = array_search($sDelimiter, $aStack, true)) !== false) {
+            $aNewStack = [];
+            $aStackLength = count($aStack);
```
JakeQZ, line 74: `$aStackLength` can be initialized before and used in the preceding early-return test.

src/Value/Value.php
JakeQZ: > `count` is relatively expensive.

[It is actually O(1)](https://stackoverflow.com/questions/5835241/is-phps-count-function-o1-or-on-for-arrays/5835419#5835419), but noneless [slightly quicker with the result cached in a local variable (maybe 50µs per reference in 2014](https://stackoverflow.com/a/23410780/2511031).

src/Value/Value.php
JakeQZ: And maybe should be named `$iStackLength`.

bartbutler: Yep, good catch. I had forgotten the variable naming scheme.

JakeQZ: That variable naming scheme may be changed in future - https://github.com/MyIntervals/PHP-CSS-Parser/issues/450#issuecomment-1931576970 - though until such time it's best to maintain consistency.

src/Value/Value.php
```diff
@@ -66,24 +66,31 @@ public static function parseValue(ParserState $oParserState, array $aListDelimit
             $oParserState->consumeWhiteSpace();
         }
         // Convert the list to list objects
+        $iStackLength = count($aStack);
         foreach ($aListDelimiters as $sDelimiter) {
```
JakeQZ, line 70: I think `$iStackLength` needs to be set at the start of the outer loop, not before it.  The stack is updated at the end of the outer loop with `$aStack = $aNewStack`, and may now be shorter for the next iteration if some elements could be joined for the specific delimiter.

bartbutler: Yep, you're right. Fixed.

JakeQZ (approved):
The changes look fine to me now.

@oliverklee, @sabberworm, do you have anything to add?  Also wondered if there should be a `CHANGELOG.md` entry - it's not a change in functioality, just a performance improvement.

oliverklee:
I don't have anything to add. And yes, I a changelog entry would be good for this.

sabberworm (approved):

JakeQZ (changes requested):
Hi @bartbutler,

Thanks for making the suggested code changes.

Would you also be able to add an entry to `CHANGELOG.md` (in the root directory), briefly describing the change.  It could probably either be in the 'Fixed' or 'Changed' sections, whichever you think is more appropraite.

(I'm guessing the slowdown would have resulted from CSS property values with long comma-sparated lists, such as `box-shadow`, or perhaps gradient fills, though would be intrigued to know what was in the 1.6Mb stylesheet specifically causing the slowdown, if indeed you do - I appreciate that using a profiler may not have provided that info.)

bartbutler:
I thought that I had added the example as a test but I guess I reconsidered, probably because it came into my possession as an email so there were privacy implications. IIRC it was someone who used 1.6MB 1px elements attached to rgb() functions to recreate a small photo using pure CSS.

JakeQZ (approved):
> IIRC it was someone who used 1.6MB 1px elements attached to rgb() functions to recreate a small photo using pure CSS.

Wow!

Changes all look good now.  Thanks for contributing <3

Issue #251, opened 2021-06-03
Discussion: Bump the minimal required PHP version?


oliverklee:
Currently, this library's PHP version requirements align to those of the WordPress project (i.e., PHP >= 5.6 as a hard requirement according to https://wordpress.org/support/article/requirements/).

I've done some googling around and found contradicting information on when the PHP version requirements for WordPress will change, and which PHP version will then be required. @westonruter Do you have more (or official) information on this which you could share?

westonruter:
There's an in-depth proposal for dropping support for old PHP versions via a fixed schedule: https://make.wordpress.org/core/2020/08/24/proposal-dropping-support-for-old-php-versions-via-a-fixed-schedule/

Matt Mullenweg's response to the proposal can be seen here: https://make.wordpress.org/core/2020/08/24/proposal-dropping-support-for-old-php-versions-via-a-fixed-schedule/#comment-39736

In short:

> The right policy for WordPress core is to have users drive what we support, measured by usage, not other projects with their own priorities and stakeholders.

It seems that once usage for PHP 5.x goes below 5% that this will then allow for WordPress to bump the minimum PHP requirement to PHP 7.

At the moment, PHP 5.6 accounts for 9.2% of WordPress installs: https://wordpress.org/about/stats/

![image](https://user-images.githubusercontent.com/134745/120691172-f87d6400-c45a-11eb-8302-772bd4bd0414.png)

Version	| Usage
--------|------
8.0 | 0.50%
7.4 | 30.80%
7.3 | 29.08%
7.2 | 16.42%
7.1 | 4.17%
7.0 | 5.53%
5.6 | 9.20%
5.5 | 0.96%
5.4 | 1.43%
5.3 | 1.35%
5.2 | 0.56%

So we're waiting on hosts to bump the PHP versions. 

There are 93 plugins in the WP directory that use PHP-CSS-Parser: https://wpdirectory.net/search/01F79GKCFAQRTSWMTQ07H7KVQS

I checked the ones with 20,000+ active installs and there's a split between the ones that require PHP 5 vs PHP 7:

Plugin | Requires PHP
-------|--------------
AMP | 5.6
Admin Menu Editor | 5.x
WooCommerce PDF Invoices & Packing Slips | 7.1
MailPoet – emails and newsletters in WordPress | 7.1
Advanced Contact form 7 DB | n/a
Jupiter X Core | 5.6
wpDataTables – Tables & Table Charts | 5.6
[...]

PR #414, opened 2023-06-10
Update PHPUnit to 10.2

bartbutler:
Running the test suite was giving me trouble by requiring PHPUnit 5.7, which is quite old. This MR makes the minimal changes necessary to update to 10.2 without modifying the test content in any substantive way.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "ext-iconv": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.27",
+        "phpunit/phpunit": "^10.2",
         "codacy/coverage": "^1.4.3"
     },
     "suggest": {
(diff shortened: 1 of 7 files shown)


sabberworm:
I think we decided on sticking with 5.7 because it supported all the PHP versions we wanted to test on.

bartbutler:
Of course, that's your prerogative, just noting that I had trouble with it on PHP 8.2 for some reason. and arguably nobody should be using PHP 5 at this point given the vast improvement from 5 to 7.

sabberworm:
I definitely think we can finally tackle #251.

bartbutler:
Looks like PHP 5 might finally be below 5% for WP. Anyway, when you decide to do it, this MR is here.

oliverklee:
Hey @bartbutler, I'd like to close this in favor of the (merged) #424. As we're keeping support for PHP 7.2 at the moment, we cannot go higher than PHPUnit 8.

I'd still like to have static data providers. Would you be willing to create a new PR for that?

bartbutler:
Took a look and seems to have already been done, so nothing to do!

PR #417, opened 2023-12-05
Drop support for PHP < 7.2

oliverklee:


oliverklee:
Note: Before this is merged, version 8.4.0 should be released (as this is a breaking change, which requires a new major version).

oliverklee:
Closing - I'm going to re-create the PR without my fork.

PR #427, opened 2024-02-02
[TASK] Move the coverage generation to a separate CI workflow

oliverklee:
This makes the CI configuration easier to understand.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,11 +41,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: [ '7.2', '7.3' ]
-        coverage: [ 'none' ]
-        include:
-          - php-version: '7.4'
-            coverage: xdebug
+        php-version: [ '7.2', '7.3', '7.4' ]
 
     steps:
       - name: Checkout
@@ -57,7 +53,50 @@ jobs:
           php-version: ${{ matrix.php-version }}
           ini-values: error_reporting=E_ALL
           tools: composer:v2
-          coverage: "${{ matrix.coverage }}"
+          coverage: none
+
+      - name: Show the Composer configuration
+        run: composer config --global --list
+
+      - name: Cache dependencies installed with composer
+        uses: actions/cache@v4
+        with:
+          path: ~/.cache/composer
+          key: php${{ matrix.php-version }}-composer-${{ hashFiles('**/composer.json') }}
+          restore-keys: |
+            php${{ matrix.php-version }}-composer-
+
+      - name: Install Composer dependencies
+        run: |
+          composer update --with-dependencies --no-progress;
+          composer show;
+
+      - name: Run Tests
+        run: ./vendor/bin/phpunit --coverage-clover build/coverage/xml
+
+  coverage:
+    name: Coverage
+
+    runs-on: ubuntu-22.04
+
+    needs: [ php-lint ]
+
+    strategy:
+      fail-fast: false
+      matrix:
+        php-version: [ '7.4' ]
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          ini-values: error_reporting=E_ALL
+          tools: composer:v2
+          coverage: xdebug
 
       - name: Show the Composer configuration
         run: composer config --global --list
@@ -81,7 +120,7 @@ jobs:
       - name: Upload coverage results to Codacy
         env:
           CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
-        if: "${{ matrix.coverage != 'none' && env.CODACY_PROJECT_TOKEN != '' }}"
+        if: "${{ env.CODACY_PROJECT_TOKEN != '' }}"
         run: |
           ./vendor/bin/codacycoverage clover build/coverage/xml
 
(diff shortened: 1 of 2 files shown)


JakeQZ (commented):
I'm going to rebase this so that I can check it properly...

I'm looking for no overall change of behaviour for this commit, after which it will be clearer what is being changed in the CI jobs, when they are changed.

JakeQZ (changes requested):
That was an educational review.  I understand now how the `matrix` works with `include`, but haven't quite got to grips with the [`with` parameter](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepswith).  I don't understand what uses it and how.  Maybe best keep those in for now.

I think you've added PHP 7.4 to the unit tests as well as code coverage, and would prefer to keep this PR to a refactoring only (with no overall change).  I think a command line argument for code-coverage can now be dropped from the original unit tests section.

The number of CI tests run should be the same as before.

.github/workflows/ci.yml
```diff
@@ -39,11 +39,50 @@ jobs:
...
+          restore-keys: |
+            php${{ matrix.php-version }}-composer-
+
+      - name: Install Composer dependencies
+        run: |
+          composer update --with-dependencies --no-progress;
+          composer show;
+
+      - name: Run Tests
+        run: ./vendor/bin/phpunit --coverage-clover build/coverage/xml
```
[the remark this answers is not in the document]
oliverklee: We can indeed. Done.

.github/workflows/ci.yml
```diff
@@ -39,11 +39,50 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        php-version: [ '7.2', '7.3' ]
-        coverage: [ 'none' ]
-        include:
-          - php-version: '7.4'
-            coverage: xdebug
+        php-version: [ '7.2', '7.3', '7.4' ]
```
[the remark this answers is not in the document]

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
         coverage: [ 'none' ]
```
JakeQZ, line 43: I don't think we need this any more.  It's merely changing a 2-row vector into a 1x2 matrix, with a dimension heading that isn't used anywhere - IIUC it sets the environment variable `INPUT_COVERAGE` as well as making available `{{matrix.coverage}}` to the script, neither of which would now be used.

.github/workflows/ci.yml
```diff
@@ -43,7 +43,7 @@ jobs:
...
         include:
           - php-version: '7.4'
-            coverage: xdebug
+            coverage: none
```
JakeQZ, line 46: We don't need this any more either.  We should add PHP versions 7.4 through 8.3 for unit tests as a separate PR.  Having this is effectively the same has changing to `php-version: [ '7.2', '7.3', '7.4' ]` above.

For this PR to be 'no overall change', there should still be 26 checks.  There are 27 as it stands.

JakeQZ (changes requested):
I like the separation with a new workflow file.  Much cleaner and more maintainable - hopefully.

A couple of gremlins still need removing from the original `ci.yml`.

.github/workflows/ci.yml
oliverklee: Done.

.github/workflows/ci.yml
oliverklee: > We should add PHP versions 7.4 through 8.3 for unit tests as a separate PR. 

Yes, absolutely.

JakeQZ (commented):
Can we lose the `coverage: none` from the original `ci.yml`?  I don't think it's needed any more.

JakeQZ:
Also, we are now not getting a 'Code coverage' on `push`, it seems.

.github/workflows/codecoverage.yml
```diff
@@ -0,0 +1,57 @@
+# https://help.github.com/en/categories/automating-your-workflow-with-github-actions
+
+on:
+  push:
+    branches:
+        - master
```
JakeQZ, line 6: Is the indentation wrong here?  Something seems to preventing a match.

JakeQZ: Extra spaces should not matter.  It looks correct, so IDK why it's not working.

.github/workflows/ci.yml
```diff
@@ -39,11 +39,50 @@ jobs:
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Install PHP
+        uses: shivammathur/setup-php@v2
+        with:
+          php-version: ${{ matrix.php-version }}
+          ini-values: error_reporting=E_ALL
+          tools: composer:v2
+          coverage: none
```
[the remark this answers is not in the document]
oliverklee: Setting `coverage: none` has the result of no code coverage tool (like Xdebug of PCOV) getting loaded, which should speed up the non-coverage tasks a bit.

.github/workflows/codecoverage.yml
oliverklee: I've just rebased and repushed, an the coverage job seems to have run fine for me.

JakeQZ: Can't see a problem now, though the checks run this time have been the `pull_request` checks rather than the `push` checks.  I'll approve and hopefully will see some checks run on 'push' from the merge.

Oh, I've just realized why the code coverage wasn't run - you're pushing to the `task/coverage-separate` branch, not `master` - it's only supposed to run on pushes to `master`.  D'oh, sorry.

oliverklee: The CI should have run for the `pull_request` trigger, though.

JakeQZ: > The CI should have run for the `pull_request` trigger, though.

Strange.  When I reviewed last night, it definitely showed 12 succussful checks on 'push', rather than 13 on 'pull_request'.  I wouldn't have spotted this potential problem otherwise.

PR #428, opened 2024-02-02
[TASK] Add the new maintainers to the `composer.json`

oliverklee:


composer.json
```diff
@@ -12,6 +12,14 @@
...
         {
             "name": "Raphael Schweikert"
+        },
+        {
+            "name": "Oliver Klee",
+            "email": "github@oliverklee.de"
+        },
+        {
+            "name": "Jake Hotson",
+            "email": "jake@qzdesign.co.uk"
```
JakeQZ, line 22: Could you please use `jake.github@qzdesign.co.uk` here?  It helps reduce spam (or at least let me know where it came from).

JakeQZ (changes requested):
I prefer a GitHub-specific email address here.

composer.json
oliverklee: Done.

JakeQZ (approved):

PR #429, opened 2024-02-03
[BUGFIX] Git-ignore the PHPUnit result cache

oliverklee:
The corresponding file gets created by more recent versions of PHPUnit.


JakeQZ (approved):
This is presumably so `git add *.*` won't inadvertently add such files to version control.

oliverklee:
Yes, exactly.

PR #436, opened 2024-02-03
[TASK] Configure PHPUnit some more

oliverklee:
- make things more strict
- stop caching the result (as we don't need this at the moment)
- make things more colorful


JakeQZ (approved):
It could be argued that all of these settings should be the default.

oliverklee:
> It could be argued that all of these settings should be the default.

Sebastian (the author of PHPUnit) actually sometimes actually does this for newer versions of PHPUnit. :-)

PR #437, opened 2024-02-03
[BUGFIX] Avoid double CI runs for pull requests

oliverklee:


.github/workflows/ci.yml
```diff
@@ -1,8 +1,10 @@
 # https://help.github.com/en/categories/automating-your-workflow-with-github-actions
 
 on:
-  pull_request:
   push:
+    branches:
+      - main
```
JakeQZ, line 6: This should currently be `master`.  The repo hasn't had the anti-slavery rename yet.

JakeQZ (changes requested):
Rename of main branch to `main` still needs to be done, unless I've missed that ticket.

JakeQZ:
cf https://github.com/MyIntervals/emogrifier/pull/1251

.github/workflows/ci.yml

JakeQZ (approved):

PR #438, opened 2024-02-03
[TASK] Move PHPStan from PHIVE to Composer

oliverklee:
This allows us to have PHPStan extensions, e.g., for PHPUnit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -130,7 +130,7 @@ jobs:
 
       - name: Install development tools
         run: |
-          phive --no-progress install --trust-gpg-keys BBAB5DF0A0D6672989CF1869E82B2FB314E9906E,95DE904AB800754A11D80B605E6DDE998AB73B8E,CA7C2C7A30C8E8E1274A847651C67305FFC2E5C0
+          phive --no-progress install --trust-gpg-keys BBAB5DF0A0D6672989CF1869E82B2FB314E9906E,95DE904AB800754A11D80B605E6DDE998AB73B8E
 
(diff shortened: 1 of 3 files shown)


JakeQZ (approved):
Personally, on Windows, I've found PHIVE to be a [PHAFF](https://dictionary.cambridge.org/dictionary/english/faff), though perhaps not fully understood the point of using it.  It seems simpler to me if we just use Composer.

oliverklee:
> Personally, on Windows, I've found PHIVE to be a [PHAFF](https://dictionary.cambridge.org/dictionary/english/faff), though perhaps not fully understood the point of using it. It seems simpler to me if we just use Composer.

The point was to avoid (or get out of) [dependency hell](https://speakerdeck.com/oliverklee/dependency-hell-composer-and-phive), i.e., being able to run PHPUnit (which requires a `composer install`) on newer PHP versions even if e.g., PHP-CS-Fixer does not install on that version yet. (And, in the past, the tools did not support the PHP version range that PHP-CSS-Parser support, i.e., from 5.x to 8.3.)

We can also go full-Composer now and switch back to PHIVE when we gave problems like described with a newer PHP version we want to support.

JakeQZ:
> We can also go full-Composer now and switch back to PHIVE when we gave problems like described with a newer PHP version we want to support.

No, let's stick with PHIVE for the reasons you've given.  Thanks for jogging my memory re the reasons for using it.  All I could remember were problems with updates and caches.  IIRC this was mostly solved by removing the binaries from version control, then it was just an issue of getting the local config right.

PR #440, opened 2024-02-03
[CLEANUP] Autoformat the code

oliverklee:
Autoformat the code with the default PhpStorm settings (and re-run `composer fix:php` to minimize formatting-related changes when code is changed and autoformatted.

### DIFF
--- a/src/OutputFormatter.php
+++ b/src/OutputFormatter.php
@@ -215,6 +215,7 @@ class OutputFormatter
     /**
      *
      * @param array<Commentable> $aComments
+     *
      * @return string
      */
     public function comments(Commentable $oCommentable)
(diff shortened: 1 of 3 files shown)


JakeQZ (approved):
NOC, all good.  Are we PSR-4, or something else?

oliverklee:
> Are we PSR-4, or something else?

For autoloading, we're using PSR-4: https://github.com/sabberworm/PHP-CSS-Parser/blob/master/composer.json#L37

And as coding standard, we're using some mix of PSR-12, PER and a home-made mix. (We might want to move to PER2 at some point to keep up with the PHP ecosystem. I've just added #442 for this.)

PR #441, opened 2024-02-03
[TASK] Add some Composer settings

oliverklee:
- prefer packages over Git
- sort the dependencies

Also actually sort the dependencies.


JakeQZ (approved):
There's a package that will auto-sort this file for you...

JakeQZ (commented):
I can't rebase this.  It needs a rebase.  Otherwise fine.

PR #443, opened 2024-02-04
[TASK] Make the data providers static

oliverklee:
Newer versions of PHPUnit require this, and we might as well do this now.

Closes #414


JakeQZ (approved):
Yes, a necessary change at some point.

Ref:
https://github.com/sebastianbergmann/phpunit/commit/9caafe2d49b33a21f87db248a8ad6ca7c7bdac09
https://stackoverflow.com/questions/4262309/problem-with-phpunit-and-data-providers#4262801

PR #444, opened 2024-02-05
[FEATURE] Add a PHPUnit-specific extension to PHPStan

oliverklee:
This enables PHPStan to understand certain PHPUnit-specific constructs.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,7 +28,9 @@
     },
     "require-dev": {
         "codacy/coverage": "^1.4.3",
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
Curiously, `phpstan/extension-installer` plugin was already specifically allowed in the `config` (though not `require`d).

oliverklee:
That with `phpstan/extension-installer` was me (a bit ahead of the needs of my PRs). ;-)

PR #445, opened 2024-02-05
[BUGFIX] Use the Composer-installed PHPStan for creating the baseline

oliverklee:
Followup to #438


JakeQZ (approved):
Ah, yes, overlooked in #438.

PR #456, opened 2024-02-10
[BUGFIX] Allow at-rules to be parsed in strict mode

### LINKED ISSUE #352: @media block gives syntax error
```
@media (min-width: 1220px) {
  div { width:auto; }
}
```

gives an "Syntax error in CSS: Identifier expected. Got “}” [line no: 3]"

This looks like a regression in version 8.4.0 - I'm suspecting https://github.com/sabberworm/PHP-CSS-Parser/pull/162

JakeQZ:
The reverts the change to `CSSList` in
https://github.com/MyIntervals/PHP-CSS-Parser/commit/134f4e62fe8ab9f316425f3c0f480b3f4f52d804 and adds a comment that `null` is an expected return value when the end of the list (or block) is reached.

Fixes #352

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+Fix (regression) failure to parse at-rules with strict parsing (#456)
+
 ## 8.5.0
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
-Fix (regression) failure to parse at-rules with strict parsing (#456)
+- Fix (regression) failure to parse at-rules with strict parsing (#456)
 
 ## 8.5.0
 
--- a/tests/CSSList/AtRuleBlockListTest.php
+++ b/tests/CSSList/AtRuleBlockListTest.php
@@ -28,7 +28,7 @@ final class AtRuleBlockListTest extends TestCase
     /**
      * @return array<string, array<int, string>>
      */
-    public static function provideSyntacticlyCorrectAtRule()
+    public static function provideSyntacticlyCorrectAtRule(): array
     {
         return [
             'media print' => ['@media print { html { background: white; color: black; } }'],
--- a/tests/CSSList/AtRuleBlockListTest.php
+++ b/tests/CSSList/AtRuleBlockListTest.php
@@ -26,7 +26,7 @@ final class AtRuleBlockListTest extends TestCase
     }
 
     /**
-     * @return array<string, array<int, string>>
+     * @return array<string, array{0: string}>
      */
     public static function provideSyntacticlyCorrectAtRule(): array
     {
--- a/tests/CSSList/AtRuleBlockListTest.php
+++ b/tests/CSSList/AtRuleBlockListTest.php
@@ -115,7 +115,7 @@ final class AtRuleBlockListTest extends TestCase
      * @dataProvider provideMinWidthMediaRule
      * @dataProvider provideSyntacticlyCorrectAtRule
      */
-    public function parsesSyntacticlyCorrectAtRuleInStrictMode($css)
+    public function parsesSyntacticlyCorrectAtRuleInStrictMode(string $css): void
     {
         $contents = (new Parser($css, Settings::create()->beStrict()))->parse()->getContents();
 
--- a/tests/CSSList/AtRuleBlockListTest.php
+++ b/tests/CSSList/AtRuleBlockListTest.php
@@ -15,7 +15,7 @@ use Sabberworm\CSS\Settings;
 final class AtRuleBlockListTest extends TestCase
 {
     /**
-     * @return array<string, array<int, string>>
+     * @return array<string, array{0: string}>
      */
     public static function provideMinWidthMediaRule(): array
     {
--- a/tests/CSSList/AtRuleBlockListTest.php
+++ b/tests/CSSList/AtRuleBlockListTest.php
@@ -7,12 +7,46 @@ use Sabberworm\CSS\Comment\Commentable;
 use Sabberworm\CSS\CSSList\AtRuleBlockList;
 use Sabberworm\CSS\Parser;
 use Sabberworm\CSS\Renderable;
+use Sabberworm\CSS\Settings;
 
 /**
  * @covers \Sabberworm\CSS\CSSList\AtRuleBlockList
  */
 final class AtRuleBlockListTest extends TestCase
 {
+    /**
+     * @return array<string, array<int, string>>
+     */
+    public static function provideMinWidthMediaRule()
+    {
+        return [
+            'without spaces around arguments' => ['@media(min-width: 768px){.class{color:red}}'],
+            'with spaces around arguments' => ['@media (min-width: 768px) {.class{color:red}}'],
+        ];
+    }
+
+    /**
+     * @return array<string, array<int, string>>
+     */
+    public static function provideSyntacticlyCorrectAtRule()
+    {
+        return [
+            'media print' => ['@media print { html { background: white; color: black; } }'],
+            'keyframes' => ['@keyframes mymove { from { top: 0px; } }'],
+            'supports' => ['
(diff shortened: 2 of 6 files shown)


JakeQZ:
> The reverts the change to `CSSList`

Or at least it should.  I found it impossible to copy and paste the previous version of the code from the GitHub diff, so went hunting for a specific version that should have contained the code to restore.  Only it didn't.  So it was a bit painstaking to actually revert that change.  Any tips to speed up the process would be greatly appreciated.

JakeQZ:
`Could not open input file: .\.phive\phpcs.phar`

Can't fix this without knowing what the problems are.  And, of course, the PHIVE set-up is a total PITA.

JakeQZ:
> `Could not open input file: .\.phive\phpcs.phar`
> 
> Can't fix this without knowing what the problems are. And, of course, the PHIVE set-up is a total PITA.

OK, fixed, using info from logs.

tests/ParserTest.php
```diff
@@ -74,7 +74,10 @@ public function files()
                     // or a future test of a as-of-now missing feature
                     continue;
                 }
-                $oParser = new Parser(file_get_contents($sDirectory . '/' . $sFileName));
+                $oParser = new Parser(
+                    \file_get_contents($sDirectory . '/' . $sFileName),
+                    Settings::create()->withLenientParsing($sFileName[0] !== '=')
```
oliverklee, line 79: Instead of having separate sets of fixtures with "magic" prefixes, I'd prefer to communicate this more clearly to the reader, e.g., with a separate testcase that uses the CSS in the test code instead of reading from files. (After all, one of the purposes of tests is to communicate the intended behavior to the reader).

oliverklee (changes requested):

tests/ParserTest.php
JakeQZ: I've moved the new tests to the specific test class responsible.  When working within an IDE, having the test content as separate files is no probem, and helps avoid test class files with literally millions of lines to scroll through, which, even if it doesn't cause the computer to break down, is a bit of a PITA.

JakeQZ:
Can't figure out how to remove the files I added that are no longer wanted.

JakeQZ:
> Can't figure out how to remove the files I added that are no longer wanted.

Non-alphanumeric characters in filenames cause problems in OSs developed in the 1970s, that are still prevelant now.  So must be avoided.

JakeQZ:
I want to convert this from draft to live.  But can't find whatever I need to do to do so.  Nothing hapens when I click the 'Draft' button.

JakeQZ:
> Change the status to “Ready for review” near the bottom of your pull request to remove the draft state

Nope.  Don't see it.

JakeQZ:
> > Change the status to “Ready for review” near the bottom of your pull request to remove the draft state
> 
> Nope. Don't see it.

Finally found it.  What an awful UI.  Thank you Microsoft.

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -7,12 +7,46 @@
...
 
 /**
  * @covers \Sabberworm\CSS\CSSList\AtRuleBlockList
  */
 final class AtRuleBlockListTest extends TestCase
 {
+    /**
+     * @return array<string, array<int, string>>
+     */
+    public static function provideMinWidthMediaRule()
```
oliverklee, line 20:
```suggestion
    public static function provideMinWidthMediaRule(): array
```

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+Fix (regression) failure to parse at-rules with strict parsing (#456)
```
oliverklee, line 22:
```suggestion
- Fix (regression) failure to parse at-rules with strict parsing (#456)
```

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -83,4 +106,19 @@ public function parsesArgumentsOfMediaQueries($css)
...
+
+    /**
+     * @test
+     *
+     * @param string $css
+     *
+     * @dataProvider provideMinWidthMediaRule
+     * @dataProvider provideSyntacticlyCorrectAtRule
+     */
+    public function parsesSyntacticlyCorrectAtRuleInStrictMode($css)
```
oliverklee, line 118:
```suggestion
    public function parsesSyntacticlyCorrectAtRuleInStrictMode(string $css): void
```

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -83,4 +106,19 @@ public function parsesArgumentsOfMediaQueries($css)
 
         self::assertSame('(min-width: 768px)', $atRuleBlockList->atRuleArgs());
     }
+
+    /**
+     * @test
+     *
+     * @param string $css
```
oliverklee, line 113: We can drop the PHPDoc type annotations for cases where we now can use native type declarations instead.

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -7,12 +7,46 @@
...
+        return [
+            'without spaces around arguments' => ['@media(min-width: 768px){.class{color:red}}'],
+            'with spaces around arguments' => ['@media (min-width: 768px) {.class{color:red}}'],
+        ];
+    }
+
+    /**
+     * @return array<string, array<int, string>>
+     */
+    public static function provideSyntacticlyCorrectAtRule()
```
oliverklee, line 31:
```suggestion
    public static function provideSyntacticlyCorrectAtRule(): array
```

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -7,12 +7,46 @@
...
+    public static function provideMinWidthMediaRule()
+    {
+        return [
+            'without spaces around arguments' => ['@media(min-width: 768px){.class{color:red}}'],
+            'with spaces around arguments' => ['@media (min-width: 768px) {.class{color:red}}'],
+        ];
+    }
+
+    /**
+     * @return array<string, array<int, string>>
```
oliverklee, line 29:
```suggestion
     * @return array<string, array{0: string}>
```

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -7,12 +7,46 @@
...
 use Sabberworm\CSS\Renderable;
+use Sabberworm\CSS\Settings;
 
 /**
  * @covers \Sabberworm\CSS\CSSList\AtRuleBlockList
  */
 final class AtRuleBlockListTest extends TestCase
 {
+    /**
+     * @return array<string, array<int, string>>
```
oliverklee, line 18:
```suggestion
     * @return array<string, array{0: string}>
```

oliverklee (changes requested):

sabberworm (approved):

oliverklee (approved):
LGTM. :pray:

JakeQZ:
Wondering it we should release an 8.5.1 with this fix, since it was a regression in 8.4.

oliverklee:
Yeah, I think backporting it to 8.5.x makes sense. We probably need to change the native type declarations to PHPDoc annotations.

Would you be willing to create the backport PR?

And should we release 8.5.1 directly after merging it, or should we first collect some more fixes?

JakeQZ:
> Yeah, I think backporting it to 8.5.x makes sense. We probably need to change the native type declarations to PHPDoc annotations.

Ah, yes, the 8.x branch supports older PHP versions which don't support native type declerations.
 
> Would you be willing to create the backport PR?

I should hopefully be able to merge the commit to main to a PR branch off the 8.x.  Then revert the type hint changes.

> And should we release 8.5.1 directly after merging it, or should we first collect some more fixes?

The other outstanding PRs awaiting review mostly don't look straightforward, and may require post-review changes, so may take some time to complete.

I was keen to resolve this issue as a priority since it was a regression.  The others aren't, AFAIAA, so it would be on a case-by-case basis whether they can be easily backported once completed on `main`.  So I'd suggest an 8.5.1 with this fix, and some time later when the outstanding PRs have been resolved one way or another, (probably) an 8.5.2 with a rollup of fixes from those.

PR #468, opened 2024-02-12
[TASK] Drop PHP_CodeSniffer

oliverklee:
This tool is not well-maintained anymore, and it has no PER2 coding standard, which is a blocker for us moving to PER2.

PHP-CS-Fixer already has a PER2 coding standard. If we use PHP-CS-Fixer on PER2 and PHP_CodeSniffer on PSR12 (the latest available standard there), both tools would keep overwriting each other's changes endlessly, and they would never agree on the correct formatting.


oliverklee:
Marking as draft - I’ll still need to remove the corresponding GPG key fingerprint from the CI job.

sabberworm (approved):
I agree we shouldn’t use two tools for the same job.

I just wanted to point out that the project, much like this one, has recently moved to a new org (see squizlabs/PHP_CodeSniffer#3932) and there seems to be lots of activity as well as an open issue for PER-2 (PHPCSStandards/PHP_CodeSniffer#29).

PR #472, opened 2024-02-13
[TASK] Mark the public class constants as `@internal`

### LINKED ISSUE #471: Mark the public class constants as `@internal`
Those class constants are not intended to be accessed from outside this package.

oliverklee:
Fixes #471


JakeQZ (approved):
They are used by other classes within the project, but not intended for third-party use, as they are subject to change or removal.

PR #474, opened 2024-02-13
[TASK] Avoid generating a `composer.lock`

### LINKED ISSUE #469: Avoid generating a `composer.lock`
Like so: https://github.com/FriendsOfTYPO3/tea/pull/1161

oliverklee:
This is a library, and libraries should not have a lock file.

Fixes #469

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -50,6 +50,7 @@
         "allow-plugins": {
             "phpstan/extension-installer": true
         },
+        "lock": false,
         "preferred-install": {
             "*": "dist"
         },


JakeQZ (changes requested):
This causes issues when working on the project directly as a developer.  It prevents Composer knowing which versions of dependencies are currently installed.  So when `composer update` is run, it will reinstall everything.

Obviously `composer.lock` should not be added to version control, or included in the released package, but I don't see the need for this settings change to enforce that.

Also, the linked PR seems to be to resolve a specific problem with GitLab, which would have been better solved by adding `lock=false` to the command line, so as not to apply an unwanted change into other development environments.

oliverklee:
Okay, closing then.

PR #476, opened 2024-02-14
[TASK] Stop collecting code coverage on the maintenance branch

oliverklee:
We're not using the code coverage of maintenance branches for anything and could as well speed up the CI build instead.


JakeQZ (approved):
Looks thorough and complete AFAICT.

PR #484, opened 2024-02-17
[BUGFIX] Invoke PHIVE-installed tools via shell

### LINKED ISSUE #479: Composer scripts invoking PHIVE Phars don't run on Windows
With `copy=false` in `phars.xml`, the Phars are installed globally in the user account, with a batch file (`*.bat`) placed in the `.phive` directory (named e.g. `php-cs-fixer.bat` not `php-cs-fixer.phar`).  The script currently invokes the PHP executable with the Phar filename explicitly with the `*.phar` extension.  Even if the extension was removed, PHP cannot open Windows batch files.

There are two possible solutions:
- Change to `copy=true` in `phars.xml`
  - This potentially occupies more disk space, if they could be shared between projects (or copies of a single project);
- Remove `php` and the file extension from the script command (e.g. so that `@php ./.phive/php-cs-fixer.phar` becomes `\"./.phive/php-cs-fixer\"`, with the backslashes escaping the quotes for Composer
  - This is what we did in Emogrifier;
  - If the OS won't implicitly execute a file with an extension given a name without it, that could be an issue, but perhaps PHIVE creates an extensionless file to chain on to the Phar to cover that base.

JakeQZ:
On Windows, the `.phive` directory contains batch files that in turn invoke the phar installed elsewhere on the system.  To be compatible with both Linux and Windows, the file must be invoked directly via the shell, and without the file extension specified.

Fixes #479.

### DIFF
--- a/.phive/phars.xml
+++ b/.phive/phars.xml
@@ -1,4 +1,4 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <phive xmlns="https://phar.io/phive">
-  <phar name="php-cs-fixer" version="^3.49.0" installed="3.49.0" location="./.phive/php-cs-fixer.phar" copy="false"/>
+  <phar name="php-cs-fixer" version="^3.49.0" installed="3.49.0" location="./.phive/php-cs-fixer" copy="false"/>
 </phive>
--- a/composer.json
+++ b/composer.json
@@ -64,7 +64,7 @@
         "ci": [
             "@ci:static"
         ],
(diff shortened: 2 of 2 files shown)


JakeQZ:
Hmm, this doesn't seem to be working on Linux now.  Though it seems to be the same as what works in the Emogrifier project.

JakeQZ:
> Hmm, this doesn't seem to be working on Linux now. Though it seems to be the same as what works in the Emogrifier project.

Ah, seems the phars also need to be installed without a file extension, to be compatible with both OSs.

oliverklee (approved):

PR #485, opened 2024-02-19
[BUGFIX] Handle incorrect RGB colors better

ali-khalili:
There is an issue when handling incorrect RGB colors which have lesser character numbers than 6. In this situation, we should probably either throw an exception OR define a fallback value for the color. 
In this PR, I have updated the code to use RGB(0,0,0) as the fallback color when this situation happens.
Thank you

### DIFF
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -63,12 +63,12 @@ class Color extends CSSFunction
                     'b' => new Size(intval($sValue[4] . $sValue[5], 16), null, true, $oParserState->currentLine()),
                 ];
             } else {
-                // the size of input color string is incorrect, use black color as fallback
-                $aColor = [
-                    'r' => new Size(0, null, true, $oParserState->currentLine()),
-                    'g' => new Size(0, null, true, $oParserState->currentLine()),
-                    'b' => new Size(0, null, true, $oParserState->currentLine()),
-                ];
+                throw new UnexpectedTokenException(
+                    'Invalid hex color value', 
+                    $sValue, 
+                    'custom', 
+                    $oParserState->currentLine()
+                );
             }
         } else {
             $sColorMode = $oParserState->parseIdentifier(true);
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -56,7 +56,7 @@ class Color extends CSSFunction
                         $oParserState->currentLine()
                     ),
                 ];
-            } else if ($oParserState->strlen($sValue) === 6) {
+            } elseif ($oParserState->strlen($sValue) === 6) {
                 $aColor = [
                     'r' => new Size(intval($sValue[0] . $sValue[1], 16), null, true, $oParserState->currentLine()),
                     'g' => new Size(intval($sValue[2] . $sValue[3], 16), null, true, $oParserState->currentLine()),
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -56,11 +56,18 @@ class Color extends CSSFunction
                         $oParserState->currentLine()
                     ),
                 ];
-            } else {
+            } else if ($oParserState->strlen($sValue) === 6) {
                 $aColor = [
                     'r' => new Size(intval($sValue[0] . $sValue[1], 16), null, true, $oParserState->currentLine()),
                     'g' => new Size(intval($sValue[2] . $sValue[3], 16), null, true, $oParserState->currentLine()),
                     'b' => new Size(intval($sValue[4] . $sValue[5], 16), null, true, $oParserState->currentLine()),
+                ];
+            } else {
+                // the size of input color string is incorrect, use black color as fallback
+                $aColor = [
+                    'r' => new Size(0, null, true, $oParserState->currentLine()),
+                    'g' => new Size(0, null, true, $oParserState->currentLine()),
+                    'b' => new Size(0, null, true, $oParserState->currentLine()),
                 ];
             }
         } else {
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -64,9 +64,9 @@ class Color extends CSSFunction
                 ];
             } else {
                 throw new UnexpectedTokenException(
-                    'Invalid hex color value', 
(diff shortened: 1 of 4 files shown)


src/Value/Color.php
```diff
@@ -56,12 +56,19 @@ public static function parse(ParserState $oParserState, $bIgnoreCase = false)
...
                     'g' => new Size(intval($sValue[2] . $sValue[3], 16), null, true, $oParserState->currentLine()),
                     'b' => new Size(intval($sValue[4] . $sValue[5], 16), null, true, $oParserState->currentLine()),
                 ];
+            } else {
+                // the size of input color string is incorrect, use black color as fallback
+                $aColor = [
+                    'r' => new Size(0, null, true, $oParserState->currentLine()),
+                    'g' => new Size(0, null, true, $oParserState->currentLine()),
+                    'b' => new Size(0, null, true, $oParserState->currentLine()),
+                ];
```
JakeQZ, line 71: If the value is not valid, we should throw an `UnexpectedTokenException`.  This wil be caught by `RuleSet::parseRuleSet()`, resulting in the invalid property declaration being dropped (while other valid properties are retained, and which matches the behaviour of most browsers).

```suggestion
                throw new UnexpectedTokenException(
                    'Invalid hex color value', 
                    $sValue, 
                    'custom', 
                    $oParserState->currentLine()
                );
```

tests/ParserTest.php
```diff
@@ -146,6 +146,13 @@ public function colorParsing()
                     'l' => new Size(220.0, '%', true, $oColor->getLineNo()),
                     'a' => new Size(0000.3, null, true, $oColor->getLineNo()),
                 ], $oColor->getColor());
+                $aColorRule = $oRuleSet->getRules('outline-color');
+                $oColor = $aColorRule[0]->getValue();
+                self::assertEquals([
+                    'r' => new Size(0, null, true, $oColor->getLineNo()),
+                    'g' => new Size(0, null, true, $oColor->getLineNo()),
+                    'b' => new Size(0, null, true, $oColor->getLineNo()),
+                ], $oColor->getColor());
```
JakeQZ, line 155: With the change to throw an exception, this would now instead need to assert that there's no `outline-color` rule (i.e. the invalid declaration has been dropped).

tests/ParserTest.php
```diff
@@ -155,7 +162,7 @@ public function colorParsing()
             '#mine {color: red;border-color: #0a64e6;border-color: rgba(10,100,231,.3);outline-color: #222;'
             . 'background-color: #232323;}'
             . "\n"
-            . '#yours {background-color: hsl(220,10%,220%);background-color: hsla(220,10%,220%,.3);}'
+            . '#yours {background-color: hsl(220,10%,220%);background-color: hsla(220,10%,220%,.3);outline-color: #000;}'
```
JakeQZ, line 165: With the change to throw an exception, the result of `render()` should now be as before.

JakeQZ (changes requested):
Hi @ali-khalili,

Thanks for this.  Good catch.

I think when an invalid property value is encountered, an `UnexpectedTokenException` should be thrown.  This will be caught upstream so that the invalid property is dropped (what most browsers do), rather than resulting in a property with a now-valid value (causing a change of behaviour if the rendered CSS is used in place of the input CSS).

Also, could you add an entry to `CHANGELOG.md` (newest first within the 'Fixed' section)?

Thanks <3

@oliverklee, @sabberworm, do you spot any other issue?

src/Value/Color.php
oliverklee: I agree.

oliverklee (changes requested):
Yes, please let's discard invalid rules instead of falling back to some default.

tests/ParserTest.php
sabberworm: If the declaration was already dropped by the previous implementation, then how was it incorrect, exactly?

JakeQZ: > If the declaration was already dropped by the previous implementation, then how was it incorrect, exactly?

This change adds an invalid declaration `outline-color: #22;` to the test fixture, which wasn't previously present, but which should be dropped.  Prior to the change, a PHP notice "Uninitialized string offset" would occur, but some semblance of the declaration would still be included (I think as if it were `#220000`, if `intval()` on the result of an out-of-bounds string access returns zero).

src/Value/Color.php
ali-khalili: Thanks, yes this is a much better solution.

JakeQZ (approved):
Looks good to me now.  Thanks for contributing 👍

oliverklee (approved):
Thanks! :heart:

JakeQZ:
It didn't go unnoticed that you had to wait for the CI checks to be triggered by a maintainer.  Thanks for persevering.  I hope we can imporove the process for future contributions/contributors (#486).

oliverklee:
@JakeQZ @sabberworm Is this something that would make sense to backport to 8.x?

JakeQZ:
> Is this something that would make sense to backport to 8.x?

I think so:
- it's a bug fix;
- it's quite a small change so should be fairly straightforward to merge.

oliverklee:
Done: #490

JakeQZ (commented):
> I sync'ed the branch with the GitHub UI (hence the merge commit) I can rebase it properly later if needed.

Thanks.  The GitHub UI sync seems to have been sufficient.  (I still don't know why I couldn't get the fork to rebase.)

I've now been able to confirm that these new tests would fail without the fix in #485, and pass with the fix in place.

I like the completeness of the test file, containing invalid color values with 1, 2, 4 and 5 characters.

Could you rename the test methods (to avoid the second one simply having the number `2` appeneded to the name)?  And resolve the PHP-CS-Fixer error?

Thanks again.

JakeQZ:
> > I sync'ed the branch with the GitHub UI (hence the merge commit) I can rebase it properly later if needed.

Sorry, posted on wrong thread.  This was meant for #359.

PR #503, opened 2024-02-26
[FEATURE] Add Composer scripts to run unit tests 

JakeQZ:
This is the change to `composer.json` which was inadvertently omitted in #488.

### DIFF
--- a/README.md
+++ b/README.md
@@ -633,5 +633,17 @@ class Sabberworm\CSS\CSSList\Document#4 (2) {
 
 ## Misc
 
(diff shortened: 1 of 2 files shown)


JakeQZ:
> This is the change to `composer.json` which was inadvertently omitted in #488.

It also includes the same change to `README.md` that was committed in that PR.  I should have added two files to the commit, but managed to only add one of them.  Doesn't seem to need a rebase as the change to the file that was committed is identical.

oliverklee (approved):

PR #505, opened 2024-02-27
[TASK] Reorder things in the class diagram source code

oliverklee:
Move the associations to the bottom of the code. Also sort them.

This helps us use an autogenerated diagram to which we add the associations (as the our tool of choice does not generate the associations for the diagram).


JakeQZ (approved):
Confirmed with WinMerge this is just a reordering.

PR #506, opened 2024-02-27
[CLEANUP] Add an interface for components of a property value

### LINKED ISSUE #499: Provide an interface for `RuleValueList|CSSFunction|CSSString|LineName|Size|URL`
This will help us reduce some overlong type annotations in `Color`, `CSSFunction`, `LineName`, `Value` and `ValueList`.

JakeQZ:
Resolves #499.

### DIFF
--- a/src/Value/Component.php
+++ b/src/Value/Component.php
@@ -1,0 +1,11 @@
+<?php
+
+namespace Sabberworm\CSS\Value;
+
+/**
+ * To be used by any class that represents a component of a CSS property value.
+ * Values often comprise 'component operator component' (recursively).
+ * This interface has no methods to implement;
+ * its purpose is abstract: to allow a unique type to be specified when a `Component` is an argument of a method.
+ */
+interface Component {}
--- a/src/Value/CSSFunction.php
+++ b/src/Value/CSSFunction.php
@@ -9,7 +9,7 @@ use Sabberworm\CSS\Parsing\ParserState;
  * A `CSSFunction` represents a special kind of value that also contains a function name and where the values are the
  * function’s arguments. It also handles equals-sign-separated argument lists like `filter: alpha(opacity=90);`.
  */
-class CSSFunction extends ValueList
+class CSSFunction extends ValueList implements Component
 {
     /**
      * @var string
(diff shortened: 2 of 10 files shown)


oliverklee (changes requested):
Looks very good to me in general.

I think this deserves an entry in the changelog.

Also, we should update the class diagram to integrate the interface and to update the associations to use the interface (maybe after #505 is merged).

src/Value/Component.php
```diff
@@ -0,0 +1,11 @@
...
+
+namespace Sabberworm\CSS\Value;
+
+/**
+ * To be used by any class that represents a component of a CSS property value.
+ * Values often comprise 'component operator component' (recursively).
+ * This interface has no methods to implement;
+ * its purpose is abstract: to allow a unique type to be specified when a `Component` is an argument of a method.
+ */
+interface Component {}
```
oliverklee, line 11: We also should make sure that the name makes sense globally (as Mermaid on GitHub doesn't allow namespaces). Maybe `CssPropertyValueComponent` oder `PropertyValueComponent`?

oliverklee, line 11: Also, we should check if there are any common methods that should be part of the interface so the interface is more than just a marker.

JakeQZ: > We also should make sure that the name makes sense globally (as Mermaid on GitHub doesn't allow namespaces). Maybe `CssPropertyValueComponent` oder `PropertyValueComponent`?

I don't think whatever limitations of Mermaid there may be should dictate class naming conventions.  Mermaid might in future be enhanced to show the namespaces.  We should choose the most apt names for classes, interfaces and namespaces independently of whatever other tools we use.

The FQN is `Value\Component`.  The namespace `Value` could perhaps be `PropertyValue` (or `Property\Value`) for clarity, though any such rename would be beyond the scope of this PR, and probably unnecessary (it would be a breaking change).

JakeQZ: > Also, we should check if there are any common methods that should be part of the interface so the interface is more than just a marker.

Looking again, I don't see any common methods, but do see that all of these classes extend `Value`, which implements `Renderable`.  A `Value` can be a simple value, or a complex value comprised of simple values connected with operators.

So perhaps a better approach would be to simply use the `Value` type whenever any of the various subclasses is expected, and not introduce a new type (or interface) at all.

JakeQZ:
Looking again, I think instead we should just use the existing ancestor `Value` type to replace the long list of possible subtypes in the annotations, rather than adding an new empty interface.

> Also, we should update the class diagram to integrate the interface and to update the associations to use the interface (maybe after #505 is merged).

Yes.  (I need to get up to speed on that.)

oliverklee:
> I think instead we should just use the existing ancestor `Value` type to replace the long list of possible subtypes in the annotations, rather than adding a new empty interface.

Sounds good to me.

PR #513, opened 2024-03-01
[CLEANUP] Move some things in the class diagram code around

oliverklee:
This gets another thing out of the way towards moving to an autogenerated class diagram.

Also sort the moved lines.

### DIFF
--- a/README.md
+++ b/README.md
@@ -651,8 +651,6 @@ classDiagram
         <<interface>>
     }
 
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
Confirmed that this just reorders things in the source.  Also looks tidier now, with all classes (and interfaces) presented first, then relationships.

PR #515, opened 2024-03-03
[TASK] Use a generated class/diagram list in the class diagram

oliverklee:
This is another step towards auto-generating more of the class diagram.

### DIFF
--- a/README.md
+++ b/README.md
@@ -621,132 +621,100 @@ class Sabberworm\CSS\CSSList\Document#4 (2) {
 classDiagram
     direction LR
 
-    %% top namespace
-
-    class OutputFormat {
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
The `abstruct` typo is also in the tool: https://github.com/tasuku43/php-mermaid-class-diagram/blob/v0.1.0/src/ClassDiagramRenderer/Node/AbstractClass_.php#L12

Would be good to get that fixed, but beyond the scope of this PR.

oliverklee:
I've opened a PR to fix the "abstruct": https://github.com/tasuku43/php-mermaid-class-diagram/pull/4

Issue #466, opened 2024-02-12
Drop the class documentation and the generation infrastructure


oliverklee:
https://myintervals.github.io/PHP-CSS-Parser/annotated.html

Also drop the doxygen configuration.

oliverklee:
Closing in favor of #522.

PR #516, opened 2024-03-03
[TASK] Drop the configuration for generating the API documentation

oliverklee:
Now that IDEs support navigating and displaying PHPDoc comments reasonibly well, and now that we have a UML class diagram, rendering the PHPDoc does not provide relevant benefits anymore.

Part of #466


JakeQZ (commented):
The title of this PR seems to be wrong.  Isn't it the configuration for generating HTML documentation (with Doxygen) that's being removed?

I don't know whether or how it's actively being used.  However, having navigable HTML could still be useful in some situations:
- if the library was compiled into a PHAR, the PHPDoc wouldn't be available;
- when considering using the library, without having yet installed it, being able to browse the HTML documention could be helpful.

So I am not sure about removing this.  Also, does keeping it cause any problem?  I think it is should not exist in the package installed as a dependency via Composer (because it would just waste disk space there) - is that already taken care of?

oliverklee:
Moving the discussion to #520.

oliverklee:
Closing in favor of #522.

PR #518, opened 2024-03-05
[TASK] Use auto-generated hierarchy in class diagram

oliverklee:
Now the only part of the class diagram which we need to maintain manually are the associations between classes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -713,46 +713,47 @@ classDiagram
     class Comment {
     }
 
+    RuleSet <|-- DeclarationBlock: inheritance
+    Renderable <|.. RuleSet: realization
+    Commentable <|.. RuleSet: realization
+    RuleSet <|-- AtRuleSet: inheritance
+    AtRule <|.. AtRuleSet: realization
+    Selector <|-- KeyframeSelector: inheritance
+    Renderable <|-- AtRule: inheritance
+    Commentable <|-- AtRule: inheritance
+    AtRule <|.. Charset: realization
+    AtRule <|.. Import: realization
+    AtRule <|.. CSSNamespace: realization
+    Renderable <|.. Rule: realization
+    Commentable <|.. Rule: realization
+    SourceException <|-- OutputException: inheritance
+    UnexpectedTokenException <|-- UnexpectedEOFException: inheritance
+    Exception <|-- SourceException: inheritance
+    SourceException <|-- UnexpectedTokenException: inheritance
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -713,46 +713,47 @@ classDiagram
...
+    Renderable <|-- AtRule: inheritance
+    Commentable <|-- AtRule: inheritance
+    AtRule <|.. Charset: realization
+    AtRule <|.. Import: realization
+    AtRule <|.. CSSNamespace: realization
+    Renderable <|.. Rule: realization
+    Commentable <|.. Rule: realization
+    SourceException <|-- OutputException: inheritance
+    UnexpectedTokenException <|-- UnexpectedEOFException: inheritance
+    Exception <|-- SourceException: inheritance
```
JakeQZ, line 731: Should inheritance from in-built classes be included?

JakeQZ (changes requested):
Could you add a Composer script to run the tool, with the arguments you are using, including the tool as a `dev` dependency?  Currently, the rest of us are somewhat in the dark as to how you are generating this automatically.

I'm guessing that at present you don't have a solution for automatically updating the readme between the "start of" and "end of" comments, though that could ultimately be achieved.  Assuming not, something that outputs an `.md` (or whaterver) file to be inserted would suffice for now.  If it outputs a fixed file (rather than to `stdout`), then that filename would probably need to be added to `.gitignore`.

This could be done as a separate PR first, using the command line that created the current content before this proposed change.  Then updated in this PR to also generate the inheritance rules that are being moved to auto-generated with this PR.

Changes themselves seem fine, though I have a query about whether inheritence rules from in-built classes should be included.  They weren't before (specifically: `SourceException extends \Exception`).

README.md
oliverklee: As far as I know, there is no option to configure this behavior. So if we want to automate this, I propose we stick to what the tool generates.

oliverklee:
I have create #521 for the Composer script and would suggest moving this change here first to get it out of the way.

JakeQZ (approved):
> moving this change here first to get it out of the way.

Presume you mean merging this PR first?  I'm happy with that.

I note the latest push is just a reordering and removal of the relationship with the built-in class.

oliverklee:
> I note the latest push is just a reordering and removal of the relationship with the built-in class.

Yes - I only rebased on top of the current `main`.

PR #545, opened 2024-06-17
[FEATURE] Add support for inserting an item in a CSSList

ziegenberg:
This is based on and will supersede #115. Suggestions from the original PR are incorporated, and the initial authorship is preserved.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -299,9 +299,7 @@ abstract class CSSList implements Renderable, Commentable
      *
      * @param RuleSet|CSSList|Import|Charset $oItem The item.
      * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
-     * @param string $sPosition The position.
-     *
-     * @return void
+     * @param $sPosition The position.
      */
     public function insert($oItem, $oSibling, string $sPosition = 'before'): void
     {
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -295,21 +295,18 @@ abstract class CSSList implements Renderable, Commentable
     }
 
     /**
-     * Insert an item before or after its sibling.
+     * Insert an item before its sibling.
      *
      * @param RuleSet|CSSList|Import|Charset $oItem The item.
      * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
-     * @param $sPosition The position.
      */
-    public function insert($oItem, $oSibling, string $sPosition = 'before'): void
+    public function insertBefore($oItem, $oSibling): void
     {
         $iIndex = in_array($oSibling, $this->aContents);
         if ($iIndex === false) {
             $this->append($oItem);
-        } elseif ($sPosition === 'before') {
-            $this->replace($oSibling, array($oItem, $oSibling));
         } else {
-            $this->replace($oSibling, array($oSibling, $oItem));
+            $this->replace($oSibling, array($oItem, $oSibling));
         }
     }
 
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -298,15 +298,15 @@ abstract class CSSList implements Renderable, Commentable
      * Inserts an item before its sibling. If the desired sibling cannot be found,
      * the item is appended at the end.
      *
-     * @param RuleSet|CSSList|Import|Charset $oItem
-     * @param RuleSet|CSSList|Import|Charset $oSibling
+     * @param RuleSet|CSSList|Import|Charset $item
+     * @param RuleSet|CSSList|Import|Charset $sibling
      */
-    public function insertBefore($oItem, $oSibling): void
+    public function insertBefore($item, $sibling): void
     {
-        if (in_array($oSibling, $this->aContents, true)) {
-            $this->replace($oSibling, [$oItem, $oSibling]);
+        if (in_array($sibling, $this->aContents, true)) {
+            $this->replace($oSibling, [$item, $sibling]);
         } else {
-            $this->append($oItem);
+            $this->append($item);
         }
     }
 
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -304,7 +304,7 @@ abstract class CSSList implements Renderable, Commentable
     public function insertBefore($item, $sibling): void
     {
         if (in_array($sibling, $this->aContents, true)) {
-            $this->replace($oSibling, [$item, $sibling]);
+            $this->replace($sibling, [$item, $sibling]);
         } else {
             $this->append($item);
         }
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -295,17 +295,18 @@ abstract class CSSList implements Renderable, Commentable
     }
 
     /**
-     * Inserts an item before its sibling.
+     * Inserts an item before its sibling. If the desired sibling cannot be found,
+     * the item is appended at the end.
      *
      * @param RuleSet|CSSList|Import|Charset $oItem
      * @param RuleSet|CSSList|Import|Charset $oSibling
      */
     public function insertBefore($oItem, $oSibling): void
     {
-        if (in_array($oSibling, $this->aContents, true) === false) {
-            $this->append($oItem);
-        } else {
+        if (in_array($oSibling, $this->aContents, true)) {
             $this->replace($oSibling, [$oItem, $oSibling]);
+        } else {
+            $this->append($oItem);
         }
     }
 
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -295,18 +295,17 @@ abstract class CSSList implements Renderable, Commentable
     }
 
     /**
-     * Insert an item before its sibling.
+     * Inserts an item before its sibling.
      *
-     * @param RuleSet|CSSList|Import|Charset $oItem The item.
-     * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
+     * @param RuleSet|CSSList|Import|Charset $oItem
+     * @param RuleSet|CSSList|Import|Charset $oSibling
      */
     public function insertBefore($oItem, $oSibling): void
     {
-        $iIndex = in_array($oSibling, $this->aContents);
-        if ($iIndex === false) {
+        if (in_array($oSibling, $this->aContents, true) === false) {
             $this->append($oItem);
         } else {
-            $this->replace($oSibling, array($oItem, $oSibling));
+            $this->replace($oSibling, [$oItem, $oSibling]);
         }
     }
 
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -295,6 +295,27 @@ abstract class CSSList implements Renderable, Commentable
     }
 
     /**
+     * Insert an item before or after its sibling.
+     *
+     * @param RuleSet|CSSList|Import|Charset $oItem The item.
+     * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
+     * @param string $sPosition The position.
+     *
+     * @return void
+     */
+    public function insert($oItem, $oSibling, string $sPosition = 'before'): void
+    {
+        $iIndex = in_array($oSibling, $this->aContents);
+        if ($iIndex === false) {
+            $this->append($oItem);
+        } elseif ($sPosition === 'before') {
+            $this->replace($oSibling, array($oItem, $oSibling));
+        } else {
+            $this->replace($oSibling, array($oSibling, $oItem));
+        }
+    }
+
+    /**
      * Removes an item from the CSS list.
      *
      * @param RuleSet|Import|Charset|CSSList $oItemToRemove
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -295,7 +295,7 @@ abstract class CSSList implements Renderable, Commentable
     }
 
     /**
-     * Inserts an item before its sibling. If the desired sibling cannot be found,
+     * Inserts an item in the CSS list before its sibling. If the desired sibling cannot be found,
      * the item is appended at the end.
      *
      * @param RuleSet|CSSList|Import|Charset $item
--- a/tests/CSSList/DocumentTest.php
+++ b/tests/CSSList/DocumentTest.php
@@ -87,9 +87,9 @@ final class DocumentTest extends TestCase
     }
 
     /**
-     * @return Generator
+     * @return \Generator<array<array<DeclarationBlock>, DeclarationBlock, DeclarationBlock, array<DeclarationBlock>>>
      */
-    public static function insertDataProvider(): \Generator
+    public static function insertBeforeDataProvider(): \Generator
     {
 
         $bogusOne = new DeclarationBlock();
@@ -110,21 +110,12 @@ final class DocumentTest extends TestCase
             'initialContent' => [$bogusOne, $oSibling, $bogusTwo],
             'oItem' => $oItem,
             'oSibling' => $oSibling,
-            'position' => 'before',
             'expectedContent' => [$bogusOne, $oItem, $oSibling, $bogusTwo],
-            ];
-        yield 'insert after' => [
-            'initialContent' => [$bogusOne, $oSibling, $bogusTwo],
-            'oItem' => $oItem,
-            'oSibling' => $oSibling,
-            'position' => 'after',
-            'expectedContent' => [$bogusOne, $oSibling, $oItem, $bogusTwo],
             ];
         yield 'append if not found' => [
             'initialContent' => [$bogusOne, $oSibling, $bogusTwo],
             'oItem' => $oItem,
             'oSibling' => $oOrphan,
-            'position' => 'before',
             'expectedContent' => [$bogusOne, $oSibling, $bogusTwo, $oItem],
             ];
     }
@@ -138,13 +129,12 @@ final class DocumentTest extends TestCase
      * @param $sPosition
      * @param $expectedContent
      *
-     * @dataProvider insertDataProvider
+     * @dataProvider insertBeforeDataProvider
      */
-    public function insertContent(
+    public function insertContentBefore(
         array $initialContent,
         DeclarationBlock $oItem,
         DeclarationBlock $oSibling,
-        string $sPosition,
         array $expectedContent
     ) {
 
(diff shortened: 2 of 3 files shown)


ziegenberg:
I fixed the Static Analysis errors. Do I need to fix the unit test? They are working with PHP 8.

oliverklee:
> I fixed the Static Analysis errors.

Thanks! :pray: 

> Do I need to fix the unit test? They are working with PHP 8.

Yes, please. AFAICT, the `mixed` type annotation is the culprit, as it's only available in PHP 8 and up, not in PHP 7.x: https://www.php.net/manual/en/language.types.mixed.php

ziegenberg:
Fixed it.

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,27 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
+    /**
+     * Insert an item before or after its sibling.
+     *
+     * @param RuleSet|CSSList|Import|Charset $oItem The item.
+     * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
+     * @param string $sPosition The position.
+     *
+     * @return void
+     */
+    public function insert(RuleSet|CSSList|Import|Charset $oItem, RuleSet|CSSList|Import|Charset $oSibling, string $sPosition = 'before'): void
```
oliverklee: As we still support PHP 7.2, we cannot have native union types. So we'll need to omit the native type declarations for those and have them in PHPDoc annotations only instead.

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,27 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
         array_splice($this->aContents, $iOffset, $iLength, $mReplacement);
     }
 
+    /**
+     * Insert an item before or after its sibling.
+     *
+     * @param RuleSet|CSSList|Import|Charset $oItem The item.
+     * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
+     * @param string $sPosition The position.
```
oliverklee, line 304: For the cases where we already have a (working) native type declaration and where the PHPDoc comment does not add anything substantial over the native annotation, please drop the now-redundant PHPDoc annotation.

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,27 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
+     *
+     * @return void
```
oliverklee, line 306: For the cases where we already have a (working) native type declaration and where the PHPDoc comment does not add anything substantial over the native annotation, please drop the now-redundant PHPDoc annotation.

tests/CSSList/DocumentTest.php
```diff
@@ -2,6 +2,7 @@
 
 namespace Sabberworm\CSS\Tests\CSSList;
 
+use Generator;
```
oliverklee, line 5: Please don't `use` native top-level PHP classes/functions. Instead, please use them like this where they're used:  `\Generator`

tests/CSSList/DocumentTest.php
```diff
@@ -85,4 +86,72 @@ public function setContentsReplacesContentsSetInPreviousCall()
 
         self::assertSame($contents2, $this->subject->getContents());
     }
+
+    /**
+     * @return Generator
```
oliverklee, line 91: Let's be as specific as possible:

```suggestion
     * @return Generator<initialContent: list<…>, …>
```

tests/CSSList/DocumentTest.php
```diff
@@ -85,4 +86,72 @@ public function setContentsReplacesContentsSetInPreviousCall()
...
+            'oSibling' => $oOrphan,
+            'position' => 'before',
+            'expectedContent' => [$bogusOne, $oSibling, $bogusTwo, $oItem],
+            ];
+    }
+
+    /**
+     * @test
+     *
+     * @param array $contents
```
oliverklee, line 136: For the cases where we already have a (working) native type declaration and where the PHPDoc comment does not add anything substantial over the native annotation, please drop the now-redundant PHPDoc annotation.

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,27 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
+     */
+    public function insert($oItem, $oSibling, string $sPosition = 'before'): void
```
oliverklee, line 308: A comment concerning architecture:

I'd prefer to have separate methods with distinct, specific responsibilities and communicative names instead of a single method that combines multiple behaviors. The `$sPosition` behavior-switching parameter is a tell-tale sign that we should split up the method.

I suggest adopting the naming of the corresponding methods from the native `DOMElement` class: https://www.php.net/manual/en/class.domelement.php

oliverklee (changes requested):
Thanks a bunch for this contribution! :heart:

I have added a few comments - one concerning the architecture, and a few concerning minor style nits.

src/CSSList/CSSList.php
ziegenberg: So the original comment by @raxbg in https://github.com/MyIntervals/PHP-CSS-Parser/issues/115#issuecomment-553319653 is no longer valid?

Would a single `insertBefore()` be ok? _(See: https://www.php.net/manual/en/domnode.insertbefore.php)_

src/CSSList/CSSList.php
ziegenberg: Yeah, figured that out. :/ It's fixed now :)

src/CSSList/CSSList.php
oliverklee: > So the original comment by @raxbg in https://github.com/MyIntervals/PHP-CSS-Parser/pull/115#issuecomment-553319653 is no longer valid?

Correct. We're a different bunch of people now with different opinions/preferences. :-)

oliverklee: > Would a single `insertBefore()` be ok?

If that's sufficient for your usecase, yes, of course.

tests/CSSList/DocumentTest.php
```diff
@@ -85,4 +85,66 @@ public function setContentsReplacesContentsSetInPreviousCall()
...
+    }
+
+    /**
+     * @test
+     *
+     * @param $initialContent
+     * @param $oItem
+     * @param $oSibling
+     * @param $sPosition
+     * @param $expectedContent
```
oliverklee, line 130: We can drop all those PHPDoc annotations as we already have the native type declarations.

tests/CSSList/DocumentTest.php
```diff
@@ -85,4 +85,66 @@ public function setContentsReplacesContentsSetInPreviousCall()
...
+     *
+     * @dataProvider insertBeforeDataProvider
+     */
+    public function insertContentBefore(
```
oliverklee, line 134: The tested method has two kinds of behavior: replacing and inserting. Please let's have two different tests for this.

The test name should be a statement communicating the tested behavior. This allows the test name and test code to form a kind of "double bookkeeping". Please have a look at this slide about naming tests: https://speakerdeck.com/oliverklee/test-driven-development-with-phpunit-915a5f2c-3d37-4e41-b5bc-ef1efca9c01e?slide=39

tests/CSSList/DocumentTest.php
```diff
@@ -85,4 +85,66 @@ public function setContentsReplacesContentsSetInPreviousCall()
 
         self::assertSame($contents2, $this->subject->getContents());
     }
+
+    /**
+     * @return \Generator<array<array<DeclarationBlock>, DeclarationBlock, DeclarationBlock, array<DeclarationBlock>>>
```
oliverklee, line 90: I think my previous comment about the array notation was misleading and incorrect. Please have a look at the PHPStan documentation on this:

https://phpstan.org/writing-php-code/phpdoc-types#general-arrays
https://phpstan.org/writing-php-code/phpdoc-types#array-shapes

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,22 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
 
+    /**
+     * Insert an item before its sibling.
+     *
+     * @param RuleSet|CSSList|Import|Charset $oItem The item.
+     * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
+     */
+    public function insertBefore($oItem, $oSibling): void
+    {
+        $iIndex = in_array($oSibling, $this->aContents);
```
oliverklee, line 307: Please always use the strict version of `in_array`:

```suggestion
        $iIndex = in_array($oSibling, $this->aContents, true);
```

oliverklee, line 307: Also, this is a boolean (so the `i` prefix is misleading). I suggest to change the name altogether to allow it to communicate the intent: `$siblingExistsInList`

What do you think?

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,22 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
+        if ($iIndex === false) {
+            $this->append($oItem);
+        } else {
+            $this->replace($oSibling, array($oItem, $oSibling));
```
oliverklee, line 311: Let's use the short array notation:

```suggestion
            $this->replace($oSibling, [$oItem, $oSibling]);
```

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,22 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
         array_splice($this->aContents, $iOffset, $iLength, $mReplacement);
     }
 
+    /**
+     * Insert an item before its sibling.
```
oliverklee, line 300: In general, please always use the third person for the first sentence of a method documentation comment: "(This method) Does …":

```suggestion
     * Inserts an item before its sibling.
```

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,22 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
+     *
+     * @param RuleSet|CSSList|Import|Charset $oItem The item.
+     * @param RuleSet|CSSList|Import|Charset $oSibling The sibling.
```
oliverklee, line 303: IMHO the two "prose parts" of the PHPDoc comment are redundant with the parameter names, and I propose removing `The item.` and `The sibling.`.

oliverklee (changes requested):

src/CSSList/CSSList.php
ziegenberg: Shouldn't we add this rule then? https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/strict/strict_param.rst

src/CSSList/CSSList.php
ziegenberg: They are removed.

tests/CSSList/DocumentTest.php
ziegenberg: I hope I got it right this time.

tests/CSSList/DocumentTest.php
ziegenberg: Got it.

Is there a possibility to automate this check?

src/CSSList/CSSList.php
ziegenberg: Fixed.

tests/CSSList/DocumentTest.php
ziegenberg: I tend to dissagree. As the method name suggests, we insert an item. Yes, it does this by internally mapping to the existing `replace()` function, but in the end, for the consumer the item is inserted and not replaced. It does append (inserts at the end, before the not found sibbling), if it can't find the spot to insert the item, but this two different insertion points are reflected in the naming of the test data in the dataprovider.

Splitting this in to two test methods basiaclly doublicates the code, because the test code would be exactly the same.

But if you insist, I will make the change.

tests/CSSList/DocumentTest.php
ziegenberg: Also, I switched it back to a simple array as I do not see the additional benefit of using a Generator in this case.

src/CSSList/CSSList.php
ziegenberg: Yes.

src/CSSList/CSSList.php
ziegenberg: Fixed.

ziegenberg:
I rebased onto the current `main`, while I was at it.

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,21 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
         array_splice($this->aContents, $iOffset, $iLength, $mReplacement);
     }
 
+    /**
+     * Inserts an item before its sibling.
```
oliverklee, line 300: We should also document what happens if the sibling is not in this list (i.e., the item gets appended).

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,21 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
+     *
+     * @param RuleSet|CSSList|Import|Charset $oItem
+     * @param RuleSet|CSSList|Import|Charset $oSibling
+     */
+    public function insertBefore($oItem, $oSibling): void
+    {
+        if (in_array($oSibling, $this->aContents, true) === false) {
```
oliverklee, line 307:
```suggestion
        if (!in_array($oSibling, $this->aContents, true)) {
```

src/CSSList/CSSList.php
```diff
@@ -296,6 +296,21 @@ public function splice($iOffset, $iLength = null, $mReplacement = null)
...
+            $this->append($oItem);
+        } else {
+            $this->replace($oSibling, [$oItem, $oSibling]);
```
oliverklee, line 310: I think having the default case (the intended case) makes this method easier to understand. What do you think?

tests/CSSList/DocumentTest.php
```diff
@@ -85,4 +85,62 @@ public function setContentsReplacesContentsSetInPreviousCall(): void
 
         self::assertSame($contents2, $this->subject->getContents());
     }
+
+    /**
+     * @return array<string, array<string, DeclarationBlock|array<int, DeclarationBlock>>>
+     */
+    public static function insertContentBeforeInsertsContentBeforeSibblingOrAppendsIfSibblingNotFoundDataProvider(): array
+    {
+
```
oliverklee, line 94: Extraneous blank line.

```suggestion
```

tests/CSSList/DocumentTest.php
```diff
@@ -85,4 +85,62 @@ public function setContentsReplacesContentsSetInPreviousCall(): void
...
+            ]
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @dataProvider insertBeforeDataProvider
+     */
+    public function insertContentBeforeInsertsContentBeforeSibblingOrAppendsIfSibblingNotFound(
```
oliverklee, line 130: Let's split this into two tests. (The "or" in the test name is a sign that this test wants to be two tests as it tests two different pieces of behavior.)

oliverklee (changes requested):
This feature also deserves a changelog entry.

src/CSSList/CSSList.php
oliverklee: Yes, definitely: #584

tests/CSSList/DocumentTest.php
oliverklee: I remember that there was a Rector rule for this, but I don't know which one off the top of my head.

tests/CSSList/DocumentTest.php
oliverklee: Yes, please. (I find some duplication in tests okay as long as it helps them tell an easy-to read story.)

oliverklee: We probably then also can do without the data providers and instead include the data directly in the corresponding tests.

src/CSSList/CSSList.php
ziegenberg: Kinda "fixed" it.

src/CSSList/CSSList.php
ziegenberg: I switched the two cases.

src/CSSList/CSSList.php
ziegenberg: Done.

tests/CSSList/DocumentTest.php
ziegenberg: I separated the two tests, as requested. Happy now?

ziegenberg: But where's the point of having data providers if you separate everything into its own testing method? This is now a lot of duplicate code. Take e.g. `expandBorderShorthand()` and its `expandBorderShorthandProvider()` in `tests/RuleSet/DeclarationBlockTest.php`. Shouldn't this rather be:

- `expandBorderShorthandExpandsBorderWithBorderStyleAndBorderColor`
- `expandBorderShorthandExpandsBorderStyle`
- `expandBorderShorthandExpandsBorderWidth`
- `expandBorderShorthandExpandsBorderColor`
- `expandBorderShorthandExpandsBorderWithBorderStyle`
- `expandBorderShorthandFixesWhitespace`

following the same approach as done here?

ziegenberg:
TLDR: Is there anything left open that needs addressing?

IMHO, this PR can be considered thoroughly reviewed in 53 comments, with about ten commits for one mere conditional and two function calls in one new method (not counting the testing code). I started this because since 2016, [Moodle](https://github.com/moodle/moodle) has been dragging PR #115 with every update to the PHP-CSS-Parser lib, and I wanted to fix this properly.

JakeQZ (approved):
This LGTM.  I agree there is a bit of duplication in setting up the test subject, but we could argue until the cows come home about the best way to avoid that duplication.

IMO, let's get this in as it is now, since it's critical path for other projects, then possibly review the test code duplication later.  (There's also #390 needed for an 8.6 release.)

tests/CSSList/DocumentTest.php
oliverklee: For testing `expandBorderShorthand`, I'd differentiate between two pieces of behavior:

1. that is expands different border things (I'd use a data provider for that)
2. that it fixes whitespace

So concerning splitting tests, I'd differentiate between "the same behavior, but with different data" and different behavior.

Let's have a non-code example - maybe this will help me communicate my thinking. :-)

- different behavior: I can eat. I can draw. I can spit things out.
- same behavior, but with different data (data providers): I can eat apples, bananas and bread.

Does this help, @ziegenberg?

Would it be helpful to have a call to talk about testing strategies and patterns?

Issue #432, opened 2024-02-03
Add Rector


oliverklee:
This will help us do automatic refactorings.

Also add a configuration file and prevent it from getting shipped.

ziegenberg:
@oliverklee can this be considered done?

oliverklee:
Yes. Thank you very much for drinin this forward! 🙏 ❤️

PR #546, opened 2024-06-17
[TASK] Add rector to composer.json with a simple null-config

ziegenberg:
Adding rector as a development dependency enables us to do automatic refactorings. Initially, this comes with a simple null-config, that does not do much but run rector, so following the suggested integration path of "taking it slow" [1].

Helps with #432

[1] https://getrector.com/documentation/integration-to-new-project


composer.json
```diff
@@ -31,7 +31,8 @@
         "phpstan/extension-installer": "^1.4.1",
         "phpstan/phpstan": "^1.11.4",
         "phpstan/phpstan-phpunit": "^1.4.0",
-        "phpunit/phpunit": "^8.5.38"
+        "phpunit/phpunit": "^8.5.38",
+        "rector/rector": "^1.1"
```
oliverklee, line 35: For development dependencies, please let's always use a specific version number:

```suggestion
        "rector/rector": "^1.1.0"
```

composer.json
```diff
@@ -70,32 +71,38 @@
         ],
         "ci:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests",
         "ci:php:stan": "phpstan --no-progress --configuration=config/phpstan.neon",
+        "ci:php:rector": "\"./vendor/bin/rector\" --no-progress-bar --no-diffs --dry-run --config=config/rector.php",
```
oliverklee, line 74:
```suggestion
        "ci:php:rector": "rector --no-progress-bar --no-diffs --dry-run --config=config/rector.php",
```

Also, I think that having the diff might be helpful to the human reader. What do you think?

composer.json
```diff
@@ -70,32 +71,38 @@
...
         "ci:static": [
             "@ci:php:fixer",
-            "@ci:php:stan"
+            "@ci:php:stan",
+            "@ci:php:rector"
```
oliverklee, line 78: Please let's sort those lines alphabetically (says my inner Monk ;-) ).

composer.json
```diff
@@ -70,32 +71,38 @@
...
         ],
         "ci:tests": [
             "@ci:tests:unit"
         ],
         "ci:tests:sof": "\"./vendor/bin/phpunit\" --stop-on-failure --do-not-cache-result",
         "ci:tests:unit": "\"./vendor/bin/phpunit\" --do-not-cache-result",
         "fix:php": [
-            "@fix:php:fixer"
+            "@fix:php:fixer",
+            "@ci"
```
oliverklee, line 87: This probably should be `"@fix:php:rector"`, I'd guess?

oliverklee (changes requested):
Thanks for contributing this PR, @ziegenberg! :heart:

composer.json
ziegenberg: Sorted!

composer.json
ziegenberg: Fixed.

composer.json
ziegenberg: Fixed it to the suggestesd v1.1.0

composer.json
ziegenberg: Yes, you're right.

ziegenberg:
I addressed your comments, rebased onto the current main and force-pushed my changes. Please let me know if you would rather have fixup commits for later squashing them.

composer.json
```diff
@@ -70,8 +71,10 @@
         ],
         "ci:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests",
         "ci:php:stan": "phpstan --no-progress --configuration=config/phpstan.neon",
+        "ci:php:rector": "\"./vendor/bin/rector\" --no-progress-bar --dry-run --config=config/rector.php",
```
oliverklee, line 74: For binaries supplied by Composer packages, we can use them directly in Composer commands without the `vendor/bin` path:

```suggestion
        "ci:php:rector": "rector --no-progress-bar --dry-run --config=config/rector.php",
```

oliverklee (changes requested):

oliverklee:
> Please let me know if you would rather have fixup commits for later squashing them.

Yes, I'd prefer that as that makes re-reviewing a PR a bit easier. Thanks for asking!

ziegenberg:
I addressed the last two comments regarding the Composer commands in a fixup commit. If you're done reviewing I can quash them.

oliverklee:
> If you're done reviewing I can quash them.

Thanks! There's no need, though - GitHub allows me to squash all commits of a PR when I merge it.

composer.json
```diff
@@ -80,22 +83,26 @@
         "ci:tests:sof": "\"./vendor/bin/phpunit\" --stop-on-failure --do-not-cache-result",
         "ci:tests:unit": "\"./vendor/bin/phpunit\" --do-not-cache-result",
         "fix:php": [
-            "@fix:php:fixer"
+            "@fix:php:fixer",
+            "@fix:php:rector"
         ],
         "fix:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix bin src tests",
+        "fix:php:rector": "vendor/bin/rector --config=config/rector.php",
```
oliverklee, line 90:
```suggestion
        "fix:php:rector": "rector --config=config/rector.php",
```

oliverklee (changes requested):

composer.json
ziegenberg: And also fixed.

oliverklee (approved):
Looks good. Thanks! :pray:

PR #550, opened 2024-06-17
[TASK] Add rector ruleset php71

ziegenberg:
This enables the rector rule set for PHP 7.1.


ziegenberg:
I rebased onto the current main.

oliverklee (approved):

PR #552, opened 2024-06-17
[CLEANUP] rector: Changes unneeded null check to ?? operator

ziegenberg:
This applies the rule TernaryToNullCoalescingRector.

For details see: https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#ternarytonullcoalescingrector


oliverklee (approved):
LGTM. Thanks!

Issue #433, opened 2024-02-03
Use PHP 7.2 language features


oliverklee:
… but not additional type declarations yet.

Rector (#432) will do this for us.

ziegenberg:
IMHO, this can be considered closed, with #554 being merged?

oliverklee:
Yes, we can indeed close it. Thanks!

PR #554, opened 2024-06-17
[TASK] rector: set PHP rule-set to the same version as composer.json

ziegenberg:
This way rector always follows the minimal required version in composer.json, which currently is PHP 7.2.

This helps with #433.


oliverklee:
This now needs a rebase.

ziegenberg:
And it's rebased.

oliverklee (approved):

Issue #512, opened 2024-03-01
Deprecate `expandShorthands`/`createShorthands`/`create*Shorthand`/`expand*Shorthand`


oliverklee:
https://github.com/MyIntervals/PHP-CSS-Parser/issues/510#issuecomment-1972628778

Expanding and creating the shorthand notation is out of scope for this library. So we should ~~remove the functionality, make all the corresponding methods a no-op and~~ mark ~~them~~ these methods as `@deprecated #511 will be removed in version 10`.

JakeQZ:
We could provide a link in the readme to a tag for the last release which includes the functionality, to help anyone who wants to include it in their own project, and/or encourage someone to build it into a separate package.

ziegenberg:
One deprecation at a time:

- [ ] #558
- [ ] #566
- [ ] #567
- [ ] #568
- [ ] #569
- [ ] #570
- [ ] #571
- [ ] #572
- [ ] #573
- [ ] #574
- [ ] #575
- [ ] #578

sabberworm:
@ziegenberg IMHO you could have put the deprecations into a single PR instead of doing a new PR for every single method. Thanks a lot for your contributions, though.

oliverklee:
> IMHO you could have put the deprecations into a single PR instead of doing a new PR for every single method.

@ziegenberg talked/wrote about this, and I requested them to be single PRs. This allows for smaller PRs and allows us to easily revert single things that haven't worked out.

ziegenberg:
I surely would have put them into one PR. Would have been much easier and would require a *lot* less rebasing, namely 0.

That's another thing to put into the CONTRIBUTING.md, together with all other kinds of policies.

sabberworm:
It seems @oliverklee and I disagree about what is an atomic change…

sabberworm:
I guess a compromise would have been to make it a single PR but one commit for each change. That would have made them revertable without all the drawbacks.

ziegenberg:
> Then we would have had a merge commit, and I find a linear history really helpful.
> 
> For cases like this, I'd be fine with having separate PRs for the deprecations and a follow-up PR to add the changelog entries. This would help avoid the merge conflicts and the need for a rebase. Would that be helpful for you, @ziegenberg?

Yes, that would be helpful next time.

This time, Whether I rebase this pile of PRs or collect all changelog entries from every PR, package it into a new PR and adjust all PRs, it makes no difference in the work I have to do.

[...]

PR #558, opened 2024-06-18
[TASK] Deprecate `DeclarationBlock::expandShorthands`

ziegenberg:
The `expandShorthands`/`createShorthands` Functions are deprecated and will be removed without substitution in version 10.0.  Expanding and creating the shorthand notation is out of the scope of this library. If you want to include this functionality in your project or build it into a separate package, get the code from the v8.5.1 version of this library.

Helps with fixing #512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Deprecated
 
-- `DeclarationBlock::expandShorthands()` is deprecated and will be removed without substitution in version 10.0.
+- Deprecate `DeclarationBlock::expandShorthands()` (#558)
 
 ### Removed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Deprecated
 
+- `DeclarationBlock::expandShorthands()` is deprecated and will be removed without substitution in version 10.0.
+
 ### Removed
 
 - Drop support for PHP < 7.2 (#420)
--- a/src/RuleSet/DeclarationBlock.php
+++ b/src/RuleSet/DeclarationBlock.php
@@ -186,8 +186,12 @@ class DeclarationBlock extends RuleSet
      */
     public function expandShorthands()
     {
-        trigger_error("Expanding Shorthands will be removed without substitution in version 10.0.");
-        return;
+        // border must be expanded before dimensions
+        $this->expandBorderShorthand();
+        $this->expandDimensionsShorthand();
+        $this->expandFontShorthand();
+        $this->expandBackgroundShorthand();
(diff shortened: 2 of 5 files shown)


ziegenberg:
I removed the now superfluous tests.

src/RuleSet/DeclarationBlock.php
```diff
@@ -181,15 +181,13 @@ public function getSelectors()
      * Splits shorthand declarations (e.g. `margin` or `font`) into their constituent parts.
      *
      * @return void
+     *
+     * @deprecated This will be removed without substitution in version 10.0.
```
sabberworm, line 185: I think we should add the deprecated marker now but not change anything about the functionality until we hit version 10.

oliverklee (changes requested):
Let's deprecate it (with a corresponding changelog entry), but keep the functionality intact.

In the commit message and PR title, this should be a `[TASK]`, not a mere (code) `[CLEANUP]`.

src/RuleSet/DeclarationBlock.php
ziegenberg: I'm not sure, because I was just following the suggestions from https://github.com/MyIntervals/PHP-CSS-Parser/issues/512#issue-2162927498 where @oliverklee mentioned to

> [...] remove the functionality, make all the corresponding methods a no-op and mark them as @deprecated [...].

Is this still the way to go, or do we want to just put a `@deprecated` on the function? Also, shall we emit a warning?

CHANGELOG.md
```diff
@@ -18,6 +18,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Deprecated
 
+- `DeclarationBlock::expandShorthands()` is deprecated and will be removed without substitution in version 10.0.
```
oliverklee, line 21:
```suggestion
- Deprecate `DeclarationBlock::expandShorthands()`
```

src/RuleSet/DeclarationBlock.php
oliverklee: I'd say we only mark it as `@deprecated`. (I honestly don't remember why I tended to make it a no-op in #512). This will allow people to adapt their code when they see that their code is calling a deprecated method.

oliverklee (changes requested):

src/RuleSet/DeclarationBlock.php
ziegenberg: So no `trigger_error()`?

oliverklee: Yes, exactly. (In the future, we're going to also have a separate test suite for deprecated code that will allow deprecation warnings. Then we'll also be able to use `trigger_error`.)

ziegenberg: I marked it as deprecated and (partly) reverted the commits.

oliverklee (approved):
LGTM.

Issue #489, opened 2024-02-22
Add contrubutor guidelines


JakeQZ:
Relates to #486.

Something along the lines of https://github.com/MyIntervals/emogrifier/blob/main/.github/CONTRIBUTING.md

The developer instructions at the end of the readme can then be moved here, and replaced with a link.

oliverklee:
Also link https://cbea.ms/git-commit/ for git commit messages and https://speakerdeck.com/oliverklee/test-driven-development-with-phpunit-915a5f2c-3d37-4e41-b5bc-ef1efca9c01e?slide=39 for test naming.

oliverklee:
Also mention that the changelog is in reverse chronological order (newest on top).

ziegenberg:
> Also link https://cbea.ms/git-commit/ for git commit messages and https://speakerdeck.com/oliverklee/test-driven-development-with-phpunit-915a5f2c-3d37-4e41-b5bc-ef1efca9c01e?slide=39 for test naming.

I'm assuming that is only additional. As someone on the consuming side of contributor guidelines, I appreciate having those things readily available in a document within the project. That way, it is directly accessible to me using the same tool I'm currently working in, and I can search for it. I wouldn't say I like having to click several links, read text from images or dig through a lengthy blog post and contribute to someone else's statistical data by visiting 3rd party websites.

ziegenberg:
Other things to mention in the guidelines: 

> In general, please always use the third person for the first sentence of a method documentation comment: "(This method) Does …":

ziegenberg:
Probably also add, that this project uses [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation).

oliverklee:
> Probably also add, that this project uses [Hungarian notation](https://en.wikipedia.org/wiki/Hungarian_notation).

We'd actually like to get rid of that as part of the [9.0 milestone](https://github.com/MyIntervals/PHP-CSS-Parser/milestone/1):

> rename internal things (including parameter names) to ditch the Hungarian notation

JakeQZ:
Also (though this is probably more for maintainers and collaborators), for PRs that (possibly indirectly) change coding style and guidance, or the set of developer tools and how they are installed, update the `CONTRIBUTING.md` file.  The file referenced in the OP has become out of date since this has not been done (https://github.com/MyIntervals/emogrifier/issues/373).

JakeQZ:
[...]

PR #560, opened 2024-06-18
[TASK] Remove `DeclarationBlock::setSelector`

ziegenberg:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 - Drop support for PHP < 7.2 (#420)
+- Final removal of `DeclarationBlock::setSelector()` (#560)
 
 ### Fixed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,8 +20,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Removed
 
+- Remove `DeclarationBlock::setSelector()` (#560)
 - Drop support for PHP < 7.2 (#420)
-- Final removal of `DeclarationBlock::setSelector()` (#560)
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ### Removed
 
 - Drop support for PHP < 7.2 (#420)
+- Final removal of `DeclarationBlock::setSelector()` (#560)
```
oliverklee, line 24: Our changelog is in reverse chronological order, i.e., please add new entries at the top of the list, not at the bottom. (Ditto for the other PRs that come with a changelog entry.)

And a note on commit messages/PR titles/changelog entries: Those should use the imperative voice. In this case:

> Remove `DeclarationBlock::setSelector()`

I found this article on git commit messages very helpful (and fun to read!): https://cbea.ms/git-commit/

We can also drop the "final" there as we already have the distinction between "deprecate …" and "remove …".

Ditto for the other PRs.

oliverklee (changes requested):

CHANGELOG.md
ziegenberg: I will revise the changelog messages.

> And a note on commit messages/PR titles/changelog entries: Those should use the imperative voice. In this case:
> 
> > Remove `DeclarationBlock::setSelector()`
> 
> I found this article on git commit messages very helpful (and fun to read!): https://cbea.ms/git-commit/

Can this be put in a CONTRIBUTING.md or similar file? It seems like every project has its own very helpful blog article someone found, (which is also fun to read), and they all contradict each other in one way or another. I wasn't able to find any notes on how to contribute to this project so I tried to infer the unwritten rules from what I found via `git log`.

I will update the commit messages as requested.

oliverklee: > Can this be put in a CONTRIBUTING.md or similar file?

Excellent suggestion!

I've added it to https://github.com/MyIntervals/PHP-CSS-Parser/issues/489#issuecomment-2176154113.

ziegenberg: I fixed the changelog message.

ziegenberg:
I'm not going to fix the commit message, as that would require a force-push to have any effect, and that would contradict the desired fix-up commit. But I will remember the commit message format for future PRs.

oliverklee (approved):
Thanks! :pray:

PR #562, opened 2024-06-18
[TASK] Remove `Rule::setValues`

ziegenberg:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,9 +22,9 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Removed
 
+- Remove `Rule::setValues()` (#562)
 - Remove `DeclarationBlock::setSelector()` (#560)
 - Drop support for PHP < 7.2 (#420)
-- Final removal of `Rule::setValues()` (#562)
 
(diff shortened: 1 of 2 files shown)


ziegenberg:
I fixed the changelog message.

ziegenberg:
Rebased and force-pushed.

JakeQZ (approved):

oliverklee:
@JakeQZ I think you're merging `main` into the PR branches to update them (which is the default way suggested by GitHub AFAIK). Instead, we should rebase the PRs on `main` .

JakeQZ:
> @JakeQZ I think you're merging `main` into the PR branches to update them (which is the default way suggested by GitHub AFAIK). Instead, we should rebase the PRs on `main` .

The GitHub method seemed to work fine in cases with just a conflict in CHANGELOG.  It was quick and easy whilst reviewing to click 'Resolve Conflicts' and simply remove the diff markers (since the conflict was multiple additions at the same line).  However, in #563 there was also a conflict in another file, which I thought I had resolved correctly, then after committing realized I hand't, so tried to revert the merge, spectacularly unsuccessfully.

Issue #551, opened 2024-06-17
Add a Composer Script for PHP linting and use it on CI


oliverklee:
Also lint `config/` in order to catch e.g. language features that are not supported in all the PHP versions we support.

@ziegenberg Would you be willing to tackle this one?

ziegenberg:
Would you be ok with introducing [PHPCSStandards/PHP_CodeSniffer](https://github.com/PHPCSStandards/PHP_CodeSniffer/)?

oliverklee:
I think we have a misunderstanding here. :-)

This is about the PHP syntax linting using `php -l` (which we [already have](https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/.github/workflows/ci.yml#L14), but not as a Composer script).

For style checking, we're already using [PHP-CS-Fixer](https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/composer.json#L72). I'd prefer to keep this (as this project seems to be maintained a lot more actively than PHP_CodeSniffer). And I'd prefer to not use both tools in the same project as in my experience it can be quite cumbersome to fix situations where the styles required by both tools disagree with each other, resulting in a style whack-a-mole. ;-)

ziegenberg:
Can this be changed so that PHP-CS-Fixer is also installed via Composer? PHP-CSS-Parser would be the only project I'm working on to require phive for installing dev tools. With Composer we would also benefit from simpler CI tasks and dependabot would inform us about updates.

_Sidenote: I'm not sure how the impression was created that PHP_CodeSniffer is not as well maintained. I can see regular releases [^1]. Also, PHP-CS-Fixer only claims PHP 7.4+ support[^2]._


[^1]: https://github.com/PHPCSStandards/PHP_CodeSniffer/releases
[^2]: https://github.com/PHP-CS-Fixer/PHP-CS-Fixer?tab=readme-ov-file#supported-php-versions

oliverklee:
> Can this be changed so that PHP-CS-Fixer is also installed via Composer? 

Unfortunately not (at least at the moment). We want to have the latest version to make use of the (WIP) PER-2 ruleset, and having friendsofphp/php-cs-fixer as a Composer (dev) dependency would make this project uninstallable in PHP 7.2 and 7.3 (as friendsofphp/php-cs-fixer  requires PHP >= 7.4).

[...]

PR #565, opened 2024-06-18
[TASK] Move linting to a Composer script and use it on CI

ziegenberg:
Also lint `config/` in order to catch e.g. language features that are not supported in all the PHP versions we support.

Fixes: #551

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -96,6 +96,7 @@
         "ci": "Runs all dynamic and static code checks.",
         "ci:dynamic": "Runs all dynamic code checks (i.e., currently, the unit tests).",
         "ci:php:fixer": "Checks the code style with PHP CS Fixer.",
+        "ci:php:lint": "Checks the syntax of the PHP code.",
         "ci:php:stan": "Checks the types with PHPStan.",
         "ci:php:rector": "Checks the code for possible code updates and refactoring.",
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -69,11 +69,13 @@
         "ci:dynamic": [
             "@ci:tests"
         ],
-        "ci:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests",
-        "ci:php:stan": "phpstan --no-progress --configuration=config/phpstan.neon",
+        "ci:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests config",
+        "ci:php:lint": "find src tests config bin -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
```
oliverklee, line 73: Please let's also have a line in the `scripts-description` section of the `composer.json`.

oliverklee (changes requested):
Thanks! Looks good! One thing to add to get this to mergeable.

composer.json
ziegenberg: Fixed.

oliverklee (approved):
Looks good to me. Thanks!

PR #566, opened 2024-06-18
[TASK] Deprecate `Document::expandShorthands`

ziegenberg:
The `expandShorthands`/`createShorthands` Functions are deprecated and will be removed without substitution in version 10.0. Expanding and creating the shorthand notation is out of the scope of this library. If you want to include this functionality in your project or build it into a separate package, get the code from the v8.5.1 version of this library.

Helps with fixing #512

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Deprecated
 
+- Deprecate `Document::expandShorthands()` (#566)
 - Deprecate `Document::createShorthands()` (#567)
 - Deprecate `DeclarationBlock::expandShorthands()` (#558)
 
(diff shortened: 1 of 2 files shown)


ziegenberg:
Rebased it.

ziegenberg:
Rebased it.

JakeQZ (commented):
I think the OP comment should somehow be included in the CHANGELOG, so that it would be in turn included in the release note, so that people are directed to previous functionality if they want to implement it themselves as a separate libraray.  Maybe immediately under the heading `## x.y.z`.  @oliverklee WDYT?

ziegenberg:
Originally I had this comment within the first PR of this series in the Readme.md at the very bottom.

oliverklee (approved):

oliverklee:
> I think the OP comment should somehow be included in the CHANGELOG, so that it would be in turn included in the release note, so that people are directed to previous functionality if they want to implement it themselves as a separate libraray.

I think it suffices to have the pair of "deprecate this" and (in the next release) "remove this" in the changelog. This pattern always means for developers to "please stop using this, and use something else instead or stick to an old version" and "the same, but definitely now".

sabberworm (approved):

ziegenberg:
It's rebased.

PR #581, opened 2024-06-18
[CLEANUP] Rector: Add strict return type based on returned strict expr type

ziegenberg:
This applies the rule **ReturnTypeFromStrictBoolReturnExprRector**. For Details see:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#returntypefromstrictboolreturnexprrector

### DIFF
--- a/src/CSSList/AtRuleBlockList.php
+++ b/src/CSSList/AtRuleBlockList.php
@@ -74,9 +74,6 @@ class AtRuleBlockList extends CSSBlockList implements AtRule
         return $sResult;
     }
 
-    /**
-     * @return bool
-     */
     public function isRootList(): bool
     {
         return false;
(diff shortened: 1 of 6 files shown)


ziegenberg:
Rebased it.

ziegenberg:
Should those PHPDoc `@return` statements also be deleted?

JakeQZ:
> Should those PHPDoc `@return` statements also be deleted?

I think so, if they don't provide anything more than the type hint.

oliverklee (changes requested):
Thanks! Yes, please let's drop those redundant `@return` PHPDoc annotations.

ziegenberg:
Rebased and removed the PHPDoc annotations.

oliverklee (approved):
LGTM. Thanks! :pray:

PR #563, opened 2024-06-18
[TASK] Remove `Rule::getValues`

ziegenberg:
Removing one function after another.

PR #582, opened 2024-06-19
[TASK] Remove `Rule::getValues`

ziegenberg:

Supersedes #563


JakeQZ (approved):
Thanks for recovering it.  Apologies for the mess-up <3

PR #584, opened 2024-06-19
[TASK] Always use the strict version of `in_array` & friends

oliverklee:


oliverklee:
We'll need some casts first (in a separate PR).

src/RuleSet/DeclarationBlock.php
```diff
@@ -338,20 +338,21 @@ public function expandFontShorthand(): void
...
                 $aFontProperties['font-style'] = $mValue;
             } elseif ($mValue == 'small-caps') {
                 $aFontProperties['font-variant'] = $mValue;
             } elseif (
-                in_array($mValue, ['bold', 'bolder', 'lighter'])
-                || ($mValue instanceof Size
-                    && in_array($mValue->getSize(), range(100, 900, 100)))
+                in_array($mValue, ['bold', 'bolder', 'lighter'], true)
+                // Note: font weights should be integers, not floats. For now, we'll keep using floats. We will
+                // drill down on this and fix it in  https://github.com/MyIntervals/PHP-CSS-Parser/issues/612.
```
JakeQZ, line 354: This comment needs changing (or removing), now that we have closed #612 on the basis that CSS treats all numerical values as a `number`, regardless of whether they may be floating-point.

JakeQZ (changes requested):
The comment re `font-weight` needs changing or removing.  Otherwise LGTM.

src/RuleSet/DeclarationBlock.php
oliverklee: Done.

JakeQZ (approved):
LG now.  (Though I didn't notice any 'friends'.)

PR #587, opened 2024-06-19
[CLEANUP] Rector: Change return type based on strict returns type operations

ziegenberg:
This applies the rule **BoolReturnTypeFromStrictScalarReturnsRector**. For Details see:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#boolreturntypefromstrictscalarreturnsrector


oliverklee (approved):
Looks good to me. Thanks!

PR #591, opened 2024-06-19
[CLEANUP] Rector: Add return type to function like with return new

ziegenberg:
This applies the rule **ReturnTypeFromReturnNewRector**. For Details see:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#returntypefromreturnnewrector

### DIFF
--- a/src/CSSList/Document.php
+++ b/src/CSSList/Document.php
@@ -29,7 +29,7 @@ class Document extends CSSBlockList
      *
      * @throws SourceException
      */
-    public static function parse(ParserState $oParserState)
+    public static function parse(ParserState $oParserState): Document
     {
         $oDocument = new Document($oParserState->currentLine());
         CSSList::parseList($oParserState, $oDocument);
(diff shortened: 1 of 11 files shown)


src/CSSList/Document.php
```diff
@@ -29,7 +29,7 @@ public function __construct($iLineNo = 0)
      *
      * @throws SourceException
      */
-    public static function parse(ParserState $oParserState)
+    public static function parse(ParserState $oParserState): Document
```
oliverklee, line 32: We also can drop the `@return` PHPDoc annotation here.

oliverklee (changes requested):
Love it! Only one PHPDoc annotation to remove.

src/CSSList/Document.php
ziegenberg: Fixed it.

oliverklee (approved):

PR #592, opened 2024-06-19
[TASK] Update PHP-CS-Fixer

oliverklee:


config/php-cs-fixer.php
```diff
@@ -21,6 +19,9 @@
             '@PHPUnit75Migration:risky' => true,
             '@PHPUnit84Migration:risky' => true,
 
+            // needed for PHP 7.x
+            'trailing_comma_in_multiline' => ['after_heredoc' => true, 'elements' => ['arrays', 'match']],
+
```
JakeQZ, line 24: I agree with the setting.  However, the comment may be misleading.  According to the [docs](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/blob/master/doc/rules/control_structure/trailing_comma_in_multiline.rst), the default is `['after_heredoc' => false, 'elements' => ['arrays']]`, which would not break PHP 7.2 (which doesn't allow trailing comma after last argument in function/method call) or PHP 7.x (which doesn't allow trailing comma after last parameter in function/method declaration).

config/php-cs-fixer.php
```diff
@@ -1,8 +1,6 @@
 <?php
 
-if (PHP_SAPI !== 'cli') {
-    die('This script supports command line usage only. Please check your command.');
-}
```
JakeQZ, line 5: Agree we probably want `strict_types` and not the CLI test.  If and as long as the tool only supports CLI, surely that would be reported elsewhere.

JakeQZ (commented):
I just have a query about the comment in the config file, otherwise looks fine.

config/php-cs-fixer.php
oliverklee: Good point! Done, rebased and repushed.

JakeQZ: What about the comment being misleading?  Really, the setting is including `match` which is a PHP >= 8.0 language construct (so won't be having any effect until we drop PHP 7.x and start using PHP 8 language constructs, however will be in place ready for such time).

JakeQZ (commented):
I still think the comment is misleading - unless the defaults have actually changed and the docs are out of date.

JakeQZ:
> I still think the comment is misleading - unless the defaults have actually changed and the docs are out of date.

It appears the defaults have changed, breaking PHP 7.4 (even though it is supported by the tool), and all CI on whatever PHP version.  The latest version thus breaks semantic versioning.

(I ran `phive update`. it installed the latest 3.x version, and I got loads of errors.)

Does the tool claim to follow semantic versioning?  If not, then we should have a tilde version contraint in `phars.xml`.  Or based on the above, should probably have that anyway.

config/php-cs-fixer.php
oliverklee: Done (hopefully).

oliverklee:
> Does the tool claim to follow semantic versioning?

No, I don't think so.

> If not, then we should have a tilde version contraint in `phars.xml`. Or based on the above, should probably have that anyway.

I'd prefer not to: In `phars.xml`, we also store the exact version that will be installed by `phive install`, so this will only be relevant when someone uses `phive update`. And in that case, I'd like to get the latest version. (And yes, I know that a new version might require autoformatting the code or changes to the configuration to not break PHP 7.x.)

oliverklee:
> It appears the defaults have changed, breaking PHP 7.4 (even though it is supported by the tool),

That PHP-CS-Fixer supports PHP 7.4 means that it will _run_ on PHP 7.4., not that it will create PHP 7.4-compliant code with all possible configurations. (The PER2 standard has some rules that only work on PHP 8.x. We'll need to avoid those for the time being.)

JakeQZ (approved):
The comment now accurately conveys the reason for the setting, I think.

JakeQZ:
> That PHP-CS-Fixer supports PHP 7.4 means that it will _run_ on PHP 7.4

Apparently not, with the `match` option explicitly enabled.

oliverklee:
> Apparently not, with the `match` option explicitly enabled.

Oh, I wasn't aware of that. So my learning is: Whether PHP-CS-Fixer runs under a specific PHP version also depends on the configuration of the rules. Thanks for pointing that out! :pray:

JakeQZ:
Looking again, it seems that the default for `trailing_comma_in_multiline` has actually changed from `['after_heredoc' => false, 'elements' => ['arrays']]` to `['after_heredoc' => false, 'elements' => ['arrays', 'arguments']]`, since PHP 7.4 supports trailing commas in function call arguments.

I think this is a bug in PHP-CS-Fixeer.  The execution environment is irrelevant.  The target environment is the important one.  Also, since `match` itself is a PHP 8 language construct, it cannot possibly matter if a rule that could only apply to PHP 8 code is run in a PHP 7 environment.  It is possible to have separate code paths based on PHP version.

We should remove the `match` option and/or file a bug with PHP-CS-Fixer.

For completeness, here is the error:

```
> "./.phive/php-cs-fixer" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests config
PHP CS Fixer 3.59.3 (064efa1) 7th Gear by Fabien Potencier, Dariusz Ruminski and contributors.
PHP runtime: 7.4.16
Running analysis on 1 core sequentially.
You can enable parallel runner and speed up the analysis! Please see https://cs.symfony.com/doc/usage.html for more information.
Loaded config default from "config/php-cs-fixer.php".
Using cache file ".php-cs-fixer.cache".

In ConfigurableFixerTrait.php line 81:

  [PhpCsFixer\ConfigurationException\InvalidForEnvFixerConfigurationException (32)]
  [trailing_comma_in_multiline] Invalid configuration for env: "match" option can only be enabled with PHP 8.0+.


Exception trace:
  at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Fixer/ConfigurableFixerTrait.php:81
 PhpCsFixer\Fixer\ControlStructure\TrailingCommaInMultilineFixer->configure() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/FixerFactory.php:174
 PhpCsFixer\FixerFactory->useRuleSet() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/ConfigurationResolver.php:336
 PhpCsFixer\Console\ConfigurationResolver->getFixers() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/Command/FixCommand.php:322
 PhpCsFixer\Console\Command\FixCommand->execute() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Command/Command.php:298
 Symfony\Component\Console\Command\Command->run() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Application.php:1040
 Symfony\Component\Console\Application->doRunCommand() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/Application.php:177
 PhpCsFixer\Console\Application->doRunCommand() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Application.php:301
 Symfony\Component\Console\Application->doRun() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/Application.php:98
 PhpCsFixer\Console\Application->doRun() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Application.php:171
 Symfony\Component\Console\Application->run() at C:\Users\Jake\.phive\phars\php-cs-fixer-3.59.3.phar:102

In TrailingCommaInMultilineFixer.php line 121:

  [PhpCsFixer\FixerConfiguration\InvalidOptionsForEnvException]
  "match" option can only be enabled with PHP 8.0+.


Exception trace:
  at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Fixer/ControlStructure/TrailingCommaInMultilineFixer.php:121
 PhpCsFixer\Fixer\ControlStructure\TrailingCommaInMultilineFixer::PhpCsFixer\Fixer\ControlStructure\{closure}() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/options-resolver/OptionsResolver.php:1152
 Symfony\Component\OptionsResolver\OptionsResolver->offsetGet() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/options-resolver/OptionsResolver.php:924
 Symfony\Component\OptionsResolver\OptionsResolver->resolve() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/FixerConfiguration/FixerConfigurationResolver.php:130
 PhpCsFixer\FixerConfiguration\FixerConfigurationResolver->resolve() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Fixer/ConfigurableFixerTrait.php:73
 PhpCsFixer\Fixer\ControlStructure\TrailingCommaInMultilineFixer->configure() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/FixerFactory.php:174
 PhpCsFixer\FixerFactory->useRuleSet() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/ConfigurationResolver.php:336
 PhpCsFixer\Console\ConfigurationResolver->getFixers() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/Command/FixCommand.php:322
 PhpCsFixer\Console\Command\FixCommand->execute() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Command/Command.php:298
 Symfony\Component\Console\Command\Command->run() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Application.php:1040
 Symfony\Component\Console\Application->doRunCommand() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/Application.php:177
 PhpCsFixer\Console\Application->doRunCommand() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Application.php:301
 Symfony\Component\Console\Application->doRun() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/src/Console/Application.php:98
 PhpCsFixer\Console\Application->doRun() at phar://C:/Users/Jake/.phive/phars/php-cs-fixer-3.59.3.phar/vendor/symfony/console/Application.php:171
 Symfony\Component\Console\Application->run() at C:\Users\Jake\.phive\phars\php-cs-fixer-3.59.3.phar:102

fix [--path-mode PATH-MODE] [--allow-risky ALLOW-RISKY] [--config CONFIG] [--dry-run] [--rules RULES] [--using-cache USING-CACHE] [--cache-file CACHE-FILE] [--diff] [--format FORMAT] [--stop-on-violation] [--show-progress SHOW-PROGRESS] [--sequential] [--] [<path>...]

Script "./.phive/php-cs-fixer" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests config handling the ci:php:fixer event returned with error code 32
```

JakeQZ:
Since found it's not a BC introduced in PHP-CS-Fixer itself, but a in the `@PER-CS2.0` ruleset update.

PR #594, opened 2024-06-19
[CLEANUP] Make use of GitHub username autoparsing in the README

oliverklee:


oliverklee:
Ah, doesn't work. :-( Closing.

sabberworm:
I would probably have rejected this even if it had worked, since I don’t think we should tie ourself too strongly to one ecosystem (and READMEs are often displayed in various contexts, not just GH).

JakeQZ:
Privacy is also a concern.  Maybe only partially related: I am getting far too much spam because I foolishly signed up to GitHub with my main email address, which is now plastered all over the place in an easily-scrapable form.

Issue #461, opened 2024-02-12
Add a debugging log


oliverklee:
For every error that currently is throwing an exception or that is worked around using the lenient mode, we should log the problem, the line number and the corresponding CSS fragment (where it makes sense).

We should log this using a [PSR-compatible logger](https://packagist.org/packages/psr/log) and provide a simple default logger to `Parser`. It should be possible to override the logger from the outside (so people can plug in their own logging if they like).

We'll also need to `psr/log` to our dependencies.

JakeQZ:
Would using `psr/log` allow us to provide a 'null' logger by default, possibly meaning we don't need to actually additionally provide an on/off switch?

oliverklee:
Yes, `psr/log` provides a `NullLogger` which we can use: https://github.com/php-fig/log/blob/master/src/NullLogger.php

For the non-null logger, we can extend `https://github.com/php-fig/log/blob/master/src/AbstractLogger.php` with our own `SimpleLogger` (or something along the lines) so people don't need to implement their own logging for simple debugging.

Issue #463, opened 2024-02-12
Drop the lenient mode switch


oliverklee:
This is blocked by #461.

We should always parse in the lenient way now. The configuration for the lenient should be a no-op, and the corresponding methods/properties should be marked as `@deprecated #462 will be removed in version 10`. (#462 is the ticket for dropping the configuration switch altogether.)

PR #596, opened 2024-06-20
[FEATURE] Add support for PSR/Log

ziegenberg:
- Add PSR/Log as dependency
- Add a new SimpleLogger
- Use the NullLogger per default

Helps with #461

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -24,7 +24,8 @@
     ],
     "require": {
         "php": ">=7.2.0",
-        "ext-iconv": "*"
+        "ext-iconv": "*",
+        "psr/log": "*"
     },
     "require-dev": {
         "codacy/coverage": "^1.4.3",
--- a/src/SimpleLogger.php
+++ b/src/SimpleLogger.php
@@ -1,0 +1,34 @@
+<?php
+
+namespace Sabberworm\CSS;
+
+use Psr\Log\AbstractLogger;
+
+/**
+ * This class provides a simple logging facility.
+ */
+class SimpleLogger extends AbstractLogger
+{
+    public function log($level, $message, array $context = []): void
+    {
+        echo self::interpolate($message, $context) . PHP_EOL;
+    }
+
+    /**
+     * Interpolates context values into the message placeholders.
+     */
+    private function interpolate(string $message, array $context = []): string
+    {
+        // build a replacement array with braces around the context keys
+        $replace = [];
+        foreach ($context as $key => $val) {
+            // check that the value can be cast to string
+            if (!is_array($val) && (!is_object($val) || method_exists($val, '__toString'))) {
+                $replace['{' . $key . '}'] = $val;
+            }
+        }
+
+        // interpolate replacement values into the message and return
+        return strtr($message, $replace);
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+
+- Add support for PSR/Log with a new `SimpleLogger` and use `NullLogger` per default (#596)
 - Support arithmetic operators in CSS function arguments (#607)
 - Add support for inserting an item in a CSS list (#545)
 - Add a class diagram to the README (#482)
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -23,14 +23,21 @@ use Sabberworm\CSS\Value\CSSString;
 use Sabberworm\CSS\Value\URL;
 use Sabberworm\CSS\Value\Value;
 
+use Psr\Log\LoggerInterface;
+use Psr\Log\LoggerAwareInterface;
+use Psr\Log\LoggerAwareTrait;
+use Psr\Log\NullLogger;
+
 /**
  * This is the most generic container available. It can contain `DeclarationBlock`s (rule sets with a selector),
  * `RuleSet`s as well as other `CSSList` objects.
  *
  * It can also contain `Import` and `Charset` objects stemming from at-rules.
  */
-abstract class CSSList implements Renderable, Commentable
+abstract class CSSList implements Renderable, Commentable, LoggerAwareInterface
 {
+    use LoggerAwareTrait;
+
     /**
      * @var array<array-key, Comment>
      */
@@ -51,6 +58,8 @@ abstract class CSSList implements Renderable, Commentable
      */
     public function __construct($iLineNo = 0)
     {
+        $this->logger = new NullLogger();
+
         $this->aComments = [];
         $this->aContents = [];
         $this->iLineNo = $iLineNo;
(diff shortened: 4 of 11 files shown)


ziegenberg:
@oliverklee would this be the direction you thought of?

JakeQZ (commented):
The motivation for adding this was to allow removal of the lenient/strict mode parsing switch.  See #463.

All CSS should be parsed as best as possible, with invalid constructs dropped and logged.  This is what browsers do.

The strict mode setting allowed parsing exceptions to propagate such that parsing failed, or caused exceptions to be thrown.  Instead, we want them to be logged.

Searching for `bLenientParsing` will indicate where the behaviour deviates.  I think initially we would want logging in only those situations, where strict parsing would cause an error, with a view to replacing the strict parsing option with logging.

I haven't looked through these changes in detail, but am aware some exceptions are intentional and should not be logged.

JakeQZ:
So I am thinking that the logging should be taking place where the exceptions are handled (if the conditional `bLenientParsing` comes into play), rather than where they are thrown.  Though in cases where `!bLenientParsing` causes an exception to be thrown, logging should take place there (and we can later remove the `throw` as part of #462).

ziegenberg:
@oliverklee Any chance I could a quick 👍🏻 or 👎🏻 for this? Yes, this is a work in progress, but before I invest more time, I'd like to know if this is going in the right direction or if I'm completely off track.

oliverklee:
@ziegenberg I regret that you didn't get feedback from us on the general direction of this PR. I'm sorry for this.

I personally am currently in a conflict in another open source project, which impacts my mental wellbeing, and I try to stick to the smaller/easier reviews during this time.

I'll do my best to write something on the architecture and direction of this PR in a few days.

Until then, please bear with me/us.

JakeQZ:
In the meantime, I've had another brief look.

I think the use of `LoggerAwareInferface`, `LoggerAwareTriat`, etc., looks perfect: exact;ly what we want.  (I just noticed you had a bunch of `static` methods to deal with too.  I like the solution there too.)

My main concern is the duplication where exceptions are thrown and something identical is logged.  Can the logging be done downstream, using the data from the exception where it is caught?  This would have to be reviewed on a case-by-case basis; I appreceiate, @ziegenberg, that you probably don't have 100% familiarity with the code base (neither do I, FWIW - you're probaby now as familiar with it as me; we really appreciate your top-notch contributions so far; besides the actual improvements, your involvment has, I think, helped motivate the rest of us to move this project forwards; much kudos).

HTH <3

@oliverklee, hope you resolved the disupute/conflict (hope it wasn't with me - if it was, I was unaware :) )

src/SimpleLogger.php
```diff
@@ -0,0 +1,34 @@
...
+            // check that the value can be cast to string
+            if (!is_array($val) && (!is_object($val) || method_exists($val, '__toString'))) {
+                $replace['{' . $key . '}'] = $val;
+            }
+        }
+
+        // interpolate replacement values into the message and return
+        return strtr($message, $replace);
+    }
+}
```
JakeQZ, line 34: Do we need to provide this?  Shouldn't it be up to the users of the package to provide their own logger, or one from a library, should they want logging?

JakeQZ:
> Can the logging be done downstream, using the data from the exception where it is caught? This would have to be reviewed on a case-by-case basis.

Sorry, I haven't had time to review the specific cases, though I suspect in most it can be.

JakeQZ:
Can wie have the introduction of the logger interfaces in solo, without any actual logging, as a separate initial pre-PR?

I feel there should be some unit testing for that, but am out of ideas right now.

JakeQZ:
> I ... am out of ideas right now.

``mplementsLogger` might be one such test.

oliverklee:
> hope you resolved the disupute/conflict (hope it wasn't with me - if it was, I was unaware :) )

@JakeQZ Well, it's mostly over (though not resolved). And no, it's not with you - if there was something to resolve with you, I would do my best to let you know so we can work together to resolve it. :heart:

composer.json
```diff
@@ -24,7 +24,8 @@
     ],
     "require": {
         "php": ">=7.2.0",
-        "ext-iconv": "*"
+        "ext-iconv": "*",
+        "psr/log": "*"
```
oliverklee, line 28: For dependencies that are actual packages (instead of PHP extensions), please always use specific versions:

```suggestion
        "psr/log": "^1.0.0 || ^2.0.0 || ^3.0.0"
```

oliverklee:
I regret that it took me so long to give some general feedback on this. So here we go. :-) 

All classes should use the same logger instance when a document is parsed. This will allow users to get all logged warnings/errors in one place. For this, we need a class that gets the original logger instance and then passes it on to the other classes. Maybe the parser?

Also, the instance should be provided when the corresponding classes are instantiated so the desired logger will be always available. (I think for the parser, it's fine to start with the `NullLogger` by default and allow setting a different logger later.

Also, maybe it would help if we split this up into smaller PRs to get this reviewed and merged faster:
1. introduce the `NullLogger` and provide it to the classes
2. add the non-null logger
3. actually log stuff

We also need to either allow the logger getting set via dependency injection (particularly, Symfony DI), or at least paint us into a corner with our API DI-wise.

JakeQZ:
> All classes should use the same logger instance when a document is parsed ... For this, we need a class that gets the original logger instance and then passes it on to the other classes. Maybe the parser?

I think the `Settings` class can handle this.  The logger instance can be added as a property of that.

> Also, the instance should be provided when the corresponding classes are instantiated so the desired logger will be always available.

An instance of `ParserState` is passed around everywhere during parsing.  This has a `getSettings` method.

> (I think for the parser, it's fine to start with the `NullLogger` by default and allow setting a different logger later.)

If the logger instance is a property of `Settings`, it can be switched midway through parsing, or later, if the various classes are referencing the `Settings` (if they need to) rather than the logger directly.  (If that's what you meant.)

JakeQZ:
> 2\. add the non-null logger

Should we be providing a non-null logger in the release package?  Shouldn't users be providing a logger of their choice (from a library)?  For development/testing, is there some kind of mock/testing logger available that plays well with PHPUnit?

oliverklee:
> Should we be providing a non-null logger in the release package? Shouldn't users be providing a logger of their choice (from a library)? For development/testing, is there some kind of mock/testing logger available that plays well with PHPUnit?

I think packaging the `NullLogger` should be enough. We can use our non-null logger in the tests (and move it the the testing namespace/folder, e.g., some `Fixtures` folder).

oliverklee:
> I think the `Settings` class can handle this.  The logger instance can be added as a property of that.
> …
> An instance of `ParserState` is passed around everywhere during parsing.  This has a `getSettings` method.

Having the logger attached to the `Settings` class doesn't feel right to me. Logging is not part of configuration. Having it attached to the `ParserState` sounds okay to me.

JakeQZ:
> I think packaging the `NullLogger` should be enough. We can use our non-null logger in the tests (and move it the the testing namespace/folder, e.g., some `Fixtures` folder).

Agree.

> Having the logger attached to the `Settings` class doesn't feel right to me. Logging is not part of configuration. Having it attached to the `ParserState` sounds okay to me.

This is debatable (in the literal sense, asking a philosophical question).  I'm happy either way.

I the practical sense, I don't think it makes much difference, as `ParserState` is always available during parsing, and we are primarily, for now, only interested in logging parsing issues.  However, the library also provides for manipulation of the CSS objects after parsing (though, offhand, I'm not sure a `Settings` object is any more readily-available then); this is worth bearing in mind but I don't think anything of particular concern for the time being.

JakeQZ:
@sabberworm, does what we're saying seem reasonable approach-wise?  (Not asking for specific code changes review at this stage, but if you spot a flaw in the plan, that would be good to know.)

PR #597, opened 2024-06-20
[CLEANUP] Rector: Add return type void to function like without any return

ziegenberg:
This applies the rule **AddVoidReturnTypeWhereNoReturnRector**. For Details see:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#addvoidreturntypewherenoreturnrector

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -57,12 +57,10 @@ abstract class CSSList implements Renderable, Commentable
     }
 
     /**
-     * @return void
-     *
      * @throws UnexpectedTokenException
      * @throws SourceException
      */
-    public static function parseList(ParserState $oParserState, CSSList $oList)
+    public static function parseList(ParserState $oParserState, CSSList $oList): void
     {
         $bIsRoot = $oList instanceof Document;
         if (is_string($oParserState)) {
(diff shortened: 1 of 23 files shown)


src/CSSList/CSSList.php
```diff
@@ -263,7 +263,7 @@ public function getLineNo()
      *
      * @return void
```
oliverklee, line 264: We can also drop all those `@return void` PHPDoc annotations now.

oliverklee (changes requested):

src/CSSList/CSSList.php
ziegenberg: PHPDoc annotations are gone.

JakeQZ (commented):
Should the Rector config file be updated - specifically to uncomment `AddVoidReturnTypeWhereNoReturnRector::class`?

oliverklee (approved):

ziegenberg:
> Should the Rector config file be updated - specifically to uncomment `AddVoidReturnTypeWhereNoReturnRector::class`?

IMHO, this can be done at a later stage. First, I'm doing all the **CLEANUP** and, later, the **TASK** of configuring the rector. Also, @oliverklee surely wants to have TASK and CLEANUP in different PRs.

After this type series, we can set the following config:
```PHP
    ->withSets(
        [
            SetList::TYPE_DECLARATION,
        ]
    )
```

PR #599, opened 2024-06-20
[CLEANUP] Extract CSSFunction methods parseName and parseArguments

JakeQZ:
Also
- Update the `parse` function signature to use PHP7 type hints;
- Remove subsequently redundant DocBlock entries;
- Ditto for same-named method in subclass to avoid contravariance.

This is the refactoring change from #390.

### DIFF
--- a/src/Value/CSSFunction.php
+++ b/src/Value/CSSFunction.php
@@ -34,21 +34,42 @@ class CSSFunction extends ValueList
     }
 
     /**
-     * @param ParserState $oParserState
-     * @param bool $bIgnoreCase
+     * @throws SourceException
+     * @throws UnexpectedEOFException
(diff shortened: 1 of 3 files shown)


JakeQZ:
Looking again, this change is independent of the functional change in #390, so isn't a prerequisite for that implementation.

oliverklee:
I think I have found a more specific solution to the `parse()` return type in #608, which I'd like to get in first as a pre-patch.

JakeQZ:
> I think I have found a more specific solution to the `parse()` return type in #608, which I'd like to get in first as a pre-patch.

I've rebased with that change included.

oliverklee (approved):
LGTM.

PR #602, opened 2024-06-21
[CLEANUP] Rector: Add strict return type based native function or class method return

ziegenberg:
This applies the rule **ReturnTypeFromStrictNativeCallRector**. For Details see:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#returntypefromstrictnativecallrector


oliverklee (approved):
Looks good to me. Thanks!

PR #604, opened 2024-06-21
[CLEANUP] Rector: Add union return type

ziegenberg:
This applies the rule **ReturnUnionTypeRector**. For Details see:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#returnuniontyperector

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -83,8 +83,8 @@
         "ci:tests": [
             "@ci:tests:unit"
         ],
-        "ci:tests:sof": "\"./vendor/bin/phpunit\" --stop-on-failure --do-not-cache-result",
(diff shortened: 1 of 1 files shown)


src/Value/Color.php
```diff
@@ -26,12 +26,10 @@ public function __construct(array $aColor, $iLineNo = 0)
      * @param ParserState $oParserState
      * @param bool $bIgnoreCase
      *
-     * @return Color|CSSFunction
```
oliverklee, line 29: For those cases with a real union type, we'll need to keep the PHPDoc.

oliverklee:
I have created #608 to add native return types to `parse()`. Maybe we can get that in first.

ziegenberg:
I rebased and it closed on itself.

PR #605, opened 2024-06-21
[CLEANUP] Rector: Add return type to function like with return new

ziegenberg:
This applies the rule **ReturnTypeFromReturnNewRector**. For Details see:
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#returntypefromreturnnewrector

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -83,8 +83,8 @@
         "ci:tests": [
             "@ci:tests:unit"
         ],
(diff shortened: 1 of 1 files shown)


oliverklee:
I have created #608 to add native return types to `parse()`. Maybe we can get that in first.

ziegenberg:
I rebased and it closed on itself.

PR #607, opened 2024-06-21
[FEATURE] Support arithmetic operators in CSS functions

JakeQZ:
This is the enhancement from #390.

### DIFF
--- a/tests/Value/ValueTest.php
+++ b/tests/Value/ValueTest.php
@@ -60,8 +60,8 @@ final class ValueTest extends TestCase
     public static function provideMalformedLengthOperands(): array
     {
         return [
-            'LHS missing value' => ['vh', '10px'],
-            'RHS missing value' => ['50vh', 'px'],
+            'LHS missing number' => ['vh', '10px'],
+            'RHS missing number' => ['50vh', 'px'],
             'LHS missing unit' => ['50', '10px'],
             'RHS missing unit' => ['50vh', '10'],
         ];
--- a/tests/Value/ValueTest.php
+++ b/tests/Value/ValueTest.php
@@ -45,13 +45,38 @@ final class ValueTest extends TestCase
     }
 
     /**
-     * @test
+     * @return array<string, array{0: string, 1: string}>
+     * The first datum is a template for the parser (using `sprintf` insertion marker `%s` for some expression).
+     * The second is for the expected result, which may have whitespace and trailing semicolon removed.
      */
-    public function parsesArithmeticWithMultipleOperatorsInFunctions(): void
+    public static function provideCssFunctionTemplates(): array
     {
-        $subject = Value::parseValue(new ParserState('calc(300px + 10% + 10vw);', Settings::create()));
+        return [
+            'calc' => [
+                'to be parsed' => 'calc(%s);',
+                'expected' => 'calc(%s)',
+            ],
+            'max' => [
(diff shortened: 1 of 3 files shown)


tests/Value/ValueTest.php
```diff
@@ -0,0 +1,82 @@
+<?php
+
```
oliverklee, line 2: Let's make this file strict.

oliverklee (changes requested):
Looks good all in all. Only one nitpick.

tests/Value/ValueTest.php
JakeQZ: Done.

oliverklee (approved):
LGTM. Thanks for picking this up! ❤️

oliverklee (changes requested):
Ah, no, stop the press: We still need a changelog entry for this new feature.

JakeQZ:
I've added strict types declaration to the new (test) source file.

@sabberworm, can you give this a quick look to confirm you think it's OK for now.

Longer term, I think it would be cleaner to introduce a `Value\Expression` class to encapsulate arithmetic.  For now, this change allows construction from parsing of a `RuleValueList` with components like `{0: Size, 1: string, 2: Size}` where the string is the operator.  Longer term we'd want to support nested expressions with brackets, and direct expressions as values (both of which I think CSS4 allows, the latter with syntax like `width: (100px + 10%)`, i.e. omission of the `calc` keyword, though can't offhand find a reference).

JakeQZ:
> Ah, no, stop the press: We still need a changelog entry for this new feature.

Indeed we do.  It's lack was a deliberate alertness test ;)

Also corrected nomenclature of the numberic part of a length value (MDN calls it a 'number' not a 'value').

JakeQZ:
I've updated this again.  I noted that arithmetic in `calc` already worked, so the test I had for `calc(A + B + C)` was already passing.  So I've extended that test to also include `max(X, A + B + C)`.

JakeQZ:
> Longer term, I think it would be cleaner to introduce a `Value\Expression` class to encapsulate arithmetic

Just re-spotted that #389 seeks to do just that.  Though it is a PR on top of a PR, so may be somewhat awkward to unravel.

oliverklee:
> Longer term, I think it would be cleaner to introduce a `Value\Expression` class to encapsulate arithmetic. 

Yes, absolutely. I'd also like to check our class naming and structure against the terms and concepts used in the the [official specification](https://www.w3.org/TR/css-syntax-3/).

JakeQZ:
> I'd also like to check our class naming and structure against the terms and concepts used in the the [official specification](https://www.w3.org/TR/css-syntax-3/).

I'm finding a close if not exact match there.  Kudos to @sabberworm.  But, yes, please do review.

JakeQZ:
> Just re-spotted that #389 seeks to do just that. Though it is a PR on top of a PR, so may be somewhat awkward to unravel.

Looking at that, the new `Expression` class doesn't seem to be connected to anything, so it seems a WIP.  Although I had briefly looked at it before, I had forgotten.  So I would consider it two people independently coming up with the same idea...

oliverklee (approved):
Neat!

oliverklee:
> Yes, absolutely. I'd also like to check our class naming and structure against the terms and concepts used in the the [official specification](https://www.w3.org/TR/css-syntax-3/).

… but not as part of this PR. :-)

PR #608, opened 2024-06-21
[CLEANUP] Add native return types to `parse()`

oliverklee:


JakeQZ (approved):
This deals with the apparent PHP 7.2/7.3 covariance bug better than my efforts in #599.

Reading again, I see that [covariance is not supposed to be supported](https://www.php.net/manual/en/language.oop5.variance.php) in those PHP versions, though they seem to allow a more-specific return type in a subclass if there is *no* return type sepecified in the superclass :S

I'll rebase #599 with this change...

PR #609, opened 2024-06-21
[CLEANUP] Ensure all tests actually have an assertion

oliverklee:
This avoids the "This test did not perform any assertions" warning from PHPUnit.


JakeQZ (approved):
I had also noticed the "1 risky test".

An alternative would have been to declare the test method to 'expect no exception', whereas what you have put in actually strenthens the testing, so is double-bubble.

PR #610, opened 2024-06-22
[BUGFIX] Support Lint on Windows

### LINKED ISSUE #598: Lint does not work on Windows
See https://github.com/MyIntervals/emogrifier/issues/740

JakeQZ:
Use `php-parallel-lint` package to avoid *nix-specific command.

Fixes #598.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -28,6 +28,7 @@
     },
     "require-dev": {
         "codacy/coverage": "^1.4.3",
+        "php-parallel-lint/php-parallel-lint": "^1.4.0",
         "phpstan/extension-installer": "^1.4.1",
         "phpstan/phpstan": "^1.11.5",
         "phpstan/phpstan-phpunit": "^1.4.0",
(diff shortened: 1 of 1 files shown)


JakeQZ:
I don't understand the errors.  This is exaclty the same as we have in Emogrifier, and it works fine locally.

Are the GitHub actions missing a `composer update`?

.github/workflows/ci.yml
```diff
@@ -30,6 +30,9 @@ jobs:
                   tools: composer:v2
                   coverage: none
 
+            - name: Composer Update
+              run: composer update
```
oliverklee, line 34: We'll also need the caching (and, to keep things consistent to ease maintenance for us, move the PHP version into a matrix). I'll create a pre-patch for this.

composer.json
```diff
@@ -70,7 +71,7 @@
             "@ci:tests"
         ],
         "ci:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests config",
-        "ci:php:lint": "find src tests config bin -name '*.php' -print0 | xargs -0 -n 1 -P 4 php -l",
+        "ci:php:lint": "\"./vendor/bin/parallel-lint\" src tests config bin",
```
oliverklee, line 74:
```suggestion
        "ci:php:lint": "parallel-lint src tests config bin",
```

JakeQZ: This works for me.  I think it used not to.  I just copied what was in Emogrifier.  I guess Composer has been updated not to require a specific path to the `bin` directory.

JakeQZ:
Had to use `git rebase --skip`.  The latest version of `git` seems a bit broken, or maybe just incompatible with GitHub.  Eggs in a basket.  Don't put them all in.

oliverklee (approved):
LGTM.

PR #611, opened 2024-06-22
[TASK] Install Composer dependencies for PHP lint CI job

oliverklee:
This is a pre-patch before we can switch the PHP linting to the parallel lint package in #610.


JakeQZ (approved):
Looks fine.  It duplicates chunks from all the other sections.

Is there any way of avoiding such duplication in GitHub actions - like calling some common function to do the set-up?

PR #614, opened 2024-06-23
[CLEANUP] Simplify calls to Composer scripts from Composer scripts

oliverklee:


JakeQZ (approved):
At some point in the last couple of years, a Composer update (note not a `composer update`) has allowed script commands to Composer-installed components to be used without a path.

PR #625, opened 2024-06-28
[TASK] Prepare the 8.5.2 release

oliverklee:


JakeQZ (approved):
I realized after creating PRs for 8.6 that this would need to be done first if using the same branch.

oliverklee:
> I realized after creating PRs for 8.6 that this would need to be done first if using the same branch.

Exactly. :slightly_smiling_face: I've just [released V8.5.2](https://github.com/MyIntervals/PHP-CSS-Parser/releases/tag/v8.5.2) so we can get on with the features targeted at 8.6.0.

PR #624, opened 2024-06-28
[FEATURE] Support arithmetic operators in CSS functions (#607)

JakeQZ:
This is the enhancement from #390.

`calc` was already supported.

PR #628, opened 2024-06-28
[TASK] Avoid poor scaling of array_search() with very long arrays (#413)

JakeQZ:
When there are many many elements in $aStack, starting the delimiter search from the beginning for each loop is very very slow. I addressed this by building a new array rather than modifying $aStack in place and iterating over it in a single pass. The particular 1.6M style string that was giving me trouble went from taking 4 minutes to parse to 5 seconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
+- Improve performance of Value::parseValue with many delimiters by refactoring to remove array_search()
+
 ### Deprecated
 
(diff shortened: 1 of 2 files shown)


JakeQZ:
This set of PRs (this, #627, #623, #624 - in chronological order of original changes) will need rebasing after each other is merged, due to conflict in `CHANGELOG.md`.

They are the current set of enhancements that can be backported for 8.6.  I've created a `v8.5.x` branch from the current `v8.x` in case we need to apply any fixes to 8.5.

oliverklee (approved):

PR #629, opened 2024-06-29
[TASK] Remove another PHP-8-only setting from the PHP-CS-Fixer config

oliverklee:


JakeQZ (approved):
Yeah, I found this post-review after approving #592.  After some disentangling, it seems the new PHP-CS-Fixer includes an updated `PER-CS2.0` ruleset that is incompatible with PHP7 (including 7.4).

PR #630, opened 2024-07-01
[TASK] Prepare the 8.6.0 release

JakeQZ:
There don't seem to be any more enhancements in the immediate pipeline, so it would be good to get this out now with the four we have, most or all of which I think have been requested by users.

(I'm not sure I can do the actual release.  I've never made one before, and don't know what's involved/needed.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,18 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 ## x.y.z
 
 ### Added
+
+### Changed
+
+### Deprecated
+
+### Removed
+
+### Fixed
+
+## 8.6.0
+
+### Added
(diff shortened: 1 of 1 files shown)


oliverklee (approved):

oliverklee:
I've just done the release. These were the steps:

```bash
git switch v8.x
git tag
git tag -a -s v8.6.0 -m "Tag version 8.6.0"
git push --tags 
```

… and then create release on GitHub from the tag, copying the changelog. Also I closed the milestone on GH.

Notes on the commands:

1. The first `git tag` is not strictly necessary. I usually do it nonetheless to check that the previous tags are there, and to remind myself of the format (which for the life of me I cannot remember).
2. The `-s` switch in the second `git tag` command means that I GPG-sign the tag (which requires that I've GPG set up for git).

PR #631, opened 2024-07-01
[TASK] Add more PHP-CS-Fixer rules

oliverklee:
Also drop some rules that are already in included in the used
rulesets.

### DIFF
--- a/config/php-cs-fixer.php
+++ b/config/php-cs-fixer.php
@@ -22,19 +22,76 @@ return (new \PhpCsFixer\Config())
             // overwrite the PER2 defaults to restore compatibility with PHP 7.x
             'trailing_comma_in_multiline' => ['elements' => ['arrays']],
 
-            // for `in_array` & friends
-            'strict_param' => true,
+            // casing
+            'magic_constant_casing' => true,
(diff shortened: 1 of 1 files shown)


oliverklee:
Note: I'll need to add some type casts (in separate PRs) before we can make all files strict.

JakeQZ (commented):
> Also drop some rules that are already in included in the used rulesets.

Actually, the rules being removed have the value being set as the default.  So I'm wondering if they should be kept, to be explicit, or removed anyway, to avoid clutter.

Otherwise, this brings the config in line with the Emogrifier project, so we have consistency.

oliverklee:
> Actually, the rules being removed have the value being set as the default. So I'm wondering if they should be kept, to be explicit, or removed anyway, to avoid clutter.

I don’t care much either way - as long as the config file is consistent between our two projects. :-)

JakeQZ (approved):
> I don’t care much either way - as long as the config file is consistent between our two projects. :-)

Neither do I, TBH.  It was just an observation.  Emogrifier doesn't have them, so for consistency we may as well remove them here.

PR #632, opened 2024-07-01
[TASK] Prefix native function calls with a backslash

oliverklee:
This slightly improves performance.

### DIFF
--- a/bin/quickdump.php
+++ b/bin/quickdump.php
@@ -7,7 +7,7 @@
 
 require_once(__DIR__ . '/../vendor/autoload.php');
 
-$sSource = file_get_contents('php://stdin');
+$sSource = \file_get_contents('php://stdin');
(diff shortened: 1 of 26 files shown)


sabberworm (approved):
IMHO the ugliness on display here is not worth any performance gains. But if we do this, then, at least, it should be and remain consistent. Can we add some rule that to one of our checkers to enforce (or, better, auto-fix) this?

JakeQZ:
> IMHO the ugliness on display here is not worth any performance gains.

It also helps clarify to the reader that function being called is in the global namespace (and thus likely to be a built-in PHP function), rather than in the current namespace or some other namespace being `use`d (without having to look that up or have encyclopaedic knowledge).  (It's true we don't have any classless functions, or AFAIK use any from another library, but the reader may not know that.)

(Ugliness is in the eye of the beholder.  I find these a thing of beauty.  They satisfy my inner OCD ;))

PR #638, opened 2024-07-02
[CLEANUP] Sort the types in union type PHPDoc annotations

oliverklee:
This was done using the `phpdoc_types_order` PHP-CS-Fixer rule.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -97,7 +97,7 @@ abstract class CSSList implements Renderable, Commentable
     }
 
     /**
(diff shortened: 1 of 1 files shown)


JakeQZ (commented):
> This was done using the `phpdoc_types_order` PHP-CS-Fixer rule.

Is this with `['null_adjustment' => 'always_last', 'sort_algorithm' => 'none']`?

JakeQZ (approved):
> Yes, exactly that.

Thanks.  I just wanted to confirm before approving.

PR #641, opened 2024-07-04
[TASK] Make all PHP files strict

oliverklee:


JakeQZ (approved):
I guess the required casts have been added in other changes.

PR #657, opened 2024-08-23
[FEATURE] Add PHP 8.4 to the CI matrix

oliverklee:


oliverklee:
> So there is no need to back-port anything for a minor release.

I'd still like to backport this for two reasons:
1. We then at least have PHP 8.4 linting for the maintenance version as well.
2. We communicate via the changelog that 8.x officially supports PHP 8.4

Would that be okay for you?

JakeQZ:
> I'd still like to backport this for two reasons:
> 
> 1. We then at least have PHP 8.4 linting for the maintenance version as well.
> 2. We communicate via the changelog that 8.x officially supports PHP 8.4
> 
> Would that be okay for you?

Yes.  Good points, particularly the first.

PR #663, opened 2024-08-24
[BUGFIX]  Fix comment parsing to support multiple comments

### LINKED ISSUE #173: bug with parsing multiple comments

Consider the following cases:

ziegenberg:
Because of an eager consumption of whitespace, the rule parsing would swallow a trailing comment, meaning the comment for the next rule would be affected. This patch addresses this by only consuming real whitespace without comments after a rule.

Fixes #173

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -220,17 +220,25 @@ class ParserState
     }
 
     /**
-     * @return array<int, Comment>|void
+     * Consumes whitespace and comments and returns the comments.
+     * If $withoutComment is true, only whitespace is consumed.
+     *
+     * @param bool $withoutComment Do not consume comments, only whitespace.
+     *
+     * @return array<int, Comment>|void List of comments.
      *
      * @throws UnexpectedEOFException
      * @throws UnexpectedTokenException
      */
-    public function consumeWhiteSpace(): array
+    public function consumeWhiteSpace(bool $withoutComment = false): array
     {
         $aComments = [];
         do {
             while (\preg_match('/\\s/isSu', $this->peek()) === 1) {
                 $this->consume(1);
+            }
+            if ($withoutComment) {
+                break;
             }
             if ($this->oParserSettings->bLenientParsing) {
                 try {
--- a/tests/ParserTest.php
+++ b/tests/ParserTest.php
@@ -1161,13 +1161,16 @@ body {background-color: red;}';
      */
     public function flatCommentExtracting(): void
     {
-        $parser = new Parser('div {/*Find Me!*/left:10px; text-align:left;}');
+        $parser = new Parser('div {/*Find Me!*/left:10px; /*Find Me Too!*/text-align:left;}');
         $doc = $parser->parse();
         $contents = $doc->getContents();
         $divRules = $contents[0]->getRules();
-        $comments = $divRules[0]->getComments();
-        self::assertCount(1, $comments);
-        self::assertSame('Find Me!', $comments[0]->getComment());
+        $rule1Comments = $divRules[0]->getComments();
+        $rule2Comments = $divRules[1]->getComments();
+        self::assertCount(1, $rule1Comments);
+        self::assertCount(1, $rule2Comments);
+        self::assertEquals('Find Me!', $rule1Comments[0]->getComment());
+        self::assertEquals('Find Me Too!', $rule2Comments[0]->getComment());
     }
 
     /**
--- a/src/Rule/Rule.php
+++ b/src/Rule/Rule.php
@@ -105,7 +105,7 @@ class Rule implements Renderable, Commentable
         while ($oParserState->comes(';')) {
             $oParserState->consume(';');
         }
-        $oParserState->consumeWhiteSpace();
+        $oParserState->consumeWhiteSpace(true);
 
         return $oRule;
     }


ziegenberg:
It would be nice to have this in an 8.6.1 release.

src/Parsing/ParserState.php
```diff
@@ -220,18 +220,26 @@ public function parseCharacter($bIsForIdentifier)
     }
 
     /**
-     * @return array<int, Comment>|void
+     * Consumes whitespace and comments and returns the comments.
+     * If $withoutComment is true, only whitespace is consumed.
```
oliverklee, line 224: Please put code in comments in backticks so it gets displayed as code.

src/Parsing/ParserState.php
```diff
@@ -220,18 +220,26 @@ public function parseCharacter($bIsForIdentifier)
...
+     *
+     * @param bool $withoutComment Do not consume comments, only whitespace.
+     *
+     * @return array<int, Comment>|void List of comments.
```
oliverklee, line 228: We can be even more specific (and omit the void return type, which is incorrect for this method and also is not allowed in union types):

```suggestion
     * @return list<Comment> list of comments
```

src/Parsing/ParserState.php
```diff
@@ -220,18 +220,26 @@ public function parseCharacter($bIsForIdentifier)
...
      *
      * @throws UnexpectedEOFException
      * @throws UnexpectedTokenException
      */
-    public function consumeWhiteSpace(): array
+    public function consumeWhiteSpace(bool $withoutComment = false): array
```
oliverklee, line 233: I'm not a big fan of boolean function parameters that switch behavior. Instead, please let's have a separate, well-named method, e.g., `comsumeWhitespaceWithoutComment` or something alike.

tests/ParserTest.php
```diff
@@ -1161,13 +1161,16 @@ public function commentExtracting(): void
      */
     public function flatCommentExtracting(): void
```
oliverklee, line 1162: We're planning to split the tests and testcases into smaller, more focused tests. To make this easier for us, please don't expand the test. Instead, could you please duplicate it into two separate tests, one with a single comment and one with multiple? Thanks!

src/Parsing/ParserState.php
oliverklee: Maybe we can then also move the new method to `Rule`. (I'm not 100% sure about this and would welcome your thoughts on this, @ziegenberg @sabberworm @JakeQZ.)

oliverklee (changes requested):
Thanks a lot for this PR, @ziegenberg! :heart: 

Looks generally good to me. :+1: I've added some review comments.

Once this is merged, I'm all for backporting this to the 8.x branch and release the 8.x branch once support for PHP 8.4 is merged as well.

src/Parsing/ParserState.php
ziegenberg: I would go even further and rename the current function `comsumeWhitespace` to `comsumeWhitespaceWithComment` and create a new function `comsumeWhitespace` which precisely only consumes white space. I was baffled to see `comsumeWhitespace` also consume comments and return them.

ziegenberg: Yes, this would probably be a breaking change. So, I introduced the boolean function parameter to stay within the current wording (which I deem wrong) and still make the change non-breaking.

I would consider having `comsumeWhitespace` and a new `comsumeWhitespaceWithoutComment` a not-so-good solution. That would cement the semantically wrong situation.

oliverklee: I like the idea of renaming this method!

Would you be willing to do this as a (or more specific), two pre-patch PRs?

1. one (to be backported) that renames the method, adds the specific type annotation, and re-adds the old method to redirect to the new method, and marks it as `@deprecated will be removed in version 9.0` (including a corresponding line in the changelog
2. another (not to be backported) that removes the old method (including a corresponding entry in the changelog)

I'd be willing to review these two pre-PRs quickly so we can rebase on them.

We also might want to mark the renamed method as `@internal` as it needs to be `public` for technical reasons, but is not expected to be called from outside this library. What do you think?

ziegenberg: So you want `comsumeWhitespaceWithComment` and `comsumeWhitespaceWithoutComment`?

oliverklee: Yes, exactly!

ziegenberg: I am not a big fan of `comsumeWhitespaceWithoutComment`, having a function name telling you what it does not do. But I'll create the two pre-patch PRs as asked.

oliverklee: Ah, it seems I misread your comment and got confused. :exploding_head: 

Okay, so I'm fine with this plan:

> I would go even further and rename the current function `comsumeWhitespace` to `comsumeWhitespaceWithComment` and create a new function `comsumeWhitespace` which precisely only consumes white space.

We'll need to do this in a non-breaking way for the parts that we backport, though.

oliverklee: > All `consume*` functions are in the class `ParserState`, so why move it to `Rule`?

I've thought about this some more and now think having it in `ParserState` is perfectly fine. :-)

JakeQZ:
> Battle plan as discussed in our call right now:
> 
>     1. 8.x We add a `consumeWhiteSpaceWithComments`
>     2. 8.x `consumeWhiteSpace` redirects to `consumeWhiteSpaceWithComments`, redirect it to the new method and mark it as deprecated
>     3. 8.x change the respective call to not call the `consume*` method and have the corresponding lines directly instead
>     4. 9.0 remove/change the deprecated method

I don't think we should ever reintroduce a method with the same name as one that previously existed, with different default funcionality.  It will catch out anyone doing a major update through several major versions, which happens even (and probably moreso) in sizeable companies as they defer the work required to support the update, deeming it low priority when updating requires some other changes.

So I think we may as well come up with a new method name now; maybe `consumeWhitespaceOnly`.  (Unfortunately, PHP method names are not case sensitive, so we can't use `consumeWhitespace`.)

oliverklee:
@ziegenberg Do we still need this PR now that #671/#672 have been merged?

oliverklee:
I propose we move the discussion of the next steps here: #679

JakeQZ:
#173 has finally been fixed by #1169; however, the suggestions here seem good for consideration moving forwards.

JakeQZ:
Various issues with comments in declaration blocks (including the trailing comment being dropped) are raised in #1445.  I've suggested an alternative approach there.

PR #664, opened 2024-08-24
[BUGFIX] Fix type errors in PHP strict mode

oliverklee:


oliverklee:
This fix fixes the type errors in strict mode as seen on this build: https://github.com/MyIntervals/PHP-CSS-Parser/actions/runs/10540068716/job/29204329414?pr=641

CHANGELOG.md
```diff
@@ -46,6 +46,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- Fix type errors in strict mode (#664)
```
JakeQZ, line 49: I think it should be clarified that this refers to PHP's strict mode.  The library itself also has a strict mode setting for the CSS (which we plan to remove).

Though I'm not sure this warrants a changelog entry at all.  It's not fixing any actual bug; merely allowing for `declare_strict` to be introduced, with a couple of explicit casts to replace the implicit ones.

JakeQZ (changes requested):
This is more of a code cleanup than an actual bugfix, so I'm not sure the changelog entry is warranted.  It's also ambiguous since the library itself (currently) has a 'strict mode' option for the CSS.

CHANGELOG.md
oliverklee: I've clarified this accordingly.

oliverklee: I'd like to keep the changelog entry. This was actually something where the code was incorrect (passing a `float` or `int` where only a `string` was allowed) - only PHP's implicit type conversations had allowed us to get away with this so far. :wink:

JakeQZ (approved):
👍

PR #666, opened 2024-08-25
[TASK] Avoid Hungarian notation in a testcase

oliverklee:


JakeQZ (approved):
Every journey begins with a single step.

PR #670, opened 2024-08-25
[FEATURE] Add consumeWhitespaceWithComments and redirect consumeWhitespace (backported)

ziegenberg:
Having `consumeWhitespace` also consumes comments, and returning them was unexpected. This also "deprecates" the consumeWhitespace method and announces a breaking change for 9.0.0.

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -226,12 +226,27 @@ class ParserState
     }
 
     /**
-     * @return array<int, Comment>|void
+     * Consumes whitespace and comments and returns a list of comments.
+     *
+     * @return list<Comment> List of comments.
      *
      * @throws UnexpectedEOFException
      * @throws UnexpectedTokenException
      */
     public function consumeWhiteSpace()
+    {
+        return $this->consumeWhiteSpaceWithComments();
+    }
+
+    /**
+     * Consumes whitespace and comments and returns a list of comments.
+     *
+     * @return list<Comment> List of comments.
+     *
+     * @throws UnexpectedEOFException
+     * @throws UnexpectedTokenException
+     */
+    public function consumeWhiteSpaceWithComments()
     {
         $aComments = [];
         do {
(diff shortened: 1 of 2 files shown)


src/Parsing/ParserState.php
```diff
@@ -226,12 +226,35 @@ public function parseCharacter($bIsForIdentifier)
...
+
+    /**
+     * Consumes whitespace and comments and returns a list of comments.
+     *
+     * @return list<Comment> List of comments.
+     *
+     * @throws UnexpectedEOFException
+     * @throws UnexpectedTokenException
+     */
+    public function consumeWhiteSpaceWithComments()
```
oliverklee, line 257:
```suggestion
    public function consumeWhiteSpaceWithComments(): array
```

oliverklee (changes requested):
Thanks! Could you also add a test for the new method? (Copying and modifying the test for the other method should be fine.)

src/Parsing/ParserState.php
ziegenberg: Ich befürchte, dass PHP 5.6 das so nicht akzeptieren wird...

ziegenberg:
> Thanks! Could you also add a test for the new method? (Copying and modifying the test for the other method should be fine.)

ziegenberg:
Sorry, clicked the wrong buttons...

oliverklee:
This should go into `main` first. Would you be willing to create a new PR targeting `main`? (If I switch the target of this PR, I fear that this would create a bunch of merge conflicts.)

JakeQZ:
`consumeWhiteSpace()` was changed in #1477 to return the whitespace consumed rather than the comments, with an optional `$comments` parameter to be filled in with the comments, akin to the other `ParserState` methods.

`ParserState` has been marked `@internal` since v8.7 and is no longer part of the public API.

If there becomes a need somewhere to consume whitespace but not comments, a method can be implemented, but I'm not currently aware of any need.

PR #671, opened 2024-08-25
[BUGFIX] Fix comment parsing to support multiple comments (backported)

### LINKED ISSUE #173: bug with parsing multiple comments

ziegenberg:
Because of an eager consumption of whitespace, the rule parsing would swallow a trailing comment, meaning the comment for the next rule would be affected. This patch addresses this by only consuming real whitespace without comments after a rule.

Fixes #173

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,8 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- Fix comment parsing to support multiple comments (#671)
+
 ## 8.6.0
 
 ### Added
--- a/src/Rule/Rule.php
+++ b/src/Rule/Rule.php
@@ -107,7 +107,12 @@ class Rule implements Renderable, Commentable
         while ($oParserState->comes(';')) {
             $oParserState->consume(';');
         }
-        $oParserState->consumeWhiteSpace();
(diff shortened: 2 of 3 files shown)


oliverklee (approved):
Thanks!

oliverklee:
I'm a bit confused … do I need to forward-port this, or is this not necessary due to other upcoming changes in `main`?

(In general, if we want to create a PR that we are also planning to backport, we first create the PR for `main` and then backport it to the maintenance branch after it has been merged.)

oliverklee:
I've now forward-ported this in #672. For future patches, please let's create the change for `main` first, and I (or another maintainer) will then take care of the backport. (This also works better with the branch protection in place, where we need another reviewer if we want to merge to `main` and would need to override this protection if we want to do a forward-port without an additional review.)

ziegenberg:
I wasn't done with the patches. As the solution for v8.x looks very different from the proposed solution for v9.0, I was about to create different patches. Either back-porting or forward-porting would have meant a completely different approach.

oliverklee:
All right. I’m looking forward to your PRs! 🙏

oliverklee:
And if you think talking things through would be helpful, please feel free to chat me up.

Issue #668, opened 2024-08-25
Make many things `@internal`


oliverklee:
We want to able to refactor internal things without being afraid of breaking things.

Probably all within `Parsing`.

PR #674, opened 2024-08-26
[TASK] Mark parsing-internal classes and methods as `@internal`

oliverklee:
Code that uses this library is not expected to call internal parsing functionality. Communicate this with the corresponding `@internal` annotation.

This allows us to boldly refactor the parser code.

Part of #668

### DIFF
--- a/src/Parsing/OutputException.php
+++ b/src/Parsing/OutputException.php
@@ -6,8 +6,6 @@ namespace Sabberworm\CSS\Parsing;
 
 /**
  * Thrown if the CSS parser attempts to print something invalid.
- *
- * @internal
  */
 class OutputException extends SourceException
 {
--- a/src/Parsing/OutputException.php
+++ b/src/Parsing/OutputException.php
@@ -6,6 +6,8 @@ namespace Sabberworm\CSS\Parsing;
 
 /**
  * Thrown if the CSS parser attempts to print something invalid.
+ *
+ * @internal
  */
 class OutputException extends SourceException
 {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
+- Mark parsing-internal classes and methods as `@internal` (#674)
 - Block installations on unsupported higher PHP versions (#691)
 - Improve performance of `Value::parseValue` with many delimiters by refactoring
   to remove `array_search()` (#413)
--- a/src/Parsing/Anchor.php
+++ b/src/Parsing/Anchor.php
@@ -4,6 +4,9 @@ declare(strict_types=1);
 
 namespace Sabberworm\CSS\Parsing;
 
+/**
+ * @internal
+ */
 class Anchor
 {
     /**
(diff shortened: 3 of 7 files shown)


src/Parser.php
```diff
@@ -33,6 +33,8 @@ public function __construct($sText, ?Settings $oParserSettings = null, $iLineNo
      * Sets the charset to be used if the CSS does not contain an `@charset` declaration.
      *
      * @param string $sCharset
+     *
+     * @internal
      */
     public function setCharset($sCharset): void
```
JakeQZ, line 39: The description suggests that this is not an internal method.  However, the example in the readme shows setting the charset via `Settings::withDefaultCharset()`.  But searching the code base, I find this is not actually called internally (internally we set the charset directly on the `ParserState` object).

So if we don't want to expose this method moving forwards, we should mark it as deprecated.  In which case, we want the logger inferface implementation before we can do so, don't we?

src/Parsing/OutputException.php
```diff
@@ -4,6 +4,8 @@
 
 /**
  * Thrown if the CSS parser attempts to print something invalid.
+ *
+ * @internal
  */
 class OutputException extends SourceException
```
JakeQZ, line 10: This is thrown on attempting to render a declaration block whose selector(s) have all been removed.  There's a test specifically for that situation and behaviour (with or without the 'throw exceptions' setting), so this doesn't look internal to me.

The same applies to the entire exception hierarchy.  Though I don't see a reason not to make them internal moving forwards..

JakeQZ (changes requested):
`Anchor` and `ParserState` both look internal to me, so marking those as internal seems fine.

I'm not sure if the exception class hierarchy should be internal.

The `Parser::get/setCharset` methods are part of the current interface (and not actually used internally).  If we want to remove them (I'm not sure), we should mark them as deprecated.

@sabberworm, what are your thoughts on these?

I thought the plan to mark methods as deprecated required the logger interface to be in place first, so that users can be informed if they are calling deprecated methods.  So if we do want to deprecate the charset get/set methods, that part of this PR would have to be put on hold.

oliverklee:
> So if we don't want to expose this method moving forwards, we should mark it as deprecated. In which case, we want the logger inferface implementation before we can do so, don't we?

No, I don't think not having the logger should be a blocker for deprecating things.

oliverklee:
> The `Parser::get/setCharset` methods are part of the current interface (and not actually used internally). If we want to remove them (I'm not sure), we should mark them as deprecated.

I have now created #688 to deprecate them.

src/Parser.php
oliverklee: They are now deprecated (#688) instead.

src/Parsing/OutputException.php
oliverklee: Done.

oliverklee:
Updated and repushed.

JakeQZ:
This is still marked as 'Draft'.

oliverklee:
> This is still marked as 'Draft'.

Oops, thanks - not anymore! :wink:

JakeQZ (approved):

PR #675, opened 2024-08-26
[FEATURE] Add PHP 8.4 to the CI matrix

oliverklee:


oliverklee:
Note: This is the backport of #657.

JakeQZ (approved):
I'm sure we've covered this before.  I'm just wondering why we can't run the unit tests against PHP 8.x on this branch.

For PHP 5.6 we may need an older version of PHPUnit, and for PHP 8 a newer one, but shouldn't Composer be able to sort that out, as long as our usage of PHPUnit is compatible with the various versions of it?

oliverklee:
> I'm just wondering why we can't run the unit tests against PHP 8.x on this branch.

I'm currently trying this in #682. Let's see how this works out …

PR #677, opened 2024-08-26
[BUGFIX] Ensure PHP 8.4 compatibility for nullable parameters

### LINKED ISSUE #634: Add a PHP-CS-Fixer rule to avoid deprecation warnings in PHP 8.4
https://cs.symfony.com/doc/rules/function_notation/nullable_type_declaration_for_default_null_value.html

oliverklee:
Ensure that parameters that have a `null` default value also have `null` as part of their type annotations.

Also remove native type declarations for nullable parameters (as nullable parameters are not possible with some of the PHP versions we currently support).

https://php.watch/versions/8.4/implicitly-marking-parameter-type-nullable-deprecated

Fixes #634


oliverklee:
This is the backport of #677.

oliverklee:
Another note: This block PHP 8.4 support for Emogrifier: https://github.com/MyIntervals/emogrifier/pull/1278

src/CSSList/AtRuleBlockList.php
```diff
@@ -57,9 +57,11 @@ public function __toString()
     }
 
     /**
+     * @param OutputFormat $oOutputFormat
+     *
      * @return string
      */
-    public function render(OutputFormat $oOutputFormat)
+    public function render($oOutputFormat)
```
JakeQZ, line 64: This doesn't look right.  It seems to be reverting a previous change.

JakeQZ (changes requested):
I think something has gone awry in the merge.

The original change affected only 4 files.  This affects 22.  It looks like you have omitted a `git pull`.

src/CSSList/AtRuleBlockList.php
oliverklee: Yes, it's removing the type declaration.

The reason for this that there are two possible ways to fix the implicit nullable parameters:

1. make the parameters explicitly nullable using a native type declaration (which is not possible with PHP 5.6, which we support in this branch)
2. drop the native type declaration for this parameter and use PHPDoc type annotations instead

I'll create a separate pre-PR to do 2. first (without changing the type).

JakeQZ: There is another way, which is to make a 9.0 release with where we're at.  The only breaking change I'm aware of is requiring PHP >= 7.2.

oliverklee:
According to #701, this change does not seem to be necessary after all. Closing.

PR #678, opened 2024-08-26
[TASK] Avoid Hungarian notation in a testcase

oliverklee:


JakeQZ (approved):
Super.  Good nomenclature.  Much easiser to understand now.  Not so much because of the Hungarian notation, but because of the abbreviations ('doc') and imprecise naming ('declaration').  Longer variable names may take the parser a few nanoseconds more, but will save many orders of magnitude more than that for humans :)

PR #682, opened 2024-08-27
[TASK] Also allow higher versions of PHPUnit

oliverklee:
This hopefully will allow us to run the unit tests for all PHP version that we claim to support, i.e., from 5.6.x to currently 8.4.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,7 +27,7 @@
         "ext-iconv": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.27"
+        "phpunit/phpunit": "^5.7.27 || ^8.5.39"
     },
     "suggest": {
         "ext-mbstring": "for parsing UTF-8 CSS"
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -27,7 +27,7 @@
         "ext-iconv": "*"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7.27"
+        "phpunit/phpunit": "^5.7.27 || ^8.5.39"
```
JakeQZ, line 30: Is there any reason not to allow 6.x or 7.x?  Or maybe there is no reason to allow them.

JakeQZ (commented):
Is fine, but just one question.

composer.json
oliverklee: I see how this can be confusing to the reader. I've now added the major versions between 5 and 8 as well. Fingers crossed that this works out well …

oliverklee: Yay, it works!

JakeQZ (approved):

PR #683, opened 2024-08-27
[BUGFIX] Avoid using `setUp()` in testcases

oliverklee:
If we want to work both with PHPUnit 5.x and 8.x, we cannot use PHPUnit's `setUp()` method: This method would need to have a `: void` return type declaration for PHPUnit 8.x, which is not possible with PHP 5.6 (the lowest version we support).

### DIFF
--- a/tests/CSSList/DocumentTest.php
+++ b/tests/CSSList/DocumentTest.php
@@ -18,7 +18,7 @@ final class DocumentTest extends TestCase
      */
     private $subject;
 
-    protected function setUp()
+    private function setUpTestcase()
     {
         $this->subject = new Document();
     }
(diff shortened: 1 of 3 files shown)


JakeQZ (commented):
I'm not keen on this.  I'd rather we make a 9.0 release which supports PHP 7.2 through 8.4.  If users are wanting to use PHP 8.4, then they're probably willing to make whatever changes are necessary to resolve any breaking changes.

JakeQZ:
> I'm not keen on this. I'd rather we make a 9.0 release which supports PHP 7.2 through 8.4. If users are wanting to use PHP 8.4, then they're probably willing to make whatever changes are necessary to resolve any breaking changes.

Sorry, I've just realized these changes are on the branch, not mainline.  In that case it's fine, and we don't need to rush 9.0.

JakeQZ (approved):

PR #686, opened 2024-08-27
[TASK] Be less strict about the `@covers` annotation

oliverklee:
We still want to require the `@covers` annotation to be used for testcase (`forceCoversAnnotation="true"`), but we also want to be able to indirectly execute code that is not referenced via an `@covers` annotation.


JakeQZ (approved):
https://github.com/MyIntervals/emogrifier/pull/1288

PR #688, opened 2024-08-27
[TASK] Deprecate `Parser::setCharset()` and `Parser::getCharset()`

### LINKED ISSUE #681: Deprecate `Parser::setCharset/getCharset`
Deprecate it in 8.x and remove it in 9.0.

oliverklee:
Fixes #681

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Deprecated
 
+- Deprecate `Parser::setCharset()` and `Parser::getCharset()` (#688) 
 - Deprecate `DeclarationBlock::createBorderShorthand()` (#578)
 - Deprecate `DeclarationBlock::createFontShorthand()` (#580)
 - Deprecate `DeclarationBlock::createDimensionsShorthand()` (#579)
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -35,6 +35,8 @@ class Parser
      * Sets the charset to be used if the CSS does not contain an `@charset` declaration.
      *
      * @param string $sCharset
+     *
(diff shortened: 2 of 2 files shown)


JakeQZ (commented):
The jury's still out.  I don't honestly know if these are useful methods, and thus whether they should be retained or not.

@sabberworm WDYT?

Not sure the deprecation comment should reference an issue number, but maybe it should.

sabberworm:
@JakeQZ I don’t know the current state of affairs regarding charset support. IIRC, the original idea was to accept lots of different charsets and convert them all to UTF-8 for parsing and but still use it when stringifying. Since the world has pretty much settled on UTF-8, I don’t think this is at all relevant anymore. Only handling UTF-8 would also rid us of the need to use the mbstring functions.

oliverklee:
Some notes on these two methods, and why I think we can deprecate (and then remove) them:

- Neither method is used internally.
- `getCharset` is missing the `return` statement and hence cannot work.
- The recommended way to set the charset (as documented in the README) is to provide the parser with a corresponding `Settings` instance, but not by calling `Parser::setCharset()`.

JakeQZ:
OK.  It seems we're in agreement that these can be deprecated.

Regarding parsing different charsets, I think we should continue to support `@charset` as long as it's part of the CSS specification.  When stringifying, we can render UTF-8.  However, this approach won't allow us to stop using the mbstring functions until `@charset` is deprecated in the CSS spec.

JakeQZ (approved):

sabberworm:
> However, this approach won't allow us to stop using the mbstring functions until `@charset` is deprecated in the CSS spec.

I think we could still drop mbstring, if we use `iconv` to convert to UTF-8 before parsing.

oliverklee:
@sabberworm Sounds good! I've added #706 for this.

PR #691, opened 2024-08-28
[TASK] Block installations on unsupported higher PHP versions

oliverklee:
We want Composer to be able to figure out which versions of which libraries it can install on which version of PHP.

To make this possible, we should only allow installations on PHP versions that we know to work.


JakeQZ (approved):
This is what we do in the sister project, Emogrifier.  It means we have to specifically make a release to support each new PHP version.  But that also involves updating the automated tests to include testing against the new PHP.  Thus we can be confident that the software works with it, having made any changes needed, as opposed to finding spurious bugs in the wild.

PR #693, opened 2024-08-28

oliverklee:
This is for the maintenance branch only.


oliverklee:
Arr … we'll need to update the GPG keys in the CI file. Then this is not worth the hassle to me. Closing.

PR #698, opened 2024-08-29
[TASK] Drop the highest/lowest Composer version switch

oliverklee:
As we don't have any direct production dependencies, it does not make sense to test both with highest and lowest Composer versions on CI.

We accidentally added this highest/lowest switch with the new coverage job. This commit removes this again, allowing for sharing the Composer caches with the main CI job.

### DIFF
--- a/.github/workflows/codecoverage.yml
+++ b/.github/workflows/codecoverage.yml
@@ -19,8 +19,6 @@ jobs:
             matrix:
                 php-version:
                     - '7.4'
-                dependencies:
-                    - highest
 
         steps:
             - name: Checkout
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
We were only using `dependencies: highest` in the code coverage matrix.  But the CI workflow does not have that dimension to the matrix (the axes of which are used for the cache key).

JakeQZ:
We may actually need the `composer install` step as well as `composer update` - see https://github.com/MyIntervals/emogrifier/pull/924.  I think we had some CI builds randomly failing.  However, if it's still a problem (it may have been fixed by GitHub), it would also be needed for the CI build, which doesn't have it.  Anyway, if we do get a random build failure at `composer update`, we'll know what to do.

PR #704, opened 2024-09-03
[TASK] Avoid Hungarian notation in a class

oliverklee:
This change does not include any breaking changes.

### DIFF
--- a/src/Comment/Comment.php
+++ b/src/Comment/Comment.php
@@ -12,21 +12,21 @@ class Comment implements Renderable
     /**
      * @var int
      */
-    protected $iLineNo;
+    protected $lineNumber;
 
     /**
      * @var string
      */
-    protected $sComment;
+    protected $commentText;
 
     /**
-     * @param string $sComment
-     * @param int $iLineNo
+     * @param string $commentText
+     * @param int $lineNumber
      */
-    public function __construct($sComment = '', $iLineNo = 0)
+    public function __construct($commentText = '', $lineNumber = 0)
     {
-        $this->sComment = $sComment;
-        $this->iLineNo = $iLineNo;
+        $this->commentText = $commentText;
+        $this->lineNumber = $lineNumber;
     }
 
     /**
@@ -34,7 +34,7 @@ class Comment implements Renderable
      */
     public function getComment()
     {
-        return $this->sComment;
+        return $this->commentText;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


JakeQZ (commented):
According to the deprecation policy discussion - https://github.com/MyIntervals/PHP-CSS-Parser/discussions/453#discussioncomment-8419934 - we should provide magic methods to access the new property names via the old names, marking the old names deprecated.

JakeQZ:
> we should provide magic methods to access the new property names via the old names, marking the old names deprecated.

The [PublicPropertyDeprecationTrait.php](https://review.typo3.org/c/Packages/TYPO3.CMS/+/52928/34/typo3/sysext/core/Classes/Compatibility/PublicPropertyDeprecationTrait.php) class from TYPO3 along with associated [TestCase](https://review.typo3.org/c/Packages/TYPO3.CMS/+/52928/34/typo3/sysext/core/Tests/Unit_Deprecated/Compatibility/PublicPropertyDeprecationTraitTest.php) look ideal for the job, assuming the licence is compatible.

JakeQZ:
> This change does not include any breaking changes.

The renamed properties are `protected`, so this would break any subclass accessing them directly.

JakeQZ:
> The renamed properties are `protected`, so this would break any subclass accessing them directly.

I think we should also make them `private` as part of this change.  (But do still need to deprecate them first.)

> The [PublicPropertyDeprecationTrait](https://review.typo3.org/c/Packages/TYPO3.CMS/+/52928/34/typo3/sysext/core/Classes/Compatibility/PublicPropertyDeprecationTrait.php) class from TYPO3 along with associated [TestCase](https://review.typo3.org/c/Packages/TYPO3.CMS/+/52928/34/typo3/sysext/core/Tests/Unit_Deprecated/Compatibility/PublicPropertyDeprecationTraitTest.php) look ideal for the job

I think I've spotted a mistake in the class: `__isset()` should `return isset($this->$propertyName)` in all branches.  (The other magic methods do the equivalent, and look correct.)

JakeQZ:
> The [PublicPropertyDeprecationTrait](https://review.typo3.org/c/Packages/TYPO3.CMS/+/52928/34/typo3/sysext/core/Classes/Compatibility/PublicPropertyDeprecationTrait.php) class from TYPO3 along with associated [TestCase](https://review.typo3.org/c/Packages/TYPO3.CMS/+/52928/34/typo3/sysext/core/Tests/Unit_Deprecated/Compatibility/PublicPropertyDeprecationTraitTest.php) look ideal for the job

Looking again, this only provides for deprecation without renaming.  To support renaming and deprecation at the same time, there'd also need to be a map from old property names to new ones.

oliverklee:
Marking as draft for now - I think I'd like to change our policy for public APIs.

oliverklee:
With our new API and deprecation policy in place, I think this is unblocked now.

JakeQZ (approved):
> With our new API and deprecation policy in place, I think this is unblocked now.

Yes, the API policy says that classes should not be extended, so `protected` methods are effectively internal.

PR #707, opened 2024-09-05
[TASK] Use fixed versions of the development dependencies

oliverklee:
We don't want to have sudden build failures when a new version of a development dependency gets released.

Instead, we'll keep having automatic Dependabot updates for our dependencies that will allow us to see the effects of each update before switchting to a new version of a dependency.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -27,12 +27,12 @@
         "ext-iconv": "*"
     },
     "require-dev": {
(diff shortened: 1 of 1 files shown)


JakeQZ (approved):
Yes, certainly updates to static analysis tools may detect additional issues, resulting in CI check failures.

Is Dependabot configured to check for updates to the tools installed with Composer, or does this still need to be done?

PR #711, opened 2024-09-05
[TASK] Mark parsing-internal classes and methods as `@internal`

oliverklee:
Code that uses this library is not expected to call internal parsing functionality. Communicate this with the corresponding `@internal` annotation.

This allows us to boldly refactor the parser code.

This is the backport of #674.

Part of #668


JakeQZ:
Forgot to click 'Approve' before merge.  But did check the changes.

PR #713, opened 2024-09-05
[TASK] Set the target milestone for the Dependabot updates

oliverklee:


JakeQZ (approved):
I checked this is the correct milestone ID.  It will need to be updated when that milestone is complete, probably as part of the steps to make a new release.

PR #714, opened 2024-09-08
[TASK] Mark the shorthand expansion as to be removed earlier

oliverklee:
This will allow us to get rid of the code earlier, which will reduce code complexity and make maintenance easier.

I will backport those deprecations to 8.x.


JakeQZ (approved):
This is fine given there will be an 8.7 release.

PR #720, opened 2024-09-14
[DOCS] Add an API and deprecation policy

oliverklee:
As discussed here:
https://github.com/MyIntervals/PHP-CSS-Parser/discussions/715

### DIFF
--- a/docs/API-and-deprecation-policy.md
+++ b/docs/API-and-deprecation-policy.md
@@ -9,8 +9,8 @@ and your code might break with any new release of this library.
 Any classes, methods and properties that are `public` and not marked as
 `@internal` are considered to be part of the API. Those methods will continue
 working in a compatible way over minor and bug-fix releases according
-to [Semantic Versioning](https://semver.org/). though we might change the native
-type declarations in a way that might could break subclasses, though.
+to [Semantic Versioning](https://semver.org/), though we might change the native
+type declarations in a way that could break subclasses.
 
 Any classes, methods and properties that are `protected` or `private` are _not_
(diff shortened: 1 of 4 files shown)


docs/API-and-deprecation-policy.md
```diff
@@ -0,0 +1,50 @@
...
+
+The code in this library is intended to be called by other projects. It is not
+intended to be extended. If you want to extend any classes, you're on your own,
+and your code might break with any new release of this library.
+
+Any classes, methods and properties that are `public` and not marked as
+`@internal` are considered to be part of the API. Those methods will continue
+working in a compatible way over minor and bug-fix releases according
+to [Semantic Versioning](https://semver.org/). though we might change the native
+type declarations in a way that might could break subclasses, though.
```
JakeQZ, line 13:
```suggestion
to [Semantic Versioning](https://semver.org/), though we might change the native
type declarations in a way that could break subclasses.
```

JakeQZ (changes requested):
Just some typos in one sentence.

It occurs to me that we ourselves need to track when we mark something as `@internal` to ensure we don't make make any breaking changes before the next major release since doing so.  I don't know if we can use something like "`@internal` may be changed or removed without notice from version 9.0.0" in the PHPDoc.  This doesn't affect this PR, I'm just concerned that we might for example start making breaking changes to a class marked as `@internal` for 8.7 for a bugfix which is then included in 8.7.1, not realizing (or forgetting) it had not been marked `@internal` in 8.6.

JakeQZ (approved):

PR #725, opened 2024-09-20
[TASK] Update PHPUnit

oliverklee:


JakeQZ (approved):
Do we want to require specific versions of PHPUnit, as we do in Emogrifier, in case an update breaks the build?  (Though that's beyond the scope of this PR.)

oliverklee:
We indeed do! PR incoming …

PR #726, opened 2024-09-20
[TASK] Use fixed PHPUnit versions

oliverklee:
This avoids build breakage when a new version of PHPUnit gets released.


JakeQZ (approved):
The equivalent change has already been made to `main`.

Issue #729, opened 2024-10-01
Add a GitHub Action that comments on pull requests if tests are missing


oliverklee:
https://github.com/infection/tests-checker-action

hemanthhhhhh:
Please assign this issue to me

oliverklee:
Hi @hemanthhhhhh, @ghostwriter already has taken this issue. Unfortunately, GitHub only allows assigning issues to project members and not to outside contributors. Sorry. :-(

oliverklee:
Fixed by #730.

PR #730, opened 2024-10-01
Add GitHub Action to Require Tests in Pull Requests

ghostwriter:
This pull request addresses issue [#729](https://github.com/MyIntervals/PHP-CSS-Parser/issues/729) by adding the  [infection/tests-checker-action](https://github.com/infection/tests-checker-action) GitHub Action that ensures tests are included in pull requests.

Changes Introduced:

- `.github/workflows/require-tests.yml`: Added a workflow named 'Require Tests' that uses the `infection/tests-checker-action` to verify the presence of test files in the `tests` directory with a specific pattern (`*Test.php`).

### DIFF
--- a/.github/workflows/require-tests.yml
+++ b/.github/workflows/require-tests.yml
@@ -6,7 +6,7 @@ on:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
     steps:
       - uses: infection/tests-checker-action@v1.0.2
         with:
--- a/.github/workflows/require-tests.yml
+++ b/.github/workflows/require-tests.yml
@@ -1,4 +1,4 @@
-name: 'Require Tests'
+name: 'Require Tests on Code Change'
 
 on:
   pull_request_target:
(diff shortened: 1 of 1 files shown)


.github/workflows/require-tests.yml
```diff
@@ -0,0 +1,17 @@
+name: 'Require Tests'
+
+on:
+  pull_request_target:
+    types: [opened]
+
+jobs:
+  check:
+    runs-on: ubuntu-latest
```
oliverklee, line 9: To be consistent with the other workflows, please use a specific Ubuntu version. This avoids sudden breakage when a new version becomes `ubuntu-latest`.

```suggestion
    runs-on: ubuntu-22.04
```

.github/workflows/require-tests.yml
```diff
@@ -0,0 +1,17 @@
...
+    steps:
```
oliverklee, line 10: For consistency with the other workflows, please add a blank line above this line, and also please provide a `name` for this job.

oliverklee (changes requested):
Thanks for contributing, @ghostwriter! :pray: Looks generally good, but still needs a bit of polish. I've added some review comments in this regard.

oliverklee (approved):
Looks good to me. Thank you, @ghostwriter! :pray:

PR #738, opened 2024-10-20
[TASK] Add some more tests for parsing comments

oliverklee:
Also add a skipped test for the broken behavior that currently
is blocking Emogrifier.


JakeQZ (approved):
This might be a case for using a 'cross' data provider to generate the combinations of CSS with/without strict parsing.  But we plan to remove the strict parsing setting anyway, and aren't currently using the package that facilitates 'cross' providers.

JakeQZ:
> This might be a case for using a 'cross' data provider to generate the combinations of CSS with/without strict parsing.

Oh, I've just spotted that the strict mode test is skipped.

PR #740, opened 2024-10-23
[BUGFIX] Revert broken support for multiple comments

oliverklee:
This reverts e4c66f62c5df3f53f48d82cb30b20b3007af7574 (#672), which broke comment parsing in strict mode.

We'll need to re-implement support for multiple comments later in a way that does not break the existing comment parsing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,8 @@ Please also have a look at our
 
 ### Fixed
 
+- Revert broken support for multiple comments (#740)
 - Fix type errors in PHP strict mode (#664)
-- Fix comment parsing to support multiple comments (#672)
 - Fix undefined local variable in `CalcFunction::parse()` (#593)
 - Fix PHP notice caused by parsing invalid color values having less than 6
   characters (#485)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,6 @@ Please also have a look at our
 
 ### Fixed
 
-- Revert broken support for multiple comments (#740)
 - Fix type errors in PHP strict mode (#664)
 - Fix undefined local variable in `CalcFunction::parse()` (#593)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -42,8 +42,8 @@ Please also have a look at our
 
 ### Fixed
 
+- Revert broken support for multiple comments (#740)
```
JakeQZ, line 45: The incorrect fix hasn't been included in a release yet (it wasn't in 8.6.0), so I don't think we need a changelog entry to say it's been taken out - just remove the one that was added for it.

tests/ParserTest.php
```diff
@@ -1171,22 +1171,6 @@ public function flatCommentExtractingOneComment(): void
...
-        $document = $parser->parse();
-        $contents = $document->getContents();
-        $divRules = $contents[0]->getRules();
-        $rule1Comments = $divRules[0]->getComments();
-        $rule2Comments = $divRules[1]->getComments();
-        self::assertCount(1, $rule1Comments);
-        self::assertCount(1, $rule2Comments);
-        self::assertEquals('Find Me!', $rule1Comments[0]->getComment());
-        self::assertEquals('Find Me Too!', $rule2Comments[0]->getComment());
-    }
```
JakeQZ, line 1189: It is worth keeping this test, but have it skipped until the correct fix is implemented?

JakeQZ (commented):
Good find.  I have a couple of queries re reverting the change.

CHANGELOG.md
oliverklee: Done.

tests/ParserTest.php
oliverklee: Done.

JakeQZ (approved):

PR #743, opened 2024-10-25
[TASK] Configure PHPUnit to be more strict

oliverklee:
This change brings the PHPUnit configuration in line with what we have at Emogrifier (except for the difference in the XSD version number).

### DIFF
--- a/phpunit.xml
+++ b/phpunit.xml
@@ -1,12 +1,14 @@
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-         xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/8.5/phpunit.xsd"
-         beStrictAboutChangesToGlobalState="true"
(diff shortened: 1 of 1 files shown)


JakeQZ (changes requested):
Changes seem mostly fine, but could we please keep the original indentation style.  This would make the diff cleaner, and avoid attributes being idented by 9 spaces.

Also, do we need `bootstrap="vendor/autoload.php"`.  It seems to work without it, and must presumably be using that by default, otherwise the classes to be tested wouldn't load.  Explicitly stating it creates a hard-coded value that makes assumptions about the behaviour of a third-party package (Composer itself).  If PHPUnit effectively deals with the location of Composer's autoload file, then we should just allow it to do so.

oliverklee:
Done and repushed.

JakeQZ (approved):
I see clearly now we already had `cacheResult="false"` in this project :)

PR #746, opened 2024-10-25
[BUGFIX] Avoid implicit nullable parameters

oliverklee:
This avoids deprecation warnings in PHP 8.4.

Unfortunately, we need to remove the native type declaration for this to keep supporting PHP down to version 5.6.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
-- Add support for PHP 8.4 (#675, #701)
+- Add support for PHP 8.4 (#675, #701, #746)
 
 ### Changed
 
--- a/src/CSSList/AtRuleBlockList.php
+++ b/src/CSSList/AtRuleBlockList.php
@@ -57,9 +57,11 @@ class AtRuleBlockList extends CSSBlockList implements AtRule
     }
 
     /**
+     * @param OutputFormat|null $oOutputFormat
+     *
      * @return string
      */
-    public function render(OutputFormat $oOutputFormat)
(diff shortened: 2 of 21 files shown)


oliverklee:
Demo with PHPUnit set to fail on deprecations: #745

JakeQZ (approved):
> Unfortunately, we need to remove the native type declaration for this to keep supporting PHP down to version 5.6.

I feared even worse: that it might not be possible to support all PHP versions, 5.6 through 8.4.

PHPDoc `@params` all present and correct where they needed to be reintroduced.

westonruter:
For what it's worth, WordPress's minimum version requirement is now PHP 7.2. So if this was what was holding back abandoning PHP 5.6 support, then this is no longer the case.

JakeQZ:
> For what it's worth, WordPress's minimum version requirement is now PHP 7.2. So if this was what was holding back abandoning PHP 5.6 support, then this is no longer the case.

Thanks.  Yes, it may have been easier to make a 9.0.0 release supporting PHP 8.4, also ditching support for PHP<7,2, with other planned or already-implemented breaking changes then moved to a 10.0.0 milestone.  WordPress wasn't a particular consideration; I think it was more about making a minor release to support PHP 8.4 if possible, so that other breaking changes already on `main` weren't included.  We could have bumped the version to 9.0 on the release branch which doesn't include other potentially breaking changes, and merged the dropping of PHP<7.2 support to that.  That may have been easier.

But by the time you've found out what the problem is, you've usually pretty-much found a fix, so there's then less work involved in carrying on that same path.  In this case, that path leads to release that supports PHP versions from 5.6 through 8.4, though is missing some type annotations (but which are still present on the `main` branch for the next major release).

PR #748, opened 2024-10-26
[CLEANUP] Autoformat the code

oliverklee:

PR #749, opened 2024-10-26
[CLEANUP] Autoformat the code

oliverklee:
This is the v8.x backport of #748.


tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -43,7 +44,8 @@ public static function provideSyntacticlyCorrectAtRule()
                         display: flex;
                     }
                 }
-            '],
+            ',
+            ],
```
JakeQZ, line 48: The leading and trailing whitespace is arbitrary, and for readability.  So the string content and closing quote should now be futher indented.

tests/Value/ValueTest.php
```diff
@@ -39,7 +39,7 @@ public function parsesArithmeticInFunctions($operator)
     {
         $subject = Value::parseValue(new ParserState('max(300px, 50vh ' . $operator . ' 10px);', Settings::create()));
 
-        self::assertSame('max(300px,50vh ' . $operator . ' 10px)', (string) $subject);
+        self::assertSame('max(300px,50vh ' . $operator . ' 10px)', (string)$subject);
```
JakeQZ, line 42: This surely contradicts PHP-Fixer, which added spaces after casts.

JakeQZ (changes requested):
I'm sure PHP-Fixer added the spaces after cast operators (somewhat against my taste).  So we shouldn't be removing them again.

There's a string with arbitrary whitespace that needs manual post-reformatting to make it readable again.

Shouldn't the autoformat be done first on `main`?

oliverklee:
> Shouldn't the autoformat be done first on `main`?

Yes: #748 

I'll mark this PR here as draft for now and do the corresponding changes there.

tests/Value/ValueTest.php
oliverklee: Done.

tests/CSSList/AtRuleBlockListTest.php
oliverklee: Done.

JakeQZ (approved):

PR #754, opened 2024-10-30
[TASK] Create directory structure for making fine-grained tests

oliverklee:
We'll rework and sort the tests into these folders:

- `Unit` and `Functional`: for tests that do not call deprecated methods, and where PHPUnit should warn if deprecated methods are called
- `UnitDeprecated` and `FunctionalDeprecated`: for tests that call deprecated methods, and where PHPUnit should not warn about those

For the time being, we still call all tests with the same configuration.


JakeQZ (approved):
Seems reasonable.  I presume the `.gitkeep` files will be removed as soon as there's something in these directories.

oliverklee:
>  I presume the `.gitkeep` files will be removed as soon as there's something in these directories.

Yes, exactly.

PR #759, opened 2024-11-03
[TASK] Add a `fix` Composer script to run all code fixers

oliverklee:
This makes our life as developers a bit more convenient.

Also run Rector before the style fixer as Rector is responsible for structural changes that need to be style-fixed after that.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -85,9 +85,12 @@
         "ci:tests:coverage": "phpunit --do-not-cache-result --coverage-clover=coverage.xml",
         "ci:tests:sof": "phpunit --stop-on-failure --do-not-cache-result",
         "ci:tests:unit": "phpunit --do-not-cache-result",
+        "fix": [
+            "@fix:php"
+        ],
         "fix:php": [
-            "@fix:php:fixer",
-            "@fix:php:rector"
+            "@fix:php:rector",
+            "@fix:php:fixer"
         ],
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -85,9 +85,12 @@
...
         "ci:tests:sof": "phpunit --stop-on-failure --do-not-cache-result",
         "ci:tests:unit": "phpunit --do-not-cache-result",
+        "fix": [
+            "@fix:php"
+        ],
         "fix:php": [
-            "@fix:php:fixer",
-            "@fix:php:rector"
+            "@fix:php:rector",
+            "@fix:php:fixer"
```
JakeQZ, line 93: F comes before R in the latin alphabet, so the existing ordering was correct.

JakeQZ (changes requested):
Looks like we don't yet have Composer-Normalize on this project, but the previous ordering within `fix:php` I think was correct, and should not be changed.

composer.json
oliverklee: I changed this on purpose:

> Also run Rector before the style fixer as Rector is responsible for structural changes that need to be style-fixed after that.

JakeQZ: > I changed this on purpose:
> 
> > Also run Rector before the style fixer as Rector is responsible for structural changes that need to be style-fixed after that.

Sorry, I missed that, and for some reason thought the script content itself would be reordered by Composer-Normalize :/

JakeQZ (approved):

PR #765, opened 2024-11-07
[FEATURE] Add composer-normalize to our development toolchain

oliverklee:


.github/workflows/ci.yml
```diff
@@ -139,7 +140,7 @@ jobs:
 
             - name: Install development tools
               run: |
-                  phive --no-progress install --trust-gpg-keys BBAB5DF0A0D6672989CF1869E82B2FB314E9906E
+                  phive --no-progress install --trust-gpg-keys 0FDE18AE1D09E19F60F6B1CBC00543248C87FB13,BBAB5DF0A0D6672989CF1869E82B2FB314E9906E
```
JakeQZ, line 143: I'm deducing that the keys are in this order to be consistent with the order in `phars.xml` so as to make maintainance less of a headache (and just happen to also be in the order of the tools in the matrix).

I notice that the equivalent section in Emogrifier has 4 GPG keys, but there are now only 3 tools installed with PHIVE - maybe when a tool was removed (or moved back to Composer) we omitted to remove the GPG key from the list.

JakeQZ (approved):
Looks complete, with implementation consistent with the same in Emogrifier.

JakeQZ:
I guess the check is failing because we needed to normilize `composer.json` first.

JakeQZ:
> I guess the check is failing because we needed to normilize `composer.json` first.

With a rebase to include #766 this can now be merged.

.github/workflows/ci.yml
oliverklee: I have created a ticket for that: https://github.com/MyIntervals/emogrifier/issues/1367

PR #766, opened 2024-11-07
[CLEANUP] Normalize the `composer.json` with `composer-normalize`

oliverklee:


JakeQZ (approved):
Re https://github.com/MyIntervals/PHP-CSS-Parser/pull/765#issuecomment-2463071265

Thought so.

PR #767, opened 2024-11-07
[DOCS] Add a section on testing to `CONTRIBUTING.md`

oliverklee:
This change mirrors exactly what we already have at our sister project.


JakeQZ (approved):
> This change mirrors exactly what we already have at our sister project.

Ineed it does.

PR #770, opened 2024-11-10
[TASK] Configure a git commit message prefix for Dependabot

oliverklee:


JakeQZ (approved):
I'm happy with the prefix, but note we currently use uppercase in the other prefixes (though I don't see a great need to).

We'll see if it works as desired the next time there's a bugfix update to a tool.

PR #771, opened 2024-11-10
[TASK] Add more Rector rules

oliverklee:
This change is almost identical to the [corresponding change for Emogrifier](https://github.com/MyIntervals/emogrifier/pull/1372). The only differences are that we're using PHPUnit 8 here, not PHPUnit 9, and PHP 7.2, not 7.3.


JakeQZ (approved):
LGTM.

PR #772, opened 2024-11-10
[TASK] Clean up the code with Rector

oliverklee:
Just the Rector changes, and some redundancy removal related to those, but nothing more.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -167,13 +167,12 @@ abstract class CSSList implements Renderable, Commentable
             $sMediaQuery = null;
             if (!$oParserState->comes(';')) {
                 $sMediaQuery = \trim($oParserState->consumeUntil([';', ParserState::EOF]));
+                if ($sMediaQuery === '') {
+                    $sMediaQuery = null;
+                }
             }
             $oParserState->consumeUntil([';', ParserState::EOF], true, true);
-            return new Import(
-                $oLocation,
-                $sMediaQuery !== null && $sMediaQuery !== '' && $sMediaQuery !== '0' ? $sMediaQuery : null,
-                $iIdentifierLineNum
-            );
+            return new Import($oLocation, $sMediaQuery, $iIdentifierLineNum);
         } elseif ($sIdentifier === 'charset') {
             $oCharsetString = CSSString::parse($oParserState);
             $oParserState->consumeWhiteSpace();
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -169,7 +169,11 @@ abstract class CSSList implements Renderable, Commentable
                 $sMediaQuery = \trim($oParserState->consumeUntil([';', ParserState::EOF]));
             }
             $oParserState->consumeUntil([';', ParserState::EOF], true, true);
-            return new Import($oLocation, $sMediaQuery ?: null, $iIdentifierLineNum);
+            return new Import(
+                $oLocation,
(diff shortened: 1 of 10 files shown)


src/CSSList/CSSList.php
```diff
@@ -169,7 +169,11 @@ private static function parseAtRule(ParserState $oParserState)
                 $sMediaQuery = \trim($oParserState->consumeUntil([';', ParserState::EOF]));
             }
             $oParserState->consumeUntil([';', ParserState::EOF], true, true);
-            return new Import($oLocation, $sMediaQuery ?: null, $iIdentifierLineNum);
+            return new Import(
+                $oLocation,
+                $sMediaQuery !== null && $sMediaQuery !== '' && $sMediaQuery !== '0' ? $sMediaQuery : null,
+                $iIdentifierLineNum
+            );
```
JakeQZ, line 176: I think instead we should initialize `$sMediaQuery` to `''`.  And can and should assume `'0'` to be valid (it might be one day).  (The curious behaviour of PHP casting a string containing the digit zero to false was chosen to make it easiser to handle HTTP request data and database queries, where the number zero would be received as a string.  I don't think that applies here, where it is merely a nonconsequential oversight.)

`Import` (AFAICT) just stores this as a string to be regurgitated later, without having any knowledge of its meaning.  Maybe `Import` should be modified so that this is always a string, with an empty string supplanting `null`.  That would allow for cleaner type declrations.

src/Value/CSSString.php
```diff
@@ -53,7 +53,7 @@ public static function parse(ParserState $oParserState): CSSString
         $sContent = null;
         if ($sQuote === null) {
             // Unquoted strings end in whitespace or with braces, brackets, parentheses
-            while (!\preg_match('/[\\s{}()<>\\[\\]]/isu', $oParserState->peek())) {
+            while (\in_array(\preg_match('/[\\s{}()<>\\[\\]]/isu', $oParserState->peek()), [0, false], true)) {
```
JakeQZ, line 56: This could be `\preg_match(...) !== 1`, though I'd prefer `\preg_match(...) < 1` if possible, in case it's later changed to `preg_match_all()`.

The Rector-proposed change makes the code unnecessarily unreadable, and perhaps should be reported as a bug there.

The mixed return types for the built in PHP functions do cause some headaches with static analysis tools, which was why I ended up implementing https://github.com/MyIntervals/emogrifier/pull/1319 and others, having originally only intended https://github.com/MyIntervals/emogrifier/pull/1308 to provide a better fallback for the `replace` functions.  Maybe that class can be moved to a separate library...

JakeQZ (changes requested):
Mostly good.  But two of the changes compromise code readability, and could do with looking at in more detail.

src/CSSList/CSSList.php
oliverklee: I've added a check for an empty string for the time being (which I guess would match the original intention of the code best). Let's make the types more strict in a separate change.

src/Value/CSSString.php
oliverklee: Done. (I've used `!== 1` in order to avoid potentially using `<` on `false`.)

src/CSSList/CSSList.php
JakeQZ: > I've added a check for an empty string for the time being

Neat - makes the code even cleaner.

JakeQZ (approved):
Neat.

PR #773, opened 2024-11-11
[TASK] Add a section on coding style to `CONTRIBUTING.md`

oliverklee:
This is an exact copy of what we have at our sister project.


JakeQZ (approved):
> This is an exact copy of what we have at our sister project.

Indeed it is.

PR #777, opened 2024-12-12
[TASK] Upgrade to PHPStan 2.x and Rector 2.x

oliverklee:


oliverklee:
For this change, we'll need to drop support for PHP < 7.4 first.

JakeQZ:
> For this change, we'll need to drop support for PHP < 7.4 first.

PHP 7.4 just becomes a dev requirement, not a minimum requirement for the package.  Ideally we can find a way of running the versions of the tools compatible with the version contraints.  Oh, Composer.  It's what it's there for.

composer.json
```diff
@@ -29,10 +29,10 @@
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "1.4.0",
         "phpstan/extension-installer": "1.4.3",
-        "phpstan/phpstan": "1.12.9",
-        "phpstan/phpstan-phpunit": "1.4.0",
+        "phpstan/phpstan": "2.0.3",
+        "phpstan/phpstan-phpunit": "2.0.1",
```
JakeQZ, line 33: I think we need an OR-list of versions, e.g. `"phpstan/phpstan": ".12.0 || 2.0.3"`.  That way we can have the new version installed when the PHP environment allows, whilst also continuing to support older PHP versions.

JakeQZ (commented):
An OR-list for the versions in `composer.json` would solve the dependency hell issue.  It's what it's there for.  There's no reason to drop support for PHP<7.4.

.github/dependabot.yml
```diff
@@ -16,6 +16,11 @@ updates:
           interval: "daily"
       allow:
           - dependency-type: "development"
+      ignore:
+        - dependency-name: "phpstan/*"
+        - dependency-name: "phpunit/phpunit"
+          versions: [ ">= 9.0.0" ]
```
JakeQZ, line 22: Yes, we probably don't want the bot to try a major update of PHPUnit, which would be incompatible.

JakeQZ (approved):
WFM.  We can deal with the PHPStan-reported issues at a later stage.

JakeQZ:
I removed the non-applicable commit message that was separately done with #778.

PR #780, opened 2024-12-17
[TASK] Allow `assertInstanceOf` calls in tests

oliverklee:
We find those checks useful.

### DIFF
--- a/config/phpstan.neon
+++ b/config/phpstan.neon
@@ -19,3 +19,8 @@ parameters:
     null_over_false: true
     narrow_param: true
     narrow_return: true
+
+  ignoreErrors:
+    -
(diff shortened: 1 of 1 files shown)


oliverklee:
> Looks like this just needs a baseline update to remove the `instanceof` error which is now squished by the config change...

I tried that: It turns out that we currently do not have any of those warnings (I tend to put this on the fact that PHPStan is still at level 1 for this project), and blocklisting warnings we don't have (yet) is something that PHPStan does not support.

So I'll keep this PR sitting here for the time being and un-draft it once we actually have any of those warnings to ignore.

JakeQZ:
> > Looks like this just needs a baseline update to remove the `instanceof` error which is now squished by the config change...
> 
> we currently do not have any of those warnings

Ah, I see.  I misunderstood the error message (did not read it properly), and thought it was referring to an existing error that now needed to be removed from the baseline.

PR #781, opened 2024-12-17
[TASK] Drop redundant PHPStan configuration options

oliverklee:
`scanDirectories` is not needed when `paths` is set.


JakeQZ (approved):
I noticed this was not present in the config for Emogrifier.

PR #785, opened 2024-12-31
[TASK] Drop the CI action that requires test coverage

oliverklee:
This tool has turned out to be less helpful for our project that we had hoped.


JakeQZ (approved):
See https://github.com/MyIntervals/emogrifier/discussions/1381 for rationale.

PR #790, opened 2025-01-17
[BUGFIX] Parse @font-face src property as comma-delimited list

### LINKED ISSUE #789: Multiple values in `src` property of `@font-face` rule not parsed correctly
Using v8.4.0 with PHP 8.2.20:

When parsing 
```
$cssContent = "@font-face {
    font-family: 'Open Sans';
    font-style: italic;
    font-weight: 300;
    font-display: swap;
    src: url('../fonts/open-sans-italic-300.woff2') format('woff2'),
         url('../fonts/open-sans-italic-300.ttf') format('truetype');
}";
$cssParser = new \Sabberworm\CSS\Parser($contentCss);
$cssDocument = $cssParser->parse();
$fontFaceSrcList = $cssDocument->getContents()[0]->getRulesAssoc()['src']->getValue()->getListComponents();
```
the resulting $fontFaceSrcList contains 3 instead of 2 components:
image

JakeQZ:
Fixes #789.

Also adds an initial `TestCase` for `Rule/Rule`.


tests/Rule/RuleTest.php
```diff
@@ -0,0 +1,58 @@
...
+use Sabberworm\CSS\Value\RuleValueList;
+use Sabberworm\CSS\Value\ValueList;
+
+/**
+ * @covers \Sabberworm\CSS\Rule\Rule
+ */
+final class RuleTest extends TestCase
+{
+    /**
+     * @return array<string, array{0: string, 1: array<int, string>}>
```
oliverklee, line 20:
```suggestion
     * @return array<string, array{0: string, 1: list<class-string>}>
```

See:
https://phpstan.org/writing-php-code/phpdoc-types#lists
https://phpstan.org/writing-php-code/phpdoc-types#class-string

tests/Rule/RuleTest.php
```diff
@@ -0,0 +1,58 @@
...
+                ",
+                [RuleValueList::class, RuleValueList::class],
+            ],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param array<int, string> $expectedTypeClassnames
```
oliverklee, line 38:
```suggestion
     * @param list<class-string> $expectedTypeClassnames
```

tests/Rule/RuleTest.php
```diff
@@ -0,0 +1,58 @@
...
+     */
+    public function parsesValuesIntoExpectedTypeList(string $rule, array $expectedTypeClassnames): void
+    {
+        $subject = Rule::parse(new ParserState($rule, Settings::create()));
+
+        $value = $subject->getValue();
+        self::assertInstanceOf(ValueList::class, $value);
+
+        $actualClassnames = \array_map(
+            function ($component): string {
```
oliverklee, line 50: We can make the function static and add a type declaration for `$component`.

tests/Rule/RuleTest.php
```diff
@@ -0,0 +1,58 @@
...
+                return \get_class($component);
+            },
+            $value->getListComponents()
+        );
+
+        self::assertSame($expectedTypeClassnames, $actualClassnames);
```
oliverklee, line 56: Could we maybe use [`assertContainsOnlyInstancesOf`](https://docs.phpunit.de/en/8.5/assertions.html#assertcontainsonlyinstancesof) instead and avoid the anonymous function?

src/Rule/Rule.php
```diff
@@ -117,13 +117,21 @@ public static function parse(ParserState $oParserState): Rule
      * @param string $sRule
      *
      * @return array<int, string>
+     * The first item is the innermost separator (or, put another way, the highest-precedence operator).
+     * The sequence continues to the outermost separator (or lowest-precedence operator).
```
oliverklee, line 121: I'd prefer this text to be part of the general description of this method instead of the `@return` annotation as the description is what I'd read first (as a human).

src/Rule/Rule.php
```diff
@@ -117,13 +117,21 @@ public static function parse(ParserState $oParserState): Rule
...
         if (\preg_match('/^font($|-)/', $sRule)) {
             return [',', '/', ' '];
         }
-        return [',', ' ', '/'];
+
+        switch ($sRule) {
+            case 'src':
+                return [' ', ','];
+            default:
+                return [',', ' ', '/'];
```
oliverklee, line 133: Maybe we can integrate the preg matching into the `switch` as `switch` also allows general conditions: https://www.php.net/manual/en/control-structures.switch.php

src/Rule/Rule.php
```diff
@@ -117,13 +117,21 @@ public static function parse(ParserState $oParserState): Rule
      * @param string $sRule
      *
      * @return array<int, string>
```
oliverklee, line 119: Should we make this more specific as we're touching the return value anyway?

```suggestion
     * @return list<non-empty-string>
```

oliverklee (changes requested):
Looks generally very good!

I've added some comments regarding readability.

tests/Rule/RuleTest.php
JakeQZ: Thanks.  I didn't know about `class-string`, and had wrongly assumed `list` was a more general type than `array`.  Changed as suggested.

tests/Rule/RuleTest.php
JakeQZ: Ditto.

tests/Rule/RuleTest.php
JakeQZ: Done.  On checking the type for `$component`, I found it could be a `Value` or a `string`, so have added an additional check.

tests/Rule/RuleTest.php
JakeQZ: It's true that the only test data added so far does on;y contain `RuleValueList`s.  But other properties may be expected to be parsed into different `Value`-based classes.

src/Rule/Rule.php
JakeQZ: Moved, adding context.

src/Rule/Rule.php
JakeQZ: I don't see how.  Doesn't using general conditions require opening with `switch (true)`?

oliverklee: Hm… or maybe we can create a data structure "regexp => delimiters" for this?

tests/Rule/RuleTest.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Value;
```
oliverklee, line 5: Please let's move this to `Tests/Unit/` (so I won't have to when I move/migrate the other existing tests).

src/Rule/Rule.php
JakeQZ: Done.  Also noticed that a PHP type hint for the parameter could be used, but I think that should be a separate PR (particularly as we may want to backport retaining support for a PHP version which doesn't allow argument type hints for primitive types).

JakeQZ:
> I've added some comments regarding readability.

I've made some changes and replied to those comments.

tests/Rule/RuleTest.php
JakeQZ: Done and repushed.

tests/Unit/Rule/RuleTest.php
```diff
@@ -0,0 +1,61 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Value;
```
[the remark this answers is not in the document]
JakeQZ: It lost my comment.

The namespace should also have been `Rule` not `Value` - mistake arising from copying and pasting from another file as a template.

src/Rule/Rule.php
JakeQZ: > Hm… or maybe we can create a data structure "regexp => delimiters" for this?

I'd suggest leaving this until more cases have been added, then we'd have a clearer idea about what would be best.

oliverklee (approved):

Issue #755, opened 2024-11-01
Parsing RGB with alpha is broken


masoudnkh:
i use php-css-parser but not coverage this code:
```
.wp-block-table.is-style-stripes tbody tr:nth-child(odd) {
    background-color: rgb(240 240 240 / 10%)!important;
}
```

remove unused css code and good work for all code but when i exclude ".wp-block-table.is-style-stripes" the output is empty like:
```
.wp-block-table.is-style-stripes tbody tr:nth-child(odd) {
}
```

oliverklee:
Hi @masoudnkh, thanks for reporting this issue.

I still don't quite understand the problem. Could you please post the exact CSS you parse, the code you execute, and the CSS you get back, and the CSS you'd expect to get back? Thanks!

oliverklee:
Also, could you please post which version of PHP-CSS-Parser you're using?

colethorsen:
PHP-CSS-Parser v 8.7.0

It appears as though the parser doesn't support rgb with alpha 
https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb

```
.countdown_field {
	border: 1px solid rgb(var(--text) / .35);
	width: 120px;
}

```

or

```
.countdown_field {
	border: 1px solid rgb(255 255 255 / .35);
	width: 120px;
}
```

```
$oCssParser   = new \Sabberworm\CSS\Parser($this->css);
$oCssDocument = $oCssParser->parse();

$rules = $oCssDocument->getAllRuleSets();
```

yields:

```
Sabberworm\CSS\RuleSet\DeclarationBlock Object
(
    [aSelectors:Sabberworm\CSS\RuleSet\DeclarationBlock:private] => Array
        (
            [0] => Sabberworm\CSS\Property\Selector Object
                (
                    [sSelector:Sabberworm\CSS\Property\Selector:private] => .countdown_field
                    [iSpecificity:Sabberworm\CSS\Property\Selector:private] => 
                )

        )

    [aRules:Sabberworm\CSS\RuleSet\RuleSet:private] => Array
        (
            [width] => Array
                (
                    [0] => Sabberworm\CSS\Rule\Rule Object
                        (
                            [sRule:Sabberworm\CSS\Rule\Rule:private] => width
                            [mValue:Sabberworm\CSS\Rule\Rule:private] => Sabberworm\CSS\Value\Size Object
                                (
[...]

PR #793, opened 2025-01-19
[TASK] Add TestCase for `Value\Color`

JakeQZ:
This covers expected behaviour for CSS level 3
(some of which is also covered by more general tests).

Also included are some commented-out tests for CSS Color Module Level 4, support for which is yet to be implemented,
as well as for some syntaxes that should be rejected (but currently are not).

Precursor to resolving #755 and supporting CSS Color Module Level 4.

### DIFF
--- a/tests/Unit/Value/ColorTest.php
+++ b/tests/Unit/Value/ColorTest.php
@@ -1,0 +1,219 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit\Value;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Parsing\ParserState;
+use Sabberworm\CSS\Parsing\SourceException;
+use Sabberworm\CSS\Settings;
+use Sabberworm\CSS\Value\Color;
+
+/**
+ * Note: some test data is currently commented-out.
+ * These cover
+ * - CSS Color Module Level 4 syntaxes that are not yet supported;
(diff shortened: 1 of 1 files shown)


JakeQZ:
@oliverklee, not sure how you feel about having commented-out tests, but it seemed to me, for now, a convenient way of capturing what needs to be done.

oliverklee (approved):

oliverklee:
> not sure how you feel about having commented-out tests, but it seemed to me, for now, a convenient way of capturing what needs to be done.

I'm okay with those.

For still-to-be-done test cases, I prefer `markTestIncomplete`, and for still-to-be-done data provider cases, I slightly prefer `//` comments over `/* */` comments as those make re-enabling single cases a bit less of a hassle.

JakeQZ:
> I slightly prefer `//` comments over `/* */` comments as those make re-enabling single cases a bit less of a hassle.

With

```php
/*
someCode();
//*/
```

the code can be temporarily uncommented by simply adding a forward-slash at the start of the comment:

```php
//*
someCode();
//*/
```

JakeQZ:
> @JakeQZ Is this something to backport?

It would be needed if we wanted to backport #797 and its follow-up.  But see https://github.com/MyIntervals/PHP-CSS-Parser/pull/797#issuecomment-2603407489.  I think we are wasting our time with the backports, and should cut a 9.0 release as soon as #755 is resolved.

PR #796, opened 2025-01-19

oliverklee:
The PHPStan version we used seems to be no longer available.


JakeQZ (approved):
I'm assuming the GPG keys are correct.

PR #797, opened 2025-01-20
[FEATURE] Support rgba and rgb (etc.) being aliases

JakeQZ:
As of CSS Color Module Level 4, `rgba` is an alias of `rgb`; likewise, `hsla` is an alias of `hsl`.

This change allows any of the above color functions to contain either three or four arguments, with alpha assumed as the fourth, and allowed to be absent.

### DIFF
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -72,23 +72,48 @@ class Color extends CSSFunction
             $oParserState->consumeWhiteSpace();
             $oParserState->consume('(');
 
+            // CSS Color Module Level 4 says that `rgb` and `rgba` are now aliases; likewise `hsl` and `hsla`.
+            // So, attempt to parse with the `a`, and allow for it not being there.
+            switch ($sColorMode) {
+                case 'rgb':
+                case 'hsl':
+                    $colorModeForParsing = $sColorMode . 'a';
+                    $mayHaveOptionalAlpha = true;
+                    break;
+                case 'rgba':
+                case 'hsla':
+                    $colorModeForParsing = $sColorMode;
+                    $mayHaveOptionalAlpha = true;
+                    break;
+                default:
+                    $colorModeForParsing = $sColorMode;
+                    $mayHaveOptionalAlpha = false;
+                    break;
+            }
+
             $bContainsVar = false;
-            $iLength = $oParserState->strlen($sColorMode);
+            $iLength = $oParserState->strlen($colorModeForParsing);
             for ($i = 0; $i < $iLength; ++$i) {
                 $oParserState->consumeWhiteSpace();
                 if ($oParserState->comes('var')) {
-                    $aColor[$sColorMode[$i]] = CSSFunction::parseIdentifierOrFunction($oParserState);
+                    $aColor[$colorModeForParsing[$i]] = CSSFunction::parseIdentifierOrFunction($oParserState);
                     $bContainsVar = true;
                 } else {
-                    $aColor[$sColorMode[$i]] = Size::parse($oParserState, true);
+                    $aColor[$colorModeForParsing[$i]] = Size::parse($oParserState, true);
                 }
 
-                if ($bContainsVar && $oParserState->comes(')')) {
-                    // With a var argument the function can have fewer arguments
+                // This must be done first, to consume comments as well, so that the `comes` test will work.
+                $oParserState->consumeWhiteSpace();
+
+                // With a var argument, the function can have fewer arguments.
+                // And as of CSS Color Module Level 4, the alpha argument is optional.
+                $canCloseNow =
+                    $bContainsVar ||
+                    $mayHaveOptionalAlpha && $i >= $iLength - 2;
+                if ($canCloseNow && $oParserState->comes(')')) {
                     break;
                 }
 
-                $oParserState->consumeWhiteSpace();
                 if ($i < ($iLength - 1)) {
                     $oParserState->consume(',');
                 }
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -83,8 +83,8 @@ class Color extends CSSFunction
                     $colorModeForParsing = 'hsla';
                     $mayHaveOptionalAlpha = true;
                     break;
-                // These two cases are handled identically.
                 case 'rgba':
+                    // This is handled identically to the following case.
                 case 'hsla':
                     $colorModeForParsing = $sColorMode;
                     $mayHaveOptionalAlpha = true;
@@ -112,7 +112,7 @@ class Color extends CSSFunction
                 // And as of CSS Color Module Level 4, the alpha argument is optional.
                 $canCloseNow =
                     $bContainsVar ||
-                    $mayHaveOptionalAlpha && $i >= $iLength - 2;
+                    ($mayHaveOptionalAlpha && $i >= $iLength - 2);
                 if ($canCloseNow && $oParserState->comes(')')) {
                     break;
                 }
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -76,10 +76,14 @@ class Color extends CSSFunction
             // So, attempt to parse with the `a`, and allow for it not being there.
             switch ($sColorMode) {
                 case 'rgb':
-                case 'hsl':
-                    $colorModeForParsing = $sColorMode . 'a';
+                    $colorModeForParsing = 'rgba';
                     $mayHaveOptionalAlpha = true;
                     break;
+                case 'hsl':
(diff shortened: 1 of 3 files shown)


JakeQZ:
I don't consider this a bug, because it wasn't a problem until CSS Color Module Level 4 came along.

JakeQZ:
Another precursor to #755.

src/Value/Color.php
```diff
@@ -72,23 +72,48 @@ public static function parse(ParserState $oParserState, bool $bIgnoreCase = fals
             $oParserState->consumeWhiteSpace();
             $oParserState->consume('(');
 
+            // CSS Color Module Level 4 says that `rgb` and `rgba` are now aliases; likewise `hsl` and `hsla`.
+            // So, attempt to parse with the `a`, and allow for it not being there.
+            switch ($sColorMode) {
+                case 'rgb':
```
oliverklee, line 78: For the two cases with the fall-through, please let's have a comment explaining that this is intentional, e.g., like this:

```php
// intentional fall-through
```

src/Value/Color.php
```diff
@@ -72,23 +72,48 @@ public static function parse(ParserState $oParserState, bool $bIgnoreCase = fals
...
+                    break;
+                case 'rgba':
+                case 'hsla':
+                    $colorModeForParsing = $sColorMode;
+                    $mayHaveOptionalAlpha = true;
+                    break;
+                default:
+                    $colorModeForParsing = $sColorMode;
+                    $mayHaveOptionalAlpha = false;
+                    break;
```
oliverklee, line 91: The `default` case does not need a `break`.

src/Value/Color.php
```diff
@@ -72,23 +72,48 @@ public static function parse(ParserState $oParserState, bool $bIgnoreCase = fals
             $oParserState->consumeWhiteSpace();
             $oParserState->consume('(');
 
+            // CSS Color Module Level 4 says that `rgb` and `rgba` are now aliases; likewise `hsl` and `hsla`.
+            // So, attempt to parse with the `a`, and allow for it not being there.
+            switch ($sColorMode) {
+                case 'rgb':
+                case 'hsl':
+                    $colorModeForParsing = $sColorMode . 'a';
```
oliverklee, line 80: I think it would reduce the cognitive complexity if we used string literals for `$colorModeForParsing` in the different cases (except for the `default` case)  instead of using `$sColorMode`.

src/Value/Color.php
```diff
@@ -72,23 +72,48 @@ public static function parse(ParserState $oParserState, bool $bIgnoreCase = fals
...
-                    $aColor[$sColorMode[$i]] = Size::parse($oParserState, true);
+                    $aColor[$colorModeForParsing[$i]] = Size::parse($oParserState, true);
                 }
 
-                if ($bContainsVar && $oParserState->comes(')')) {
-                    // With a var argument the function can have fewer arguments
+                // This must be done first, to consume comments as well, so that the `comes` test will work.
+                $oParserState->consumeWhiteSpace();
+
+                // With a var argument, the function can have fewer arguments.
```
oliverklee, line 108: What does "var" mean here? Is it an abbreviation? Oder is it code (and hence should be in backticks)?

src/Value/Color.php
```diff
@@ -72,23 +72,48 @@ public static function parse(ParserState $oParserState, bool $bIgnoreCase = fals
...
+                // And as of CSS Color Module Level 4, the alpha argument is optional.
```
oliverklee, line 109:
```suggestion
                // And as of CSS Color Module Level 4, the `alpha` argument is optional.
```

oliverklee (changes requested):

oliverklee:
Also, is this something to backport?

src/Value/Color.php
JakeQZ: It's common for `case`s with identical handlers to be grouped like this without comment - I've also seen them written on the same line (e.g. `case 'rgba': case 'hsla':`).  I think it would be clunky to have a comment between the `case`s, and make the code less readable.  However, I've added a comment above the paried `case`s to clarify the intent.

src/Value/Color.php
JakeQZ: Done.

src/Value/Color.php
oliverklee: Thanks! This makes is easy to see that not having the `break` was intentional instead of an oversight.

src/Value/Color.php
JakeQZ: I believe it to be a CSS custom property substition, and refers to the CSS `var` function.  I moved the comment from within the `if` statement, without changing it, but have now added backticks.

src/Value/Color.php
JakeQZ: 'Alpha' here is not code, it is a description of the argument.

JakeQZ:
Blocked by #798.  I've pushed changes suggested in review, and included the #798 change in those, just to confirm that all CI tests would pass.  Once #798 is merged, and this is rebased, that change should drop out of this.

JakeQZ:
> Also, is this something to backport?

If 9.0 is still some way off, then yes, but see below.

This is mutlipart change.  The key change to support space-separated arguments with a forward-slash separating the alpha argument is yet to come.  It's an enhancement rather than a bugfix, but one that is needed sooner rather than later, since the CSS syntax has evolved.

I noted in a recent backport that some syntaxes had to be changed to support PHP 5.6.  OTOH, I recently saw a message from my hosting provider saying that if I wanted to continue using a PHP version below 8.0, I'd have to pay an additional premium.

The current `main` requires PHP >= 7.2, which itself is well beyond EOL.

I think it would be more than reasonable to cut a 9.0 release fairly soon, regardless of not having completed all that we planned for it, and stop wasting time on the backports (focusing on something more interesting instead).

src/Value/Color.php
JakeQZ: > This is delayed because I need to figure out how to work around [PHP-CS-Fixer/PHP-CS-Fixer#6572](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/issues/6572). [PHP-CS-Fixer/PHP-CS-Fixer#7624](https://github.com/PHP-CS-Fixer/PHP-CS-Fixer/pull/7624) might offer a solution.

#798 - Unfortunately is seems we need to disable a rule that has become broken.  [One repository](https://github.com/akeneo/pim-community-dev/pull/17431) has even placed an upper version limit on PHP CS Fixer, pending resolution to the breakage.

oliverklee (changes requested):

src/Value/Color.php
```diff
@@ -72,23 +72,51 @@ public static function parse(ParserState $oParserState, bool $bIgnoreCase = fals
...
-                if ($bContainsVar && $oParserState->comes(')')) {
-                    // With a var argument the function can have fewer arguments
+                // This must be done first, to consume comments as well, so that the `comes` test will work.
+                $oParserState->consumeWhiteSpace();
+
+                // With a `var` argument, the function can have fewer arguments.
+                // And as of CSS Color Module Level 4, the alpha argument is optional.
+                $canCloseNow =
+                    $bContainsVar ||
+                    $mayHaveOptionalAlpha && $i >= $iLength - 2;
```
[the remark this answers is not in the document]
JakeQZ: The same argument could apply to the other operators, but I think

```php
$canCloseNow = (
    $bContainsVar ||
    ($mayHaveOptionalAlpha && ($i >= ($iLength - 2)))
);
```

would be clunky and less readable.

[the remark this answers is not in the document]
oliverklee: I'd be fine with either change (yours or mine).

src/Value/Color.php
```diff
@@ -72,23 +72,51 @@ public static function parse(ParserState $oParserState, bool $bIgnoreCase = fals
...
+            switch ($sColorMode) {
+                case 'rgb':
+                    $colorModeForParsing = 'rgba';
+                    $mayHaveOptionalAlpha = true;
+                    break;
+                case 'hsl':
+                    $colorModeForParsing = 'hsla';
+                    $mayHaveOptionalAlpha = true;
+                    break;
+                // These two cases are handled identically.
```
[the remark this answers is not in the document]
JakeQZ: Done.

src/Value/Color.php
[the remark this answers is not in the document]
JakeQZ: Given I prefer not to have too many extra brackets cluttering up the code, I've gone with yours.

oliverklee (approved):

PR #799, opened 2025-01-23
[CLEANUP] Split `Color::parse` into separate methods

JakeQZ:
One for hex colors, and one for color functions.
This reduces cyclomatic complexity on a per-method basis.

### DIFF
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -30,104 +30,125 @@ class Color extends CSSFunction
      */
     public static function parse(ParserState $oParserState, bool $bIgnoreCase = false): CSSFunction
     {
-        $aColor = [];
         if ($oParserState->comes('#')) {
-            $oParserState->consume('#');
-            $sValue = $oParserState->parseIdentifier(false);
-            if ($oParserState->strlen($sValue) === 3) {
-                $sValue = $sValue[0] . $sValue[0] . $sValue[1] . $sValue[1] . $sValue[2] . $sValue[2];
-            } elseif ($oParserState->strlen($sValue) === 4) {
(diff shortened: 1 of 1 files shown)


JakeQZ:
I'd recommend reviewing the changes with the 'hide whitespace' option enabled; otherwise it will look like there are a lot of changes, when in fact there aren't, because all that's happened is that the indentation has been decreased.  (I don't know if there's a similar option for merging/rebasing, should it be needed.)

src/Value/Color.php
```diff
@@ -30,104 +30,125 @@ public function __construct(array $aColor, $iLineNo = 0)
...
-            if ($oParserState->strlen($sValue) === 3) {
-                $sValue = $sValue[0] . $sValue[0] . $sValue[1] . $sValue[1] . $sValue[2] . $sValue[2];
-            } elseif ($oParserState->strlen($sValue) === 4) {
-                $sValue = $sValue[0] . $sValue[0] . $sValue[1] . $sValue[1] . $sValue[2] . $sValue[2] . $sValue[3]
-                    . $sValue[3];
-            }
+            return self::parseHexColor($oParserState);
+        } else {
+            return self::parseColorFunction($oParserState);
+        }
```
oliverklee, line 37: This should either be a guard clause (with the code in the `else' branch then without the `else` keyword, or something along those lines (which is what I'd prefer):

```php
return $oParserState->comes('#') ? self::parseHexColor($oParserState) : self::parseColorFunction($oParserState);
```

oliverklee (changes requested):

src/Value/Color.php
JakeQZ: Yes, I had wondered about that given that I actually reduced the `return` statement in `parseColorFunction` using the ternary conditional operator.  Changed.

oliverklee (approved):

PR #800, opened 2025-01-23
[FEATURE] Parse color functions with modern syntax

### LINKED ISSUE #755: Parsing RGB with alpha is broken

JakeQZ:
Also include tests for color functions with CSS variable substitutions where they use the "legacy" syntax.

Resolves #755.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@ Please also have a look at our
 
 ### Added
 
-- Partial support for CSS Color Module Level 4 syntax:
+- Partial support for CSS Color Module Level 4:
   - `rgb` and `rgba`, and `hsl` and `hsla` are now aliases (#797}
+  - Parse color functions that use the "modern" syntax
 - Add official support for PHP 8.4 (#657)
 - Support arithmetic operators in CSS function arguments (#607)
 - Add support for inserting an item in a CSS list (#545)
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -10,8 +10,9 @@ Please also have a look at our
 
 ### Added
 
-- Partial support for CSS Color Module Level 4 syntax:
+- Partial support for CSS Color Module Level 4:
   - `rgb` and `rgba`, and `hsl` and `hsla` are now aliases (#797}
+  - Parse color functions that use the "modern" syntax
```
oliverklee, line 15:
```suggestion
  - Parse color functions that use the "modern" syntax (#800)
```

oliverklee (approved):
Wow… :heart:

CHANGELOG.md
JakeQZ: I'd intended to do this once I'd determined what the PR number would be, but then forgot.  Now added.

oliverklee (approved):

JakeQZ:
> Wow… ❤️

I'd done most of this before I figured that the `parse` method should probably first be split up.

I've added #801 to capture all tasks and issues relating to colours that I can think of, while I'm in this neck of the woods.

PR #802, opened 2025-01-24
[CLEANUP] Avoid Hungarian notation in `Color` class

JakeQZ:
Also rename some variables to be more descriptive and/or avoid abbreviations.


oliverklee (changes requested):
🧹

oliverklee (approved):

PR #803, opened 2025-01-25
[BUGFIX] Don't render `rgb` colors with `%` values as hex

JakeQZ:
The only percentage values that could be reliably converted to hex notation are 0%, 20%, 40%, etc.
It's beyond the scope of this library to do that.

Also add additional tests to confirm parsing of percentage values. Some of these are commented out, because the input data would result in rendering in an invalid format. (The "legacy" syntax does not allow a mixture of `percentage`s and `number`s, so it would be necessary to implement rendering in the "modern" syntax to resolve those cases, which is beyond the scope of this PR.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -45,6 +45,7 @@ Please also have a look at our
 
 ### Fixed
 
+- Don't render `rgb` colors with percentage values using hex notation (#803)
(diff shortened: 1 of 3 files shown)


oliverklee (approved):
LGTM.

PR #804, opened 2025-01-25
[CLEANUP] Avoid Hungarian notion for the `Commentable` interface

oliverklee:
Also adapt the implementing classes accordingly.

Part of #756.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -36,7 +36,7 @@ abstract class CSSList implements Renderable, Commentable
     /**
      * @var array<array-key, Comment>
      */
-    protected $aComments;
+    protected $comments;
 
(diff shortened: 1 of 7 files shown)


JakeQZ (approved):
I can't help but notice that the various classes implementing `Commentable` seen to have exactly the same implementation.  So I think this could be tidied up by providing a `trait` for a 'standard' `Commentable` implementation - but is beyond the scope of this PR.

JakeQZ:
> I can't help but notice that the various classes implementing `Commentable` seen to have exactly the same implementation. So I think this could be tidied up by providing a `trait` for a 'standard' `Commentable` implementation - but is beyond the scope of this PR.

Added #813 to capture this.

PR #805, opened 2025-01-25
[TASK] Use native `void` return type declarations

oliverklee:
Also include previous changes for adding native type declarations or strict mode into the changelog.

### DIFF
--- a/src/CSSList/CSSBlockList.php
+++ b/src/CSSList/CSSBlockList.php
@@ -30,10 +30,8 @@ abstract class CSSBlockList extends CSSList
 
     /**
      * @param array<int, DeclarationBlock> $aResult
-     *
-     * @return void
      */
-    protected function allDeclarationBlocks(array &$aResult)
+    protected function allDeclarationBlocks(array &$aResult): void
     {
         foreach ($this->aContents as $mContent) {
             if ($mContent instanceof DeclarationBlock) {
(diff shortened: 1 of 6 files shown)


oliverklee:
This will probably clash with #804.

src/CSSList/CSSBlockList.php
```diff
@@ -30,10 +30,8 @@ public function __construct($iLineNo = 0)
 
     /**
      * @param array<int, DeclarationBlock> $aResult
-     *
-     * @return void
      */
-    protected function allDeclarationBlocks(array &$aResult)
+    protected function allDeclarationBlocks(array &$aResult): void
```
JakeQZ, line 34: Can't help thinking that these `all...` methods should perhaps be returning the result, rather than having it as a parameter passed by reference - unless the array is supposed to be merged with some original array - but that is beyond the scope of this PR.

JakeQZ (approved):
LGTM.  Yes it does clash with #804.

JakeQZ:
> Yes it does clash with #804.

I used the visual merge, which seems to have worked OK.

src/CSSList/CSSBlockList.php
JakeQZ: Looked at the code, and these functions call each other to build up the array, so it does need to be a parameter passed by reference.

PR #807, opened 2025-01-25
[DOCS] Remove section on testing from `README.md`

oliverklee:
We now have instructions on testing in `CONTRIBUTING.md`.


JakeQZ (approved):
It is (or should be) covered in `CONTRIBUTING.md`.

I've added #815 to cover what's removed here but doesn't seem to be quite fully covered there.

PR #809, opened 2025-01-25
[BUGFIX] Drop `@covers` annotation referencing inexistent method

oliverklee:


JakeQZ (changes requested):
The method definitely exists.  The error is for some other reason.  I don't know what.

The `@covers` annotation looks correct to me.

oliverklee:
As `@covers` annotations referencing class methods are not recommended anyway ([documentation on this](https://docs.phpunit.de/en/8.5/annotations.html#covers)), I have now created #820 as an alternative.

PR #818, opened 2025-01-26
[TASK] Use native types for all constructor parameters

oliverklee:
Part of #811


oliverklee:
Blocked by (at least) #814 and #817.

JakeQZ:
This is blocked so marking as draft.

PR #820, opened 2025-01-26
[TASK] Avoid using `@covers` annotations referencing methods

oliverklee:
According to the PHPUnit documentation [1], `@covers` annotations are only recommended to reference classes (and global functions), but not methods from classes.

Also sort the `@covers` annotations.

Note: This will probably artificially increase our calculated code coverage numbers until we migrate our tests to more focused unit tests.

Closes #809

[1] https://docs.phpunit.de/en/8.5/annotations.html#covers

### DIFF
--- a/tests/ParserTest.php
+++ b/tests/ParserTest.php
@@ -26,16 +26,17 @@ use Sabberworm\CSS\Value\Size;
 use Sabberworm\CSS\Value\URL;
 
 /**
+ * @covers \Sabberworm\CSS\CSSList\Document
  * @covers \Sabberworm\CSS\Parser
- * @covers \Sabberworm\CSS\CSSList\Document::parse
- * @covers \Sabberworm\CSS\Rule\Rule::parse
- * @covers \Sabberworm\CSS\RuleSet\DeclarationBlock::parse
- * @covers \Sabberworm\CSS\Value\CalcFunction::parse
- * @covers \Sabberworm\CSS\Value\Color::parse
- * @covers \Sabberworm\CSS\Value\CSSString::parse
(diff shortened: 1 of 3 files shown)


tests/ParserTest.php
```diff
@@ -26,16 +26,17 @@
...
- * @covers \Sabberworm\CSS\Value\URL::parse
+ * @covers \Sabberworm\CSS\RuleSet\DeclarationBlock
+ * @covers \Sabberworm\CSS\Rule\Rule
+ * @covers \Sabberworm\CSS\Value\CSSString
+ * @covers \Sabberworm\CSS\Value\CalcFunction
+ * @covers \Sabberworm\CSS\Value\Color
+ * @covers \Sabberworm\CSS\Value\LineName
+ * @covers \Sabberworm\CSS\Value\Size
+ * @covers \Sabberworm\CSS\Value\URL
+ * @covers \Sabberworm\CSS\Value\Value
```
JakeQZ, line 39: Was it intentional to add that it covers `Value\Value`?

JakeQZ (commented):
I've added a query about an extra `@covers`.  Otherwise looks fine.

> This will probably artificially increase our calculated code coverage numbers

Maybe not artificially, as the tests are probably covering more than we previously stated they did.

tests/ParserTest.php
oliverklee: Yes, as `missingPropertyValueStrict` and `missingPropertyValueLenient` had been marked as covering `\Sabberworm\CSS\Value\Value::parseValue()`.

JakeQZ (approved):

PR #821, opened 2025-01-26
[TASK] Migrate part of `CommentTests` to unit tests

oliverklee:
Part of #758.


JakeQZ (approved):
Agree the remaining tests probably belong in a different `TestCase`, such as one for `OutputFormat`.

PR #824, opened 2025-01-26
[CLEANUP] Use spaces for indenting the fixture CSS files

oliverklee:
This makes editing the fixture files a bit easier as all other files in this project also use spaces for indentation, not tabs.

Note: This has no effect on CSS is rendered (which still defaults to tabs at the moment).


JakeQZ (approved):
Some tests might be intended to confirm that the parse can handle alternative whitespace forms, but that doesn't seem to be the case for any of these, not even `whitespace.css`.

PR #825, opened 2025-01-27
[CLEANUP] Extract method `Color::renderAsHex`

JakeQZ:
An additional method to render in the "modern" syntax will soon be needed. Splitting up the `render` method into sub-methods will help as a precursor to that.

### DIFF
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -230,15 +230,9 @@ class Color extends CSSFunction
             && \implode('', \array_keys($this->aComponents)) === 'rgb'
             && $this->allComponentsAreNumbers()
         ) {
-            $result = \sprintf(
-                '%02x%02x%02x',
-                $this->aComponents['r']->getSize(),
-                $this->aComponents['g']->getSize(),
-                $this->aComponents['b']->getSize()
-            );
-            return '#' . (($result[0] == $result[1]) && ($result[2] == $result[3]) && ($result[4] == $result[5])
-                    ? "$result[0]$result[2]$result[4]" : $result);
+            return $this->renderAsHex();
         }
+
         return parent::render($outputFormat);
     }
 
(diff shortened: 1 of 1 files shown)


src/Value/Color.php
```diff
@@ -256,4 +250,23 @@ private function allComponentsAreNumbers(): bool
...
         return true;
     }
+
+    /**
+     * Note that this method assumes the following:
+     * - The {@see aComponents} array has keys for `r`, `g` and `b`;
+     * - The values in the array are all instances of {@see Size}.
+     *
+     * Errors will be triggered or thrown if this is not the case.
+     */
```
oliverklee, line 260: We can make the return type even more specific:

`@return non-empty-string`

src/Value/Color.php
```diff
@@ -256,4 +250,23 @@ private function allComponentsAreNumbers(): bool
 
         return true;
     }
+
+    /**
+     * Note that this method assumes the following:
+     * - The {@see aComponents} array has keys for `r`, `g` and `b`;
```
oliverklee, line 256:
```suggestion
     * - The `aComponents` array has keys for `r`, `g` and `b`;
```

src/Value/Color.php
```diff
@@ -256,4 +250,23 @@ private function allComponentsAreNumbers(): bool
...
+     */
+    private function renderAsHex(): string
+    {
+        $result = \sprintf(
+            '%02x%02x%02x',
+            $this->aComponents['r']->getSize(),
+            $this->aComponents['g']->getSize(),
+            $this->aComponents['b']->getSize()
+        );
+        return '#' . (($result[0] == $result[1]) && ($result[2] == $result[3]) && ($result[4] == $result[5])
```
oliverklee, line 269: While we're at it, I propose adding a local variable `$canUseShorthand`.

oliverklee (changes requested):

src/Value/Color.php
JakeQZ: Done

src/Value/Color.php
JakeQZ: Done.  I also changed the expression to avoid variable substitution within a string.

JakeQZ:
Made suggested changes, also the one to the DocBlock which was in the email I received but doesn't appear here.

oliverklee (approved):

PR #826, opened 2025-01-27
[BUGFIX] Correct type annotation for `ValueList` components

JakeQZ:
The `Color` class uses this property (`aComponents`) as an associative array. Therefore all related type annotations need `array-key` rather than `int` for the array key type.


JakeQZ:
It's a bug in the DocBlock, rather than the actual code, so I don't think warrants a changelog entry.

oliverklee (approved):

PR #828, opened 2025-01-27
[BUGFIX] Create `Size` with correct types in `expandBackgroundShorthand`

oliverklee:
This is the v8.x backport of #814.


oliverklee:
It looks like I'll need to update the GPG keys (again). Sigh.

JakeQZ (approved):

PR #832, opened 2025-01-27
[TASK] Drop the PHIVE-installed development tools in V8.x

oliverklee:
It has turned out that our current set of PHIVE-installed development tools for the V8.x maintenance branch is no longer installable due to issues with the GPG keys used for signing the PHARs.

As the maintenance branch has very little activity anyway now, relying on PHP linting and the unit tests should suffice, saving us the hassle with PHIVE on that branch.

(We're still keeping some `.gitignore` entries related to the
tools to make switching branches easier.)

Closes #831
Closes #830


JakeQZ (approved):
We still have the unit tests which should catch any major mistakes.  The other tools check for coding style and integrity, and have only relatively recently been added.

I've forgotton the reason for having the other tools installed with PHIVE rather than Composer.

oliverklee:
The reason was that the latest version of the tools (which we want to use) does not cover the PHP version range we support, and hence would make any `composer install` impossible.

PR #835, opened 2025-01-27
[CLEANUP] Avoid Hungarian notation for `oRule`/`oRuleSet`

oliverklee:
Part of #756


JakeQZ (approved):

JakeQZ:
Will need a rebase...

JakeQZ:
Note this change also does the same for `$oRuleSet`.

oliverklee:
> Note this change also does the same for `$oRuleSet`.

Indeed - good catch! I've updated the PR title and commit message accordingly.

JakeQZ (approved):

Issue #801, opened 2025-01-23
Supporting CSS Color Module Level 4 and 5


JakeQZ:
This is a task list of enhancements required to fully support the above (strictly).

High priority (ideally included in next release)

- [x] Allow interchangeability of `rgb` and `rgba`, and `hsl` and `hsla` (#797)
- [x] Parse color functions which use the "modern" syntax (#800)
- [ ] Support the `none` keywod in the "modern" syntax
- [x] Add tests to confirm that all varieties of `angle` are permitted for `hue` (#793)
- [ ] Add tests to confirm that both `number` and `percentage` are permitted where allowed (i.e. in "modern" `rgb` syntax)

Low priority (would be nice to have, but we have other fish to fry in the meantime, so to speak)

- [ ] Add rendering option to output color functions using "modern" syntax
- [ ] Reject color functions with invalid `length` arguments, such as `10px`
- [ ] Parse color functions containing `var()` substitutions as `Color` rather than generic `CSSFunction`
- [ ] Support additional color functions (these will currently be parsed as generic CSS functions, which may not work reliably):
  - [ ] `hwb`
  - [ ] `lab`
  - [ ] `lch`
  - [ ] `oklab`
  - [ ] `oklch`
- [ ] Support `color` function:
  - [ ] [Level 4](https://drafts.csswg.org/css-color/#color-function)
  - [ ] [Level 5](https://drafts.csswg.org/css-color-5/#color-function)
- [ ] Support [relative colors](https://drafts.csswg.org/css-color-5/#relative-RGB) (level 5)

PR #840, opened 2025-01-27
[BUGFIX] Render RGB functions with "modern" syntax if required

JakeQZ:
The "legacy" syntax does not allow a mixture of `percentage`s and `number`s for the red, green and blue components.
So if `rgb`/`rgba` functions that have such a mixture are rendered in the "legacy" syntax, an invalid property value will result.

An `OutputFormat` option to use the "modern" syntax throughout will be added later (see #801).

### DIFF
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -231,6 +231,8 @@ class Color extends CSSFunction
             && $this->allComponentsAreNumbers()
         ) {
             return $this->renderAsHex();
+        } elseif ($this->shouldRenderInModernSyntax()) {
+            return $this->renderInModernSyntax($outputFormat);
         }
 
         return parent::render($outputFormat);
@@ -281,5 +283,67 @@ class Color extends CSSFunction
         $canUseShortVariant = ($result[0] == $result[1]) && ($result[2] == $result[3]) && ($result[4] == $result[5]);
 
         return '#' . ($canUseShortVariant ? $result[0] . $result[2] . $result[4] : $result);
+    }
+
+    /**
+     * The "legacy" syntax does not allow RGB colors to have a mixture of `percentage`s and `number`s.
+     */
+    private function shouldRenderInModernSyntax(): bool
+    {
+        $function = $this->getRealName();
+        if ($function !== 'rgb' && $function !== 'rgba') {
+            return false;
+        }
+
+        $hasPercentage = false;
+        $hasNumber = false;
+        foreach ($this->aComponents as $key => $value) {
+            if ($key === 'a') {
+                // ALpha can have units that don't match those of the RGB components in the "legacy" syntax.
+                // So it is not necessary to check it.  It's also always last, hence `break` rather than `continue`.
+                break;
+            }
+            if (!$value instanceof Size) {
+                // Unexpected, unknown, or modified via the API
+                return false;
+            }
+            $unit = $value->getUnit();
+            // `switch` only does loose comparison
+            if ($unit === null) {
+                $hasNumber = true;
+            } elseif ($unit === '%') {
+                $hasPercentage = true;
+            } else {
+                // Invalid unit
+                return false;
+            }
+        }
+
+        return $hasPercentage && $hasNumber;
+    }
+
+    /**
+     * @return non-empty-string
+     */
+    private function renderInModernSyntax(OutputFormat $outputFormat): string
+    {
+        \end($this->aComponents);
+        if (\key($this->aComponents) === 'a') {
+            $alpha = $this->aComponents['a'];
+            $componentsWithoutAlpha = \array_diff_key($this->aComponents, ['a' => 0]);
+        } else {
+            $componentsWithoutAlpha = $this->aComponents;
+        }
+
+        $arguments = $outputFormat->implode(' ', $componentsWithoutAlpha);
+        if (isset($alpha)) {
+            $arguments = $outputFormat->implode(
+                $outputFormat->spaceBeforeListArgumentSeparator('/') . '/'
+                    . $outputFormat->spaceAfterListArgumentSeparator('/'),
+                [$arguments, $alpha]
+            );
+        }
+
+        return $this->getName() . '(' . $arguments . ')';
     }
 }
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -299,7 +299,7 @@ class Color extends CSSFunction
         $hasNumber = false;
         foreach ($this->aComponents as $key => $value) {
             if ($key === 'a') {
-                // ALpha can have units that don't match those of the RGB components in the "legacy" syntax.
+                // Alpha can have units that don't match those of the RGB components in the "legacy" syntax.
                 // So it is not necessary to check it.  It's also always last, hence `break` rather than `continue`.
                 break;
             }
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -303,9 +303,7 @@ class Color extends CSSFunction
      */
     private function shouldRenderInModernSyntax(): bool
     {
-        static $colorFunctionsThatWithMixedValueTypesCannotBeRenderedInLegacySyntax = ['rgb', 'rgba'];
-        $colorFunction = $this->getRealName();
-        if (!\in_array($colorFunction, $colorFunctionsThatWithMixedValueTypesCannotBeRenderedInLegacySyntax, true)) {
+        if (!$this->colorFunctionMayHaveMixedValueTypes($this->getRealName())) {
             return false;
         }
 
@@ -317,7 +315,7 @@ class Color extends CSSFunction
                 // So it is not necessary to check it.  It's also always last, hence `break` rather than `continue`.
                 break;
             }
-            if (!$value instanceof Size) {
+            if (!($value instanceof Size)) {
                 // Unexpected, unknown, or modified via the API
                 return false;
             }
@@ -334,6 +332,19 @@ class Color extends CSSFunction
         }
 
         return $hasPercentage && $hasNumber;
+    }
+
+    /**
+     * Some color functions, such as `rgb`,
+     * may have a mixture of `percentage`, `number`, or possibly other types in their arguments.
+     *
+     * Note that this excludes the alpha component, which is treated separately.
+     */
+    private function colorFunctionMayHaveMixedValueTypes(string $function): bool
+    {
+        $functionsThatMayHaveMixedValueTypes = ['rgb', 'rgba'];
+
+        return \in_array($function, $functionsThatMayHaveMixedValueTypes, true);
     }
 
     /**
--- a/src/Value/Color.php
+++ b/src/Value/Color.php
@@ -231,7 +231,9 @@ class Color extends CSSFunction
             && $this->allComponentsAreNumbers()
         ) {
             return $this->renderAsHex();
-        } elseif ($this->shouldRenderInModernSyntax()) {
+        }
+
+        if ($this->shouldRenderInModernSyntax()) {
             return $this->renderInModernSyntax($outputFormat);
         }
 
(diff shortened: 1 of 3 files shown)


JakeQZ:
This is part-bug, part-feature.  It could be viewed as a bug in the new feature, or just a continuation of the implementation of that feature.

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
...
+        $function = $this->getRealName();
+        if ($function !== 'rgb' && $function !== 'rgba') {
+            return false;
+        }
+
+        $hasPercentage = false;
+        $hasNumber = false;
+        foreach ($this->aComponents as $key => $value) {
+            if ($key === 'a') {
+                // ALpha can have units that don't match those of the RGB components in the "legacy" syntax.
```
oliverklee, line 302:
```suggestion
                // Alpha can have units that don't match those of the RGB components in the "legacy" syntax.
```

JakeQZ: Good spot.  Committed suggestion.

oliverklee (changes requested):
Looks functionally okay to me. I think we can make the code more readable and easier to grok, though. I've left some comments.

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
...
+            }
+            $unit = $value->getUnit();
+            // `switch` only does loose comparison
+            if ($unit === null) {
+                $hasNumber = true;
+            } elseif ($unit === '%') {
+                $hasPercentage = true;
+            } else {
+                // Invalid unit
+                return false;
```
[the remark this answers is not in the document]
JakeQZ: > both `$hasPercentage` and `$hasNumber` are `false` at this moment,

No, a previous loop iteration may have set one or other.

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
...
+            $alpha = $this->aComponents['a'];
+            $componentsWithoutAlpha = \array_diff_key($this->aComponents, ['a' => 0]);
+        } else {
+            $componentsWithoutAlpha = $this->aComponents;
+        }
+
+        $arguments = $outputFormat->implode(' ', $componentsWithoutAlpha);
+        if (isset($alpha)) {
+            $arguments = $outputFormat->implode(
+                $outputFormat->spaceBeforeListArgumentSeparator('/') . '/'
```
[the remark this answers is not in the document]
JakeQZ: Note that `spaceBeforeListArgumentSeparator()` is only called once; the other call is to `spaceAfterListArgumentSeparator()`.

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
...
+        return $hasPercentage && $hasNumber;
+    }
+
+    /**
+     * @return non-empty-string
+     */
+    private function renderInModernSyntax(OutputFormat $outputFormat): string
+    {
+        \end($this->aComponents);
+        if (\key($this->aComponents) === 'a') {
```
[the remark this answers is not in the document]
JakeQZ: The order is specific and comes from the CSS function name (e.g. `rgba`).  There's a newer one `lab` in which `a` is not alpha.  `a` is only alpha if it is the final element (in all cases I'm aware of).

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
 
         return '#' . ($canUseShortVariant ? $result[0] . $result[2] . $result[4] : $result);
     }
+
+    /**
+     * The "legacy" syntax does not allow RGB colors to have a mixture of `percentage`s and `number`s.
```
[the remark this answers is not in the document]
JakeQZ: Those terms are part of the formal spec.  I've added links in the DocBlock.  I hope `{@link}` works with line breaks, otherwise we'll need some very long lines and probably need to disable some fixer rule.

src/Value/Color.php
```diff
@@ -231,6 +231,8 @@ public function render(OutputFormat $outputFormat): string
             && $this->allComponentsAreNumbers()
         ) {
             return $this->renderAsHex();
+        } elseif ($this->shouldRenderInModernSyntax()) {
```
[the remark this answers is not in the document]
JakeQZ: Yeah, that's cleaner.  Done.

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
...
         return '#' . ($canUseShortVariant ? $result[0] . $result[2] . $result[4] : $result);
     }
+
+    /**
+     * The "legacy" syntax does not allow RGB colors to have a mixture of `percentage`s and `number`s.
+     */
+    private function shouldRenderInModernSyntax(): bool
+    {
+        $function = $this->getRealName();
+        if ($function !== 'rgb' && $function !== 'rgba') {
```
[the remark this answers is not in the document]
JakeQZ: Done as a `static` local variable for now.  As this functionality is in its infancy, and subject to change to accommodate more features, I'd prefer not to introduce a (more concrete) class constant at this stage.

The variable name may be a record for the longest I've ever used.  But it saves a comment, and still fits on one line even in the `if (!\in_array(...))`

I also renamed `$function` to `$colorFunction` to help with out-of-context grokking.

src/Value/Color.php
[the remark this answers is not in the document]
JakeQZ: If there's an invalid unit, rendering in either "modern" or "legacy" will result in an invalid property value.

On the to-do list is to reject invalid units at the parsing stage (as browsers do), but that's high-hanging fruit.

For now, I'd like to retain current behaviour as much as possible, which means only rendering in the "modern" syntax when that would give a valid value and using the "legacy" syntax would not.  Note that it's possible to provide something like `rgb(0, 119, 0px)` in the "legacy" syntax, and that would currently be regurgitaged.  I don't want to change the output to the "modern" syntax unneccessarily, when that would not be valid anyway.

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
...
+
+    /**
+     * @return non-empty-string
+     */
+    private function renderInModernSyntax(OutputFormat $outputFormat): string
+    {
+        \end($this->aComponents);
+        if (\key($this->aComponents) === 'a') {
+            $alpha = $this->aComponents['a'];
+            $componentsWithoutAlpha = \array_diff_key($this->aComponents, ['a' => 0]);
```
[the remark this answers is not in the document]
JakeQZ: I originally coded this like you're suggesting.  But then thought it was a bit odd that `$componentsWithoutAlpha` would initially (possibly) have the alpha.  I've reverted, and added a comment.

src/Value/Color.php
```diff
@@ -282,4 +284,66 @@ private function renderAsHex(): string
...
+        } else {
+            $componentsWithoutAlpha = $this->aComponents;
```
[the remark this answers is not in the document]
JakeQZ: Indeed.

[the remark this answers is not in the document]
JakeQZ: Except if we want

> a local variable `$hasAlpha`

we might still need an `else` clause.

src/Value/Color.php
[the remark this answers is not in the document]
JakeQZ: We can still have a local variable for the combined separator.  Done.

JakeQZ:
I've made most suggested changes, and commented on some as to why they can't be made.  There's a remaining dichotomy (I can only make or other of two suggested changes).  Hopfully this will be clearer now, with the other changes made...

JakeQZ:
I added some spec links to the latest commit message (re "legacy" and "modern") which will need to be integrated into the final commit message somehow, if desired.

src/Value/Color.php
```diff
@@ -233,6 +233,10 @@ public function render(OutputFormat $outputFormat): string
             return $this->renderAsHex();
```
oliverklee, line 233: Should we move these three conditions into a method `shouldRenderInLegacySyntax`?

src/Value/Color.php
```diff
@@ -282,4 +286,76 @@ private function renderAsHex(): string
...
+     *     https://www.w3.org/TR/css-color-4/#rgb-functions
+     *     The same in the CSS Color Module Level 4 W3C Candidate Recommendation Draft
+     *   } (as of 13 February 2024, at time of writing).
+     */
+    private function shouldRenderInModernSyntax(): bool
+    {
+        static $colorFunctionsThatWithMixedValueTypesCannotBeRenderedInLegacySyntax = ['rgb', 'rgba'];
+        $colorFunction = $this->getRealName();
+        if (!\in_array($colorFunction, $colorFunctionsThatWithMixedValueTypesCannotBeRenderedInLegacySyntax, true)) {
+            return false;
```
oliverklee, line 309: I propose we move the (static) variable and the `in_array` to a dedicated, well-named method.

Also, please generally let's not use `static` variables as they keep state across instances and are a pain to test. In this case, I think a (normal) local variable or a constant would be fine.

src/Value/Color.php
```diff
@@ -282,4 +286,76 @@ private function renderAsHex(): string
...
+        }
+
+        $hasPercentage = false;
+        $hasNumber = false;
+        foreach ($this->aComponents as $key => $value) {
+            if ($key === 'a') {
+                // Alpha can have units that don't match those of the RGB components in the "legacy" syntax.
+                // So it is not necessary to check it.  It's also always last, hence `break` rather than `continue`.
```
oliverklee, line 317: :heart:

src/Value/Color.php
```diff
@@ -282,4 +286,76 @@ private function renderAsHex(): string
...
+                break;
+            }
+            if (!$value instanceof Size) {
```
oliverklee, line 320: Just for the reader:

```suggestion
            if (!($value instanceof Size)) {
```

src/Value/Color.php
```diff
@@ -282,4 +286,76 @@ private function renderAsHex(): string
...
+        return $hasPercentage && $hasNumber;
+    }
+
+    /**
+     * @return non-empty-string
+     */
+    private function renderInModernSyntax(OutputFormat $outputFormat): string
+    {
+        // Maybe not yet without alpha, but will be...
+        $componentsWithoutAlpha = $this->aComponents;
```
oliverklee, line 345: I'm confused … according to the comment, the variable could contain alpha, but according the to the name, it could.

src/Value/Color.php
```diff
@@ -282,4 +286,76 @@ private function renderAsHex(): string
...
+        \end($componentsWithoutAlpha);
+        if (\key($componentsWithoutAlpha) === 'a') {
+            $alpha = $this->aComponents['a'];
+            unset($componentsWithoutAlpha['a']);
+        }
```
oliverklee, line 350: I'd still prefer something along those lines (not tested):

```php
$alpha = $this->aComponents['a'] ?? null;
$components = $this->aComponents;
unset($components['alpha']);
```

oliverklee (changes requested):

src/Value/Color.php
JakeQZ: That's a good idea.  Though it's "hex" rather than "legacy" syntax in this case ("legacy" is with commas between the arguments; "modern" is with spaces between the argumens but a slash separating the alpha argument; "hex" is the notation that begins `#` - it's a complete mess IMO, but those are the standards).

But it's beyond the scope of this PR, so I've added #852.

src/Value/Color.php
JakeQZ: > I propose we move the (static) variable and the `in_array` to a dedicated, well-named method.

Done.

> Also, please generally let's not use `static` variables as they keep state across instances and are a pain to test. In this case, I think a (normal) local variable or a constant would be fine.

Point taken.  In the new method it's not `static`.  For the time being, until these changes have 'settled down', I'd prefer not to introduce a class constant, only to remove it later.

src/Value/Color.php
JakeQZ: Agree.  Done.  (It looks odd having the `!` biddying-up to the variable, and then spaces as separators after.)

src/Value/Color.php
JakeQZ: See https://github.com/MyIntervals/PHP-CSS-Parser/pull/840#discussion_r1933072212

I added the comment to explain that the variable may not match its description until the `unset` was done.

Perhaps the comment can be improved, or we can rework this section of code.

src/Value/Color.php
JakeQZ: That doesn't work.  The initial test needs to be that `a` is the last key in the array, because the order of the keys is important.  As previously said, 'lab` does not have an alpha component, but would have an `a` value in the array (when supported explicitly) .

Perhaps some additional commenting is the way forward here.

JakeQZ:
I've addressed most issues, but there are still two unresolved that require your feedback (see the comments).

oliverklee (approved):

oliverklee:
By the way, I see this in the commit messages of your PRs:

```
Co-authored-by: Jake Hotson <jake.github@qzdesign.co.uk>
```

Are you using two different email addresses? And are both registered in your GitHub account?

JakeQZ:
> By the way, I see this in the commit messages of your PRs:
> 
> ```
> Co-authored-by: Jake Hotson <jake.github@qzdesign.co.uk>
> ```
> 
> Are you using two different email addresses? And are both registered in your GitHub account?

I've edited some PR commit messages to have this email address, because I don't want my main one listed on GitHub.  I'm getting increasing amounts of spam arising from email addresses being scraped from GitHub.  Ideally I would change my GitHib account to use this address, but I suspect that would be some hassle.

oliverklee:
Have you tried adding it as an additional email address?

JakeQZ:
> Have you tried adding it as an additional email address?

Didn't know that was possible.  Thanks.

PR #841, opened 2025-01-28
[TASK] Add native type declarations for anonymous functions

oliverklee:
Also make then static if possible.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -413,7 +413,7 @@ abstract class CSSList implements Renderable, Commentable
             $oNextLevel = $oOutputFormat->nextLevel();
         }
         foreach ($this->aContents as $oContent) {
-            $sRendered = $oOutputFormat->safely(function () use ($oNextLevel, $oContent) {
+            $sRendered = $oOutputFormat->safely(function () use ($oNextLevel, $oContent): string {
                 return $oContent->render($oNextLevel);
             });
             if ($sRendered === null) {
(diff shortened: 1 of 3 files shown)


src/CSSList/CSSList.php
```diff
@@ -413,7 +413,7 @@ protected function renderListContents(OutputFormat $oOutputFormat)
             $oNextLevel = $oOutputFormat->nextLevel();
         }
         foreach ($this->aContents as $oContent) {
-            $sRendered = $oOutputFormat->safely(function () use ($oNextLevel, $oContent) {
+            $sRendered = $oOutputFormat->safely(function () use ($oNextLevel, $oContent): string {
```
JakeQZ, line 416: Could this anonymous function also be made `static`?

src/RuleSet/RuleSet.php
```diff
@@ -269,7 +269,7 @@ protected function renderRules(OutputFormat $oOutputFormat)
         $oNextLevel = $oOutputFormat->nextLevel();
         foreach ($this->aRules as $aRules) {
             foreach ($aRules as $rule) {
-                $sRendered = $oNextLevel->safely(function () use ($rule, $oNextLevel) {
+                $sRendered = $oNextLevel->safely(function () use ($rule, $oNextLevel): string {
```
JakeQZ, line 272: And this?

JakeQZ (changes requested):
I may be wrong, but think a couple more closures can be made `static`, while we're here.

src/CSSList/CSSList.php
oliverklee: It can indeed. (I first thought that the `use` made `static` impossible, but was mistaken. Fixed.

src/RuleSet/RuleSet.php
oliverklee: Done.

JakeQZ (approved):

PR #847, opened 2025-01-28
[TASK] Add accessor tests for `OutputFormat` (part 1)

oliverklee:
Currently, this class uses magic getters and setters. These
should be proper methods instead.

But first, we should have tests for the accessors. This is
part 1.

### DIFF
--- a/tests/Unit/OutputFormatTest.php
+++ b/tests/Unit/OutputFormatTest.php
@@ -1,0 +1,307 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\OutputFormat;
+
+/**
+ * @covers \Sabberworm\CSS\OutputFormat
+ */
+final class OutputFormatTest extends TestCase
+{
+    /**
+     * @var OutputFormat
+     */
+    private $subject;
+
+    protected function setUp(): void
+    {
+        $this->subject = new OutputFormat();
+    }
+
+    /**
+     * @test
+     */
+    public function getStringQuotingTypeInitiallyReturnsDoubleQuote(): void
+    {
+        self::assertSame('"', $this->subject->getStringQuotingType());
+    }
+
+    /**
+     * @test
+     */
+    public function setStringQuotingTypeSetsStringQuotingType(): void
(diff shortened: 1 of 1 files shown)


JakeQZ (changes requested):
Brilliant.

I tried experimentally ramping up PHPStan to level 2 and got loads of warnings re using the magic `__call()`.  If we can resolve this and increase the PHPStan checking level, it will give us more confidence about future changes.

However, I think some of the test cases could be made more realistic.

tests/Unit/OutputFormatTest.php
```diff
@@ -0,0 +1,307 @@
...
+    {
+        self::assertSame('"', $this->subject->getStringQuotingType());
+    }
+
+    /**
+     * @test
+     */
+    public function setStringQuotingTypeSetsStringQuotingType(): void
+    {
+        $value = 'x';
```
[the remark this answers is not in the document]
JakeQZ: Or maybe have a `dataProvider` to provide more than one quote type.

oliverklee:
> Or maybe have a `dataProvider` to provide more than one quote type.

The accessor tests are basically for two things:

1. which default value the getter returns
2. that setting (and then getting) a value using the setter works (for this to be checkable, we need to set value that is different to the default value); for the benefit of the human reader, the used value should vaguely make sense for this property

Checking that different quote types work is out of scope - I see it more in scope of the tests for the CSS elements that use this particular property.

tests/Unit/OutputFormatTest.php
```diff
@@ -35,7 +35,7 @@ public function getStringQuotingTypeInitiallyReturnsDoubleQuote(): void
      */
     public function setStringQuotingTypeSetsStringQuotingType(): void
     {
-        $value = 'x';
+        $value = '"';
```
JakeQZ, line 38: Isn't this now the same as the default value?

JakeQZ (changes requested):
`setStringQuotingTypeSetsStringQuotingType()` appears to now be setting the same as the default value.

tests/Unit/OutputFormatTest.php
oliverklee: It is indeed! Changed to a single quote now.

JakeQZ (approved):

PR #851, opened 2025-01-29
[CLEANUP] Avoid Hungarian notation in `CSSBlockList` (part 2)

oliverklee:
Part of #756

### DIFF
--- a/src/CSSList/CSSBlockList.php
+++ b/src/CSSList/CSSBlockList.php
@@ -105,11 +105,11 @@ abstract class CSSBlockList extends CSSList
                     $result[] = $selector;
                 } else {
                     $comparator = '===';
-                    $specificitySearchResults = \explode(' ', $specificitySearch);
-                    $targetSpecificity = $specificitySearchResults[0];
-                    if (\count($specificitySearchResults) > 1) {
-                        $comparator = $specificitySearchResults[0];
-                        $targetSpecificity = $specificitySearchResults[1];
+                    $expressionParts = \explode(' ', $specificitySearch);
+                    $targetSpecificity = $expressionParts[0];
+                    if (\count($expressionParts) > 1) {
+                        $comparator = $expressionParts[0];
+                        $targetSpecificity = $expressionParts[1];
                     }
                     $targetSpecificity = (int) $targetSpecificity;
                     $selectorSpecificity = $selector->getSpecificity();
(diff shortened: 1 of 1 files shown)


src/CSSList/CSSBlockList.php
```diff
@@ -96,42 +96,42 @@ protected function allValues(
...
                     $result[] = $selector;
                 } else {
-                    $sComparator = '===';
-                    $aSpecificitySearch = \explode(' ', $specificitySearch);
-                    $iTargetSpecificity = $aSpecificitySearch[0];
-                    if (\count($aSpecificitySearch) > 1) {
-                        $sComparator = $aSpecificitySearch[0];
-                        $iTargetSpecificity = $aSpecificitySearch[1];
+                    $comparator = '===';
+                    $specificitySearchResults = \explode(' ', $specificitySearch);
```
JakeQZ, line 108: I think this should be named with something like 'parts', 'elements' or 'components'.  The input string contains either a number (e.g. "1"), or an operator and a number (e.g. ">= 1").  The result goes into `$result`.

JakeQZ (changes requested):
I think renaming `$aSpecificitySearch` to `$specificitySearchResults` is not accurate.

src/CSSList/CSSBlockList.php
oliverklee: Maybe `$expressionParts`?

JakeQZ: Yes I think that works.  It's only used locally in that small section of code.

src/CSSList/CSSBlockList.php
```diff
@@ -96,42 +96,42 @@ protected function allValues(
...
+                        $comparator = $specificitySearchResults[0];
+                        $targetSpecificity = $specificitySearchResults[1];
                     }
-                    $iTargetSpecificity = (int) $iTargetSpecificity;
-                    $iSelectorSpecificity = $selector->getSpecificity();
-                    $bMatches = false;
-                    switch ($sComparator) {
+                    $targetSpecificity = (int) $targetSpecificity;
+                    $selectorSpecificity = $selector->getSpecificity();
+                    $hasMatches = false;
```
JakeQZ, line 116: Apologies, I missed this in the previous review.

I don't think "has" is correct here.  The equivalent linguistic phrase is "it matches" rather than "it has matches"...

JakeQZ (changes requested):
I spotted another naming issue I missed when I reviewed previously.

src/CSSList/CSSBlockList.php
oliverklee: Better late than never. ;-) I've come up with a different name and repushed.

sabberworm: Maybe `comparatorMatched`?

JakeQZ (approved):

PR #854, opened 2025-01-30
[CLEANUP] Avoid Hungarian notation for `aContents` and items

oliverklee:
Part of #756


JakeQZ (approved):
`$item` is as good generally as anything I can think of, and matches the DocBlock comment in various cases  `$listItem` also works where you've used it.  Good renaming :)

PR #859, opened 2025-01-31
[FEATURE] Support `none` as color function component value

JakeQZ:
Note that this keyword is only permitted in the "modern" syntax (and thus functions using it must be rendered in that syntax).


oliverklee (approved):
LGTM.

PR #862, opened 2025-02-01
[TASK] Add accessor tests for `OutputFormat` (part 2)

oliverklee:
Currently, this class uses magic getters and setters. These should be proper methods instead.

But first, we should have tests for the accessors. This is part 2.


JakeQZ (changes requested):
One test method is missing `Initially` from its name.

Also wondering about `'    '` as separator test values not being particularly real-world.

JakeQZ (approved):

PR #866, opened 2025-02-02
[BUGFIX] Drop test that passes in the wrong type

oliverklee:
`OutputFormat::spaceAfterListArgumentSeparator` is expected to be a string, not an array.

This was discovered by adding dedicated, strictly typed property accessors for `OutputFormat`.

### DIFF
--- a/tests/OutputFormatTest.php
+++ b/tests/OutputFormatTest.php
@@ -99,23 +99,6 @@ EOT;
     /**
      * @test
      */
-    public function spaceAfterListArgumentSeparatorComplex(): void
-    {
-        self::assertSame(
-            '.main, .test {font: italic normal bold 16px/1.2 "Helvetica",	Verdana,	sans-serif;background: white;}'
-            . "\n@media screen {.main {background-size: 100% 100%;font-size: 1.3em;background-color: #fff;}}",
-            $this->document->render(OutputFormat::create()->setSpaceAfterListArgumentSeparator([
-                'default' => ' ',
-                ',' => "\t",
-                '/' => '',
-                ' ' => '',
-            ]))
-        );
-    }
-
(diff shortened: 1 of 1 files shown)


JakeQZ (changes requested):
> `OutputFormat::spaceAfterListArgumentSeparator` is expected to be a string, not an array.

I think it is expected to be a string or an array.  If an array, the keys are list separators and the values are the space string to insert after them, the the first item in the array (also) serving as the default space separator.  See `OutputFormat::createPretty()` for examples and `OutputFormmatter::space()` for implementation.

oliverklee:
> I think it is expected to be a string or an array.

Yes, that's how it currently works (even if it violates the type annotation). This behavior is not documented, and I'd like to get rid of it.

JakeQZ:
> > I think it is expected to be a string or an array.
> 
> Yes, that's how it currently works (even if it violates the type annotation). This behavior is not documented, and I'd like to get rid of it.

It's documented by this:

> To see what you can do with output formatting, look at the tests in `tests/OutputFormatTest.php`.

It's also part of the public API.  So if we remove it we'll first need to mark it as deprecated.

JakeQZ:
If we do want to remove the functionality, we'd first need to create the dedicated (non-majic) methods (which @oliverklee I see you're on a path towards, e.g. with #867), whilst retaining current functionality, and correcting the type definitions to include `array` as a possibility.

Then it would be possibke to deprecate the array parameter option to have different spacing depending on list separator, with e.g. `@depracated From 10.0 the $space parameter can no longer be an array`.

Having done the first part, it would be extra effort to remove the feature.  I think we may as well retain it.  @oliverklee, @sabberworm, WDYT?

PR #868, opened 2025-02-02

oliverklee:


tests/Unit/OutputFormatTest.php
```diff
@@ -684,4 +685,181 @@ public function setIndentationLevelProvidesFluentInterface(): void
...
+
+    /**
+     * @test
+     */
+    public function levelReturnsIndentationLevel(): void
+    {
+        $value = 4;
+        $this->subject->setIndentationLevel($value);
+
+        self::assertSame($value, $this->subject->level());
```
JakeQZ, line 844: Should this be calling `getIndentationLevel()`, since `level()` was deprecated in #870.

JakeQZ (changes requested):
One of the tests is using a method that has just been deprecated.

tests/Unit/OutputFormatTest.php
oliverklee: No, this test is specific to the (deprecated) method. I'll remove the test together with the method.

(`setIndentationLevel` itself has a dedicated test.)

JakeQZ: Ah, yes, I overlooked the test method name.

JakeQZ (approved):

PR #875, opened 2025-02-02
[TASK] Add native type declarations to `OutputFormat`

oliverklee:
Also fix some type annotations.

### DIFF
--- a/src/OutputFormat.php
+++ b/src/OutputFormat.php
@@ -170,11 +170,9 @@ class OutputFormat
     public function __construct() {}
 
     /**
-     * @param string $sName
-     *
-     * @return string|null
+     * @return string|int|bool|null
      */
-    public function get($sName)
+    public function get(string $sName)
     {
         $aVarPrefixes = ['a', 's', 'm', 'b', 'f', 'o', 'c', 'i'];
         foreach ($aVarPrefixes as $sPrefix) {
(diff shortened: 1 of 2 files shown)


src/OutputFormat.php
```diff
@@ -170,11 +170,9 @@ class OutputFormat
     public function __construct() {}
 
     /**
-     * @param string $sName
-     *
-     * @return string|null
+     * @return string|int|bool|null
```
JakeQZ, line 173: Possibly even `array`, but we'll see (and that's beyond the scope of this)...

src/OutputFormat.php
```diff
@@ -258,10 +256,7 @@ public function indentWithSpaces(int $numberOfSpaces = 2): self
         return $this->setIndentation(\str_repeat(' ', $numberOfSpaces));
     }
 
-    /**
-     * @return OutputFormat
-     */
-    public function nextLevel()
+    public function nextLevel(): self
```
JakeQZ, line 259: IIUC, we are only PHPDoccing `@return $this fluent interface` when the method returns itself, as opposed to when it returns some relative of the same genus or class.

JakeQZ (approved):
LGTM.  Added some questions as comments, but which I think are beyond the scope of approving this.

src/OutputFormat.php
oliverklee: Yes, correct.

PR #877, opened 2025-02-04
[TASK] Raise PHPStan to level 2

oliverklee:


JakeQZ (approved):
I guess it doesn't matter whether we resolve the issues reported before or after bumping the checking level.

oliverklee:
I actually prefer resolving them after bumping the level, because seeing my changes remove an issue feels really good to me. :muscle:

PR #884, opened 2025-02-08
[CLEANUP] Add more type assertions in the tests

oliverklee:


JakeQZ (approved):
In the tests we can assert the type because it's what we're expecting given the input test data.

PR #885, opened 2025-02-08
[TASK] Only allow `string` for some `OutputFormat` properties

oliverklee:
`sBeforeAfterListArgumentSeparator` and
`sSpaceAfterListArgumentSeparator` now can only hold strings, and no longer arrays (which was deprecated in V8.8.0).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Please also have a look at our
 
 ### Changed
 
+- Only allow `string` for some `OutputFormat` properties (#885)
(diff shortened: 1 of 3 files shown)


JakeQZ (approved):

JakeQZ (changes requested):
Actually, I think we need a changelog entry.  Otherwise fine.

oliverklee:
Rebased, added a changelog entry, and repushed.

JakeQZ (approved):

PR #888, opened 2025-02-09
[TASK] Make constructors of CSS element classes `@internal`

### LINKED ISSUE #858: Make constructors of CSS elements `@internal`
This will allow us to refactor fearlessly.

oliverklee:
These classes are intended to be instantiated only from within this library.

Fixes #858

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Please also have a look at our
 
 ### Changed
 
+- Make constructors of CSS element classes `@internal` (#888)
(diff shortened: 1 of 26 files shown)


JakeQZ (commented):
I don't think this is a valid change.  The library allows manipulation of the parsed CSS, including adding to it, which would require constructing new objects of various types - see https://github.com/MyIntervals/PHP-CSS-Parser#manipulation.

oliverklee:
Ah, you're right … closing.

PR #890, opened 2025-02-09

oliverklee:


JakeQZ (approved):
Strangely, coverage remained the same.  I guess the code paths were already being exercised by tests, though the results not so strictly checked.

PR #892, opened 2025-02-09
[BUGFIX] Fix return type annotation of `Value::isSize()`

oliverklee:


src/Value/Size.php
```diff
@@ -186,7 +186,7 @@ public function isColorComponent()
     /**
      * Returns whether the number stored in this Size really represents a size (as in a length of something on screen).
      *
-     * @return false if the unit an angle, a duration, a frequency or the number is a component in a Color object.
```
JakeQZ, line 189: I think just change this comment to not involve a type annotation - i.e. "Returns false if the unit [is]..." (and add the missing "is") - so the meaning is retained.

CHANGELOG.md
```diff
@@ -45,6 +45,7 @@ Please also have a look at our
 
 ### Fixed
 
+- Fix return type annotation of `Value::isSize()` (#892)
```
JakeQZ, line 48: Not sure if we're just changing a DocBlock comment this warrants a changelog entry.

JakeQZ (changes requested):
I think the comment can just be changed to use the word "Returns" instead of `@return`.

JakeQZ (approved):

PR #893, opened 2025-02-09

oliverklee:


tests/Unit/OutputFormatTest.php
```diff
@@ -1025,4 +1035,133 @@ public function createCompactReturnsInstanceWithRenderCommentsDisabled(): void
 
         self::assertFalse($newInstance->getRenderComments());
     }
+
+    /**
+     * @test
+     */
+    public function createPrettyReturnsNewOutputFormatInstance(): void
```
JakeQZ, line 1042: I think "New" is redundant in the test method name, but it's consistent with `createCompactReturnsNewOutputFormatInstance`, which I signed off :shrug:

tests/Unit/OutputFormatTest.php
```diff
@@ -1025,4 +1035,133 @@ public function createCompactReturnsInstanceWithRenderCommentsDisabled(): void
...
+    {
+        self::assertInstanceOf(OutputFormat::class, OutputFormat::createPretty());
+    }
+
+    /**
+     * @test
+     */
+    public function createPrettyCalledTwoTimesReturnsDifferentInstances(): void
```
JakeQZ, line 1050: I'm thinking (and agreeing) "TwoTimes" is better than "Twice" for non-native-English speaking coders.

tests/Unit/OutputFormatTest.php
```diff
@@ -1025,4 +1035,133 @@ public function createCompactReturnsInstanceWithRenderCommentsDisabled(): void
...
+    {
+        $newInstance = OutputFormat::createPretty();
+
+        self::assertSame("\n", $newInstance->getSpaceAfterBlocks());
+    }
+
+    /**
+     * @test
+     */
+    public function createPrettyReturnsInstanceWithSpaceAfterRuleNameSetSpace(): void
```
JakeQZ, line 1121:
```suggestion
    public function createPrettyReturnsInstanceWithSpaceAfterRuleNameSetToSpace(): void
```

tests/Unit/OutputFormatTest.php
```diff
@@ -1025,4 +1035,133 @@ public function createCompactReturnsInstanceWithRenderCommentsDisabled(): void
...
+    }
+
+    /**
+     * @test
+     */
+    public function createPrettyReturnsInstanceWithSpaceAfterListArgumentSeparatorsSetToSpaceForComma(): void
+    {
+        $newInstance = OutputFormat::createPretty();
+
+        self::assertSame([',' => ' '], $newInstance->getSpaceAfterListArgumentSeparators());
```
JakeQZ, line 1155: Should this just test that the key `,` has the value {space} - instead of the whole array?

JakeQZ (changes requested):
I like the way these tests are narrowing-down expected behaviour.

A couple of minor issues, and some comments beyond the scope of this PR.

tests/Unit/OutputFormatTest.php
JakeQZ: Or maybe the test method should be renamed to indicate it's checking the whole array.

oliverklee:
Updated and repushed.

JakeQZ (approved):
WFM now

PR #894, opened 2025-02-09
[TASK] Deprecate method forwarding `OutputFormat` to `OutputFormatter`

oliverklee:
Instead, the corresponding method on the formatter should be called directly.

This increases type safety and helps static code analysis. Also, it makes the code easier to understand.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,8 @@ Please also have a look at our
 
 ### Deprecated
 
+- Deprecate magic method forwarding from `OutputFormat` to `OutputFormatter`
+  (#894)
 - Deprecate greedy calculation of selector specificity (#1018)
 - Deprecate `__toString()` (#1006)
 - `OutputFormat` properties for space around list separators as an array (#880)
--- a/src/OutputFormat.php
+++ b/src/OutputFormat.php
@@ -244,6 +244,7 @@ class OutputFormat
     public function __call(string $sMethodName, array $aArguments)
     {
         if (\method_exists(OutputFormatter::class, $sMethodName)) {
+            // @deprecated since 8.8.0, will be removed in 9.0.0. Call the method on the formatter directly instead.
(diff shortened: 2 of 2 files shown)


JakeQZ (changes requested):
I think we need to provide chaining methods before we can do this.

If you edit `__call()` to throw an exception in all cases, there are many test failures.

The chaining methods required are clearly-defined by `OutputFormatter`.  There aren't *that* many.

I think this would be easier and more preferable than rewriting the tests and changing the API - bear in mind that all users would have to expend similar effort rewriting their code as we would rewriting the tests, if we were to change the API.

JakeQZ:
> The chaining methods required are clearly-defined by `OutputFormatter`.

When they are all in place, `__call()` can simply be removed.

oliverklee:
I've updated the calls in #898.

JakeQZ:
> > > I think this would be easier and more preferable than rewriting the tests and changing the API - bear in mind that all users would have to expend similar effort rewriting their code as we would rewriting the tests, if we were to change the API.
> > 
> > 
> > I consider the `OutputFormatter` to be internal to our library (#896), and hence this change should have no effect on users of our library.
> 
> I agree, I think. Which is all the more reason to add these methods to `OutputFormat` so that users don't need to access the underlying `OutputFormatter`, whether that be by obtaining the object or using `__cal()`.

... unless these methods are all meant for internal use.  Will need to check...

JakeQZ:
> I've updated the calls in #898.

I think that makes it clear.  These are all (only) used by the internal `render` methods, as a convenience to avoid the extra `getFormatter` call, and are not intended to be part of the public API.  @sabberworm, is that correct?

oliverklee:
Ping @sabberworm - we need your input here in order to proceed.

oliverklee:
Please let's get this merged. @sabberworm And input from your side?

@JakeQZ If there's no input from @sabberworm, please let's move this forward nonetheless.

sabberworm (approved):

Issue #895, opened 2025-02-10
Dependency loop `OutputFormat`/`OutputFormatter`


oliverklee:
Both have references to an instance of each other. Dependencies should go only one way, though, and loops are a really bad thing.

JakeQZ:
We're talking about class dependencies, rather than object dependencies.  A linked-list in C++ (before stdlib) would have loops in the objects, but they would all be of the same class.

As I understand, `OutputFormat` is where the settings are, and `OutputFormatter` is the worker.  So `OutputFormatter` should have a reference to `OutputFormat`, but there shouldn't be a need for one the other way.

PR #896, opened 2025-02-10
[TASK] Mark `OutputFormatter` as `@internal`

oliverklee:
This class should only be used for formatting CSS from within this library. It is not intended to be called from outside.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@ Please also have a look at our
 
 ### Changed
 
+- Mark `OutputFormatter` as `@internal` (#896)
 - Make `Selector` a `Renderable` (#1017)
 - Mark `Selector::isValid()` as `@internal` (#1037)
 - Mark parsing-related methods of most CSS elements as `@internal` (#908)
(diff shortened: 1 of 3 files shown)


JakeQZ (commented):
I think so.

@sabberworm, would you agree?

JakeQZ:
Actually maybe not.  See https://github.com/MyIntervals/PHP-CSS-Parser/issues/895#issuecomment-2649664498.

It might not be used from the outside, but may need to be constructed and passed around, depending on how we resolve the circular dependency.

oliverklee:
I've proposed a path forward in #907.

oliverklee:
@sabberworm ping

JakeQZ:
Marking as draft pending #907,

oliverklee:
Ping @JakeQZ @sabberworm - I'd like to get this merged soon-ish so I can reduce the number of open PRs created by me (which reduces the amount of rebasing I need to do).

JakeQZ:
Yes, I think currently users of this library provide an `OutputFormat` method which internally has a reference to an `OutputFormatter` which does the work and maintains state.  I guess it's analogous to a `Car` and an `Engine`.  The car has an engine rather than the other way round, and the user drives the car without needing to touch the engine directly.  The same car can visit the various CSS elements.

Also, this is consistent with the proposed change in #898.  So 👍 from me now.

PR #897, opened 2025-02-10
[BUGFIX] Add missing imports for `@throws` annotations

oliverklee:


JakeQZ (approved):
Impressed that Stan managed to identify the FQN of the exception class without the `use` statement.

oliverklee:
> Impressed that Stan managed to identify the FQN of the exception class without the `use` statement.

It didn't - that's why this change fixes some PHPStan warnings. ;-)

JakeQZ:
> > Impressed that Stan managed to identify the FQN of the exception class without the `use` statement.
> 
> It didn't - that's why this change fixes some PHPStan warnings. ;-)

You're right, it just assumed FQN from the current namespace.

PR #898, opened 2025-02-10
[TASK] Avoid magic method forwarding in `OutputFormat`

oliverklee:


oliverklee:
We should cover the changed lines with tests first, though.

oliverklee:
Sorry, I seem to have missed your comment. I've added a changelog entry now and repushed.

JakeQZ (approved):

PR #900, opened 2025-02-10
[BUGFIX] Add missing imports for `@throws` annotations

oliverklee:
This is the V8.x backport of #897.


JakeQZ (approved):
Finally something I can readily approve :)

PR #901, opened 2025-02-10
[TASK] Mark `OutputFormat::nextLevel()` as `@internal`

oliverklee:
This method is used internally for rendering the CSS, and it is not intended to be called from outside this library.


JakeQZ (approved):
This functionality perhaps belongs in `OutputFormatter`.

PR #908, opened 2025-02-11
[TASK] Mark many parsing-related methods as `@internal`

oliverklee:
The only parsing method that is expected to be called from outside of this library is `Parser::parse()`.


JakeQZ (approved):
Agree.  I wou;dn't expect anyone to be attempting to parse the constructs in isolation.  And not even `Document::parse()` needs to be part of the public API.

PR #911, opened 2025-02-11
[TASK] Add tests for the exceptions

oliverklee:
Part of #757

### DIFF
--- a/tests/Unit/Parsing/OutputExceptionTest.php
+++ b/tests/Unit/Parsing/OutputExceptionTest.php
@@ -6,12 +6,21 @@ namespace Sabberworm\CSS\Tests\Unit\Parsing;
 
 use PHPUnit\Framework\TestCase;
 use Sabberworm\CSS\Parsing\OutputException;
+use Sabberworm\CSS\Parsing\SourceException;
 
 /**
  * @covers \Sabberworm\CSS\Parsing\OutputException
  */
 final class OutputExceptionTest extends TestCase
 {
+    /**
+     * @test
+     */
+    public function extendsSourceException(): void
+    {
+        self::assertInstanceOf(SourceException::class, new OutputException('foo'));
+    }
+
     /**
      * @test
      */
--- a/tests/Unit/Parsing/OutputExceptionTest.php
+++ b/tests/Unit/Parsing/OutputExceptionTest.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit\Parsing;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Parsing\OutputException;
+
+/**
+ * @covers \Sabberworm\CSS\Parsing\OutputException
+ */
+final class OutputExceptionTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function getMessageReturnsMessageProvidedToConstructor(): void
+    {
+        $message = 'The cake is a lie.';
+        $exception = new OutputException($message);
+
+        self::assertStringContainsString($message, $exception->getMessage());
+    }
+
+    /**
+     * @test
+     */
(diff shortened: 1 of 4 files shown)


tests/Unit/Parsing/UnexpectedEOFExceptionTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $lineNumber = 17;
+        $exception = new UnexpectedEOFException('expected', 'found', 'literal', $lineNumber);
+
+        self::assertSame($lineNumber, $exception->getLineNo());
+    }
+
+    /**
+     * @test
+     */
+    public function getMessageForLineNumberProvidedIncludesMessage(): void
```
JakeQZ, line 39: Ditto.

tests/Unit/Parsing/UnexpectedTokenExceptionTest.php
```diff
@@ -0,0 +1,168 @@
...
+        $lineNumber = 17;
+        $exception = new UnexpectedTokenException('expected', 'found', 'literal', $lineNumber);
+
+        self::assertSame($lineNumber, $exception->getLineNo());
+    }
+
+    /**
+     * @test
+     */
+    public function getMessageForLineNumberProvidedIncludesMessage(): void
```
JakeQZ, line 39: Ditto.

JakeQZ (changes requested):
I like the running theme of amusing and bizarre messages used in the tests.  (Some hosts give me tea in the morning when coffee is expected; YMMV.)

I think one of the test methods (repeated 4 times) is conflated, and that there should be two separate test methods: one for the message, and one for the line number.

I'm also wondering about the duplication.  I think it's right to also test that the extended classes' behaviour match the base (where that is expected).  But I wonder if it's possible to have a base class or trait for the `TestCase` to provide the common tests - if that would work with PHPUnit which I know relies on `@test` in the DocBlock.

tests/Unit/Parsing/OutputExceptionTest.php
```diff
@@ -0,0 +1,67 @@
...
+        $lineNumber = 17;
+        $exception = new OutputException('foo', $lineNumber);
+
+        self::assertSame($lineNumber, $exception->getLineNo());
+    }
+
+    /**
+     * @test
+     */
+    public function getMessageForLineNumberProvidedIncludesMessage(): void
```
JakeQZ, line 50: GitHub lost my comment here.  The various 'dittos' refer to this.

I think this is conflating two tests, one that the line number is included, and one that the message is included when a line number is provided.  Also the name is confusing.  I'd suggest
- `getMessageWithLineNumberProvidedIncludesMessage` and
- `getMessageWithLineNumberProvidedIncludesLineNumber`.

JakeQZ (commented):
I added back (roughly) a key comment that GitHub lost.

oliverklee:
> I like the running theme of amusing and bizarre messages used in the tests. (Some hosts give me tea in the morning when coffee is expected; YMMV.)

:heart:

tests/Unit/Parsing/OutputExceptionTest.php
oliverklee: Thanks, I mixed up the tests names indeed. Fixed.

As far as the message provided to the constructor is concerned, we already have the `getMessageReturnsMessageProvidedToConstructor` tests for cases where the exception actually uses the message provided to the constructor.

JakeQZ: True, but we don't have a test to confirm the message when a line number argument is passed.

JakeQZ:
> As far as the behavior is concerned, I'd prefer to have the tests duplicated. This makes the test easier to understand, and it will allow us to refactor more easily.

Duplicated code shrieks at me.  It's much less maintainable,  It's fine for now, but I would like to come up with a way of avoiding it.

JakeQZ (approved):
Added #919 to cover possible clean-up.

PR #912, opened 2025-02-11
[TASK] Move some tests for `Comment` to functional tests

oliverklee:
Move the tests that rely on another class (`OutputFormat`) (or might in the future).

Part of #757.

### DIFF
--- a/tests/Functional/Comment/CommentTest.php
+++ b/tests/Functional/Comment/CommentTest.php
@@ -49,7 +49,7 @@ final class CommentTest extends TestCase
 
         $subject->setComment($comment);
 
-        self::assertSame('/*' . $comment . '*/', $subject->render(new OutputFormat()));
+        self::assertSame('/*' . $comment . '*/', $subject->render(OutputFormat::createCompact()));
     }
 
     /**
--- a/tests/Functional/Comment/CommentTest.php
+++ b/tests/Functional/Comment/CommentTest.php
@@ -1,0 +1,67 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Functional\Comment;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Comment\Comment;
+use Sabberworm\CSS\OutputFormat;
+
+/**
+ * @covers \Sabberworm\CSS\Comment\Comment
+ */
+final class CommentTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function toStringRendersCommentEnclosedInCommentDelimiters(): void
+    {
+        $comment = 'There is no spoon.';
+        $subject = new Comment();
+
+        $subject->setComment($comment);
+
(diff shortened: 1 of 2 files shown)


tests/Functional/Comment/CommentTest.php
```diff
@@ -0,0 +1,67 @@
...
+     * @test
+     */
+    public function renderWithCompactOutputFormatRendersCommentEnclosedInCommentDelimiters(): void
+    {
+        $comment = 'There is no spoon.';
+        $subject = new Comment();
+
+        $subject->setComment($comment);
+
+        self::assertSame('/*' . $comment . '*/', $subject->render(new OutputFormat()));
```
JakeQZ, line 52:
```suggestion
        self::assertSame('/*' . $comment . '*/', $subject->render(OutputFormat::createCompact()));
```

tests/Functional/Comment/CommentTest.php
```diff
@@ -0,0 +1,67 @@
...
+     * @test
+     */
+    public function renderWithPrettyOutputFormatRendersCommentEnclosedInCommentDelimiters(): void
+    {
+        $comment = 'There is no spoon.';
+        $subject = new Comment();
+
+        $subject->setComment($comment);
+
+        self::assertSame('/*' . $comment . '*/', $subject->render(new OutputFormat()));
```
JakeQZ, line 65:
```suggestion
        self::assertSame('/*' . $comment . '*/', $subject->render(OutputFormat::createPretty()));
```

JakeQZ (changes requested):
The tests for 'pretty' and 'compact' should actually use the applicable creation methods.

Perhaps there should also be a test for the `create` method, as well as construction with `new`.

oliverklee:
Done and repushed.

JakeQZ (approved):
Virgin was an interesting choice for the test method name, and slightly amusing, but works :)

oliverklee:
> Virgin was an interesting choice for the test method name, and slightly amusing, but works :)

I actually was struggling with the name a bit. A close contender was "newly minted". Any other suggestions (from a native BE speaker)?

JakeQZ:
> > Virgin was an interesting choice for the test method name, and slightly amusing, but works :)
> 
> I actually was struggling with the name a bit. A close contender was "newly minted". Any other suggestions (from a native BE speaker)?

Coincidentally, I just listened to a [programme about cooking oils](https://www.bbc.co.uk/sounds/play/m00274y5).  You might not be able to access it outside the UK.  14 minutes in the 'expert' talks about olive oil, and says: "Extra virgin overall means it's as close to the natural state as possible."

So I think you've hit the nomeclature nail directly on the head here, and the naming cannot be improvied upon.  Well done.  (Virgin does not always imply lack of promiscuity, particularly when applied to inanimate objects.)

Freshly-minted has a strong association with the bank prressing coins.  It would not have worked as well, but would have been acceptably whimsical :))

Issue #90, opened 2015-05-20
[Right to Left] and [Left to Right] conversion


Natshah:
We do need to have a class to work around Conversion.
RTL to LTR .. or LTR to RTL

devmsh:
:) I already built one on top of PHP-CSS-Parser

You can take a look here: https://github.com/devmsh/RTLer

sabberworm:
Great, thanks.

PR #914, opened 2025-02-13
[DOCS] Fix a PR id in the changelog

oliverklee:


JakeQZ (approved):
I've just noticed the backport has `#90` in the changelog, so that will need correcting too :/

oliverklee:
> I've just noticed the backport has #90 in the changelog, so that will need correcting too :/

Fixed in #920.

PR #925, opened 2025-02-14
[TASK] Drop redundant `OutputException` constructor

oliverklee:


JakeQZ (commented):
There is an issue with PHP and undefined constructors, which is why I always provide an empty constructor in the base class.  I can't recall the details.  I don't think that's a problem here, but will need to check.

oliverklee:
In this case, the direct parent class `SourceException` has a constructor.

JakeQZ (approved):
This is fine: `SourceException` has a constructor.  This might have already been covered by #932...

JakeQZ:
> This might have already been covered by #932...

That was for a different class (`PrimitiveValue`), but a likewise change.

> In this case, the direct parent class `SourceException` has a constructor.

I missed that comment before commenting on having reached the same conclusion :)

PR #932, opened 2025-02-15
[TASK] Drop redundant constructor code

oliverklee:
- use default values for properties instead of setting them in the constructor
- drop constructors that are redundant to the constructor of the parent class

### DIFF
--- a/src/CSSList/CSSBlockList.php
+++ b/src/CSSList/CSSBlockList.php
@@ -21,14 +21,6 @@ use Sabberworm\CSS\Value\ValueList;
 abstract class CSSBlockList extends CSSList
 {
     /**
-     * @param int<0, max> $lineNumber
-     */
-    public function __construct($lineNumber = 0)
-    {
-        parent::__construct($lineNumber);
-    }
-
-    /**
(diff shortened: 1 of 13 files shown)


JakeQZ (commented):
Only had a brief look at this.  Initializing properties to `null` is not necessary since that is the default, and by convention this is not normally explicitly done.

I found https://stackoverflow.com/questions/13114602/are-parent-constructors-called-if-a-child-class-does-not-define-a-constructor#13114603 which confirms constructors can be removed provided the base class defines one.

oliverklee:
Done and repushed.

JakeQZ (approved):
I think this works now.

IIRC, the PHP language issue with undefined constructors is when there is a base class with no constructor defined:
- the derived class cannot call `parent::__construct()`;
- if the base class is later modified to add a constructor to perform an important initialization task, it won't be implicitly called;
- to mitigate against this, the base class should have an empty constructor from the outset, so that the derived class can call it, should it later need to do something.
- (C++ does not have this problem because default constructors are always implicitly called.)

The issue doesn't apply here, since all the removed constructors are in derived classes for which there is a already a constructor in the base class.

PR #934, opened 2025-02-15
[TASK] Add some more unit tests for `AtRuleBlockList`

oliverklee:
Part of #757


JakeQZ (changes requested):
> This does not cover much code from `CSSList`, though.

If it covers some, I guess the `@covers` annotation is correct.

I found a typo.  Otherwise all good.

JakeQZ (approved):

PR #937, opened 2025-02-16
[CLEANUP] Use the accessor for the charset in `ParserState`

oliverklee:
Also make the getter `private` as it's not (intended to be) called from the outside.

As the whole `ParserState` class is `@internal`, this does does not require a deprecation process.

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -87,7 +87,7 @@ class ParserState
      *
      * @return string
      */
-    public function getCharset()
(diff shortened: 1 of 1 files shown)


JakeQZ (commented):
It isn't usual to use an accessor method for a class property within the class itself.  The accessor methods are typically provided for use by other classes (and may be marked as internal if only intended for intra-library use).  So this doesn't seem right to me.

oliverklee:
Okay, closing in favor of #941 then.

PR #940, opened 2025-02-16
[CLEANUP] Avoid Hungarian notation for `text`/`characters`

oliverklee:
Part of #756

### DIFF
--- a/src/Parser.php
+++ b/src/Parser.php
@@ -19,15 +19,15 @@ class Parser
     private $parserState;
 
     /**
-     * @param string $sText the complete CSS as text (i.e., usually the contents of a CSS file)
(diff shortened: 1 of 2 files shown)


JakeQZ (changes requested):
I think where it is `$aText` it is actually an array of possibly-multibyte characters (derived by splitting with an empty 'separator').  @sabberworm, would you be able to confirm this?

Much as I'd like a quaint reference to great literary works, I think this should be renamed to `$characters`, [which also has a similar quirky literary metaphor](https://en.wikipedia.org/wiki/Six_Characters_in_Search_of_an_Author).

JakeQZ (approved):
Looking at the `substr` method also confirms that the formerly-named `$aText` property is an array of (multibyte) characters.

PR #943, opened 2025-02-16
[TASK] Add some tests for `OutputFormatter` (part 1)

oliverklee:
Note: The new PHPStan warnings will go away once we add
native type declarations for `OutputFormatter`.

Part of #757

### DIFF
--- a/tests/Unit/OutputFormatterTest.php
+++ b/tests/Unit/OutputFormatterTest.php
@@ -229,7 +229,7 @@ final class OutputFormatterTest extends TestCase
         $values = [$value1, $value2];
         $separator = ', ';
 
-        $result = $this->subject->implode(', ', $values);
+        $result = $this->subject->implode($separator, $values);
 
         self::assertSame($value1 . $separator . $value2, $result);
     }
--- a/tests/Unit/OutputFormatterTest.php
+++ b/tests/Unit/OutputFormatterTest.php
@@ -159,7 +159,7 @@ final class OutputFormatterTest extends TestCase
     /**
      * @test
      */
-    public function spaceAfterListArgumentSeparatorForExistingSpaceAfterProvidedSeparatorReturnsThat(): void
+    public function spaceAfterListArgumentSeparatorReturnsSpaceSetForSpecificSeparator(): void
     {
         $separator = ',';
         $space = '        ';
--- a/tests/Unit/OutputFormatterTest.php
+++ b/tests/Unit/OutputFormatterTest.php
@@ -258,7 +258,7 @@ final class OutputFormatterTest extends TestCase
         $renderedRenderable1 = 'tea';
         $renderable1->method('render')->with($this->outputFormat)->willReturn($renderedRenderable1);
         $renderable2 = $this->createMock(Renderable::class);
-        $renderedRenderable2 = 'tea';
+        $renderedRenderable2 = 'coffee';
         $renderable2->method('render')->with($this->outputFormat)->willReturn($renderedRenderable2);
         $values = [$renderable1, $renderable2];
         $separator = ', ';
--- a/tests/Unit/OutputFormatterTest.php
+++ b/tests/Unit/OutputFormatterTest.php
@@ -1,0 +1,300 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit;
+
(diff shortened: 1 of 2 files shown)


tests/Unit/OutputFormatterTest.php
```diff
@@ -0,0 +1,300 @@
...
+     * @test
+     */
+    public function implodeWithMultipleStringValuesReturnsValuesSeparatedBySeparator(): void
+    {
+        $value1 = 'tea';
+        $value2 = 'coffee';
+        $values = [$value1, $value2];
+        $separator = ', ';
+
+        $result = $this->subject->implode(', ', $values);
```
JakeQZ, line 232:
```suggestion
        $result = $this->subject->implode($separator, $values);
```

tests/Unit/OutputFormatterTest.php
```diff
@@ -0,0 +1,300 @@
...
+    /**
+     * @test
+     */
+    public function implodeWithMultipleRenderablesReturnsRenderedRenderablesSeparatedBySeparator(): void
+    {
+        $renderable1 = $this->createMock(Renderable::class);
+        $renderedRenderable1 = 'tea';
+        $renderable1->method('render')->with($this->outputFormat)->willReturn($renderedRenderable1);
+        $renderable2 = $this->createMock(Renderable::class);
+        $renderedRenderable2 = 'tea';
```
JakeQZ, line 261: Can I have coffee please? ;)

tests/Unit/OutputFormatterTest.php
```diff
@@ -0,0 +1,300 @@
...
+    {
+        $renderable = $this->createMock(Renderable::class);
+        $renderedRenderable = 'tea';
+        $renderable->method('render')->with($this->outputFormat->nextLevel())->willReturn($renderedRenderable);
+        $values = [$renderable];
+
+        $result = $this->subject->implode(', ', $values, true);
+
+        self::assertSame($renderedRenderable, $result);
+    }
```
JakeQZ, line 299: I don't see the level increment being asserted here.  Shouldn't we be getting tea after some space?

JakeQZ (changes requested):
Excellent demonstration of how to use mock objects (which I have not yet tried).

Some test method names I found difficult to understand.  But then disocovered trying to come up with better alternatives was equally difficult.  I finally worked out some suggested renames.

A variable set up in one test should be used in that test.

The level increment test doesn't appear to be testing the level increment.

tests/Unit/OutputFormatterTest.php
```diff
@@ -0,0 +1,300 @@
...
+        $defaultSpace = "\t\t\t\t";
+        $this->outputFormat->setSpaceAfterListArgumentSeparator($defaultSpace);
+
+        self::assertSame($space, $this->subject->spaceAfterListArgumentSeparator($separator));
+    }
+
+    /**
+     * @test
+     */
+    public function spaceAfterListArgumentSeparatorForNoExistingSpaceAfterProvidedSeparatorReturnsDefaultSeparator(
```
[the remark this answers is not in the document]
JakeQZ: (Failure to submit review comments in GitHub results in them being lost.  There is no warning about unsubmitted data.  Had to redo this one.)

oliverklee:
> Some test method names I found difficult to understand. But then disocovered trying to come up with better alternatives was equally difficult. I finally worked out some suggested renames.

Thanks! (Yes, I had a hard time coming up with good names for these cases, too …)

tests/Unit/OutputFormatterTest.php
JakeQZ: I missed the change to the argument passed to `with()` compared with the previous test.  I think I see and understand what's going on now.  Reviewing this has been an educative process 👍

JakeQZ (approved):
This has been an educative PR, including perhaps also learning that we have different preferences for caffeinated hot drinks.

JakeQZ:
I also found this quite amusing:

have-a-coffee

JakeQZ:
> I prefer coffee

Seems I misunderstood since you preferred tea in the tests.

> So, what's your preference?

Coffee with/after breakfast and lunch, and after evening dinner.  Usually from a cafatiere.  When I smoked, I would have half a cup with each cigarette too.

:)

PR #944, opened 2025-02-16
[TASK] Add one more unit test for `Document`

oliverklee:
Part of #757


JakeQZ (approved):
I expect you are already planning to add tests for this method in the derived classes :)

PR #945, opened 2025-02-16
[TASK] Narrow the some type annotations in `OutputFormatter`

oliverklee:


src/OutputFormatter.php
```diff
@@ -136,6 +136,7 @@ public function safely($cCode)
     /**
      * Clone of the `implode` function, but calls `render` with the current output format instead of `__toString()`.
      *
+     * @param non-empty-string $sSeparator
```
JakeQZ, line 139: There might be use cases where this is an empty string, for example concatenating hex colour values `['r' => '00', 'g' => 'FF', 'b' => '00']`.

Given that the inbuilt `implode()` that this is based on accepts an empty string as the `$glue` parameter, I would prefer not to change this.  I note that the [PHP docs](https://www.php.net/manual/en/function.implode.php) have [renamed this parameter since PHP 8](https://web.archive.org/web/20201205012911/https://www.php.net/manual/en/function.implode.php) to `$separator`.  However, IMO that is a misnomer, implying, um, separation, which an empty string (the default) does not achieve.

oliverklee:
Good point! Changed and repushed.

JakeQZ (approved):

PR #950, opened 2025-02-18
[CLEANUP] Avoid Hungarian notation for `length`

oliverklee:
Part of #756


JakeQZ (approved):
This has reminded me to add #953.

PR #952, opened 2025-02-18
[TASK] Use more real-life data in a unit test

oliverklee:
This makes the test easier to read. Also, this prepares for when an `AtRuleBlockList` might not allow an empty type anymore.


JakeQZ (approved):
👍 to the [Foo Fighters](https://tidal.com/browse/track/3146740?u) :)

PR #954, opened 2025-02-19
[CLEANUP] Don't store array length as a property

### LINKED ISSUE #953: Remove `ParserState::$length`
This property merely represents `\count(ParserState::$characters)`.

Unlike C, PHP stores arrays internally with the length included, in such a way that obtaining it is atomic (O(1)).  We don't need to duplicate a feature already provided by the language.

JakeQZ:
Since PHP 5, if not earlier, the length of an array is internally stored by PHP and can be accessed in O(1) time.
Maintaining both the array and its length in class properties is error-prone (particularly regarding possible future code changes). When used in a loop, `\count($array)` is more repetitively expensive than `$arrayLength`,
but the latter can be set up as a local variable as and when needed.

Reference:
https://stackoverflow.com/questions/5835241/is-phps-count-function-o1-or-on-for-arrays

This change also ensures that the characters are always a (dazzling) array.

Resolves #953.

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -30,7 +30,7 @@ class ParserState
     /**
      * @var array<int, string>
      */
-    private $characters;
+    private $characters = [];
 
     /**
      * @var int
@@ -43,11 +43,6 @@ class ParserState
      * @var string
      */
     private $charset;
-
-    /**
-     * @var int
-     */
-    private $length;
 
     /**
      * @var int
@@ -75,7 +70,6 @@ class ParserState
     {
         $this->charset = $charset;
         $this->characters = $this->strsplit($this->text);
-        $this->length = \count($this->characters);
     }
 
     /**
@@ -225,7 +219,7 @@ class ParserState
                 try {
                     $oComment = $this->consumeComment();
                 } catch (UnexpectedEOFException $e) {
-                    $this->currentPosition = $this->length;
+                    $this->currentPosition = \count($this->characters);
                     return $comments;
                 }
             } else {
@@ -257,7 +251,7 @@ class ParserState
     public function peek($length = 1, $offset = 0): string
     {
         $offset += $this->currentPosition;
-        if ($offset >= $this->length) {
+        if ($offset >= \count($this->characters)) {
             return '';
         }
         return $this->substr($offset, $length);
@@ -286,7 +280,7 @@ class ParserState
             $this->currentPosition += $this->strlen($mValue);
             return $mValue;
         } else {
-            if ($this->currentPosition + $mValue > $this->length) {
+            if ($this->currentPosition + $mValue > \count($this->characters)) {
                 throw new UnexpectedEOFException((string) $mValue, $this->peek(5), 'count', $this->lineNumber);
             }
             $result = $this->substr($this->currentPosition, $mValue);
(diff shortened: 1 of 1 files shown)


src/Parsing/ParserState.php
```diff
@@ -69,15 +64,21 @@ public function __construct($text, Settings $parserSettings, $lineNumber = 1)
...
     {
+        // The charset property must be set for `strsplit()`, but may need to be reverted if it fails.
+        $oldCharset = $this->charset;
         $this->charset = $charset;
-        $this->characters = $this->strsplit($this->text);
-        if (\is_array($this->characters)) {
-            $this->length = \count($this->characters);
+
+        $newCharacters = $this->strsplit($this->text);
+        if (!\is_array($newCharacters)) {
```
oliverklee, line 76: I don't think we need to catch this case. `$this->strsplit` is expected to return `array`s only. (And if the `preg_split` in that function fails, we should handle that there, not here.)

src/Parsing/ParserState.php
```diff
@@ -69,15 +64,21 @@ public function __construct($text, Settings $parserSettings, $lineNumber = 1)
     /**
      * Sets the charset to be used if the CSS does not contain an `@charset` declaration.
      *
-     * @param string $charset
+     * @throws SourceException
      */
-    public function setCharset($charset): void
+    public function setCharset(string $charset): void
```
oliverklee, line 69: I'd prefer to add the native type declaration in a separate PR (so this PR here is more focused).

oliverklee (changes requested):

src/Parsing/ParserState.php
JakeQZ: #958

oliverklee:
#958 is merged now.

src/Parsing/ParserState.php
JakeQZ: #962 to ensure `strsplit()` will only ever return arrays.

JakeQZ:
> This PR now is in need of rebasing.

Done.  Was tricky, and at times I wondered if restarting from scratch would be easier.

oliverklee (approved):

src/Parsing/ParserState.php
```diff
@@ -30,7 +30,7 @@ class ParserState
     /**
      * @var array<int, string>
      */
-    private $characters;
+    private $characters = [];
```
oliverklee, line 33: This change is not needed (as we set it in `setCharset()`, which is called from the constructor. Also, it's misleading as this property does not have a default value after construction.
```suggestion
    private $characters;
```

oliverklee (changes requested):

src/Parsing/ParserState.php
JakeQZ: This change gives it a default value, ensuring it is always an array, in case the construction chain somehow fails.

oliverklee: AFAIK, if construction fails (e.g., if an exception is thrown during construction), there will be no constructed object: https://onlinephp.io/c/7a627

(Also, I think if construction fails, we shouldn't try return a half-broken object, but have properly not-yet-initialized properties if they have not been initialized with meaningful data.)

JakeQZ: Would PHPStan be happy that it's always an array, even if only initialized to one by a method called by the constructor?

oliverklee: Yes, AFAIK.

oliverklee (approved):

PR #956, opened 2025-02-19
[TASK] Add Tests for `Document::getAllDeclarationBlocks()`

oliverklee:
Also remove a misleading and incorrect comment.

Part of #757

### DIFF
--- a/tests/Unit/CSSList/DocumentTest.php
+++ b/tests/Unit/CSSList/DocumentTest.php
@@ -215,7 +215,7 @@ final class DocumentTest extends TestCase
     {
         $subject = new Document();
 
-        $import = new Import(new URL(new CSSString('')), '');
+        $import = new Import(new URL(new CSSString('https://www.example.com/')), '');
         $subject->setContents([$import]);
 
         $result = $subject->getAllDeclarationBlocks();
@@ -230,7 +230,7 @@ final class DocumentTest extends TestCase
     {
         $subject = new Document();
 
-        $charset = new Charset(new CSSString(''));
+        $charset = new Charset(new CSSString('UTF-8'));
         $subject->setContents([$charset]);
 
         $result = $subject->getAllDeclarationBlocks();
--- a/tests/Unit/CSSList/DocumentTest.php
+++ b/tests/Unit/CSSList/DocumentTest.php
@@ -194,7 +194,7 @@ final class DocumentTest extends TestCase
     /**
      * @test
      */
-    public function getAllDeclarationBlocksAlsoReturnsDeclarationBlocksWithinAtRuleBlockList(): void
+    public function getAllDeclarationBlocksReturnsDeclarationBlocksWithinAtRuleBlockList(): void
     {
         $subject = new Document();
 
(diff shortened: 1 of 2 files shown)


JakeQZ (commented):
The tests look fine.  But I find them difficult to follow because I can't immediately see how `$this->subject` has been `setUp()`.  Would it perhaps be clearer for the subject to be set up on a per-test basis?  And maybe a separate pre-PR to remove the `setUp()` method, which I don't think is warranted.

oliverklee:
Updated an repushed.

tests/Unit/CSSList/DocumentTest.php
```diff
@@ -142,4 +149,92 @@ public function insertContentBeforeAppendsIfSiblingNotFound(): void
...
+    }
+
+    /**
+     * @test
+     */
+    public function getAllDeclarationBlocksIgnoresImport(): void
+    {
+        $subject = new Document();
+
+        $import = new Import(new URL(new CSSString('')), '');
```
JakeQZ, line 218: A nonempty and valid URL like `https://example.org/` would be better here.

tests/Unit/CSSList/DocumentTest.php
```diff
@@ -142,4 +149,92 @@ public function insertContentBeforeAppendsIfSiblingNotFound(): void
...
+
+        $result = $subject->getAllDeclarationBlocks();
+
+        self::assertSame([$declarationBlock1, $declarationBlock2], $result);
+    }
+
+    /**
+     * @test
+     */
+    public function getAllDeclarationBlocksAlsoReturnsDeclarationBlocksWithinAtRuleBlockList(): void
```
JakeQZ, line 197: Niggly, but we're not *actually* writing an essay :)

The test method name should be able to stand alone.

```suggestion
    public function getAllDeclarationBlocksReturnsDeclarationBlocksWithinAtRuleBlockList(): void
```

tests/Unit/CSSList/DocumentTest.php
```diff
@@ -142,4 +149,92 @@ public function insertContentBeforeAppendsIfSiblingNotFound(): void
...
+    }
+
+    /**
+     * @test
+     */
+    public function getAllDeclarationBlocksIgnoresCharset(): void
+    {
+        $subject = new Document();
+
+        $charset = new Charset(new CSSString(''));
```
JakeQZ, line 233: A nonempty valid charset like `UTF-8` would be better here.

JakeQZ (changes requested):
I think a test method name should not include 'also', and that it would be better to provide non-empty strings which are ideally real-world examples for URL and charset.  Otherwise fine.

tests/Unit/CSSList/DocumentTest.php
oliverklee: Done.

tests/Unit/CSSList/DocumentTest.php
oliverklee: Done.

tests/Unit/CSSList/DocumentTest.php
JakeQZ: Unless I've missed something, this is not yet resolved.

JakeQZ (changes requested):
The method name issue does not seem to be resolved.

oliverklee:
> The method name issue does not seem to be resolved.

Indeed. Done.

JakeQZ (approved):

PR #963, opened 2025-02-20
[CLEANUP] Use a per-test subject for `DocumentTest`

oliverklee:


JakeQZ (approved):
IMO it does indeed make it much clearer to see what's going on in each test :)

And there's barely any more code involved by replacing `$this->subject` with `new Document()`.

PR #966, opened 2025-02-20
[CLEANUP] Use accessors in `OutputFormatter`

oliverklee:
Avoid direct access to `OutputFormat` properties as well as variable access and magic methods.

Also drop some dead code.

We might possibly want to drop the `space` method altogether, but that's out of scope for this change.

### DIFF
--- a/src/OutputFormatter.php
+++ b/src/OutputFormatter.php
@@ -19,18 +19,54 @@ class OutputFormatter
         $this->outputFormat = $outputFormat;
     }
 
-    public function space(string $sName, ?string $sType = null): string
+    /**
+     * @param non-empty-string $sName
+     *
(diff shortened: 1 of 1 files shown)


JakeQZ (changes requested):
It does indeed seem that the `$sType` parameter is now unused and can be removed.

Whilst making this change, could we not also remove the Hungarian notation from `$sSpaceString`?

oliverklee:
> Whilst making this change, could we not also remove the Hungarian notation from `$sSpaceString`?

I'd prefer to do that in a separate change.

JakeQZ (approved):
> > Whilst making this change, could we not also remove the Hungarian notation from `$sSpaceString`?
> 
> I'd prefer to do that in a separate change.

PR #969, opened 2025-02-20
[CLEANUP] Avoid Hungarian notation in `OutputFormatter`

oliverklee:
Part of #756

### DIFF
--- a/src/OutputFormatter.php
+++ b/src/OutputFormatter.php
@@ -20,54 +20,54 @@ class OutputFormatter
     }
 
     /**
-     * @param non-empty-string $sName
+     * @param non-empty-string $name
      *
      * @throws \InvalidArgumentException
      */
-    public function space(string $sName): string
+    public function space(string $name): string
     {
-        switch ($sName) {
+        switch ($name) {
             case 'AfterRuleName':
-                $sSpaceString = $this->outputFormat->getSpaceAfterRuleName();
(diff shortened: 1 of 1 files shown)


src/OutputFormatter.php
```diff
@@ -143,62 +143,62 @@ public function spaceBeforeOpeningBrace(): string
...
+    public function removeLastSemicolon(string $string): string
     {
         if ($this->outputFormat->getSemicolonAfterLastRule()) {
-            return $sString;
+            return $string;
         }
-        $sString = \explode(';', $sString);
-        if (\count($sString) < 2) {
-            return $sString[0];
+        $parts = \explode(';', $string);
```
JakeQZ, line 194: I'd also noted the rewriting of the parameter and thought a fresh local variable would be better 👍

src/OutputFormatter.php
```diff
@@ -20,54 +20,54 @@ public function __construct(OutputFormat $outputFormat)
     }
 
     /**
-     * @param non-empty-string $sName
+     * @param non-empty-string $name
      *
      * @throws \InvalidArgumentException
      */
-    public function space(string $sName): string
+    public function space(string $name): string
```
JakeQZ, line 27: `$name` is not very descriptive, but that is beyond the scope of this PR.

JakeQZ (approved):

PR #972, opened 2025-02-23
[CLEANUP] Avoid Hungarian notation for `string*`

oliverklee:
Part of #756

### DIFF
--- a/src/RuleSet/DeclarationBlock.php
+++ b/src/RuleSet/DeclarationBlock.php
@@ -45,18 +45,18 @@ class DeclarationBlock extends RuleSet
         $result = new DeclarationBlock($parserState->currentLine());
         try {
             $aSelectorParts = [];
-            $sStringWrapperChar = false;
+            $stringWrapperChar = false;
             do {
                 $aSelectorParts[] = $parserState->consume(1)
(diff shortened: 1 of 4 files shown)


src/RuleSet/DeclarationBlock.php
```diff
@@ -45,18 +45,18 @@ public static function parse(ParserState $parserState, $list = null)
         $result = new DeclarationBlock($parserState->currentLine());
         try {
             $aSelectorParts = [];
-            $sStringWrapperChar = false;
+            $stringWrapperChar = false;
```
JakeQZ, line 48: As we're renaming this variable anyway, could we not change it to `$stringWrapperCharacter` in this PR?

I note a couple of other very minor issues with this section of code that are beyond the scope of this PR, but should be tidied up (and can't be until this PR is merged):
1. The non-set value for the above variable should be `null` rather then `false`;
2. The comparison in the `elseif` should be strict (using `===`).

JakeQZ (changes requested):
I think a better rename is possible for `$sStringWrapperChar`.

src/RuleSet/DeclarationBlock.php
oliverklee: I have create #975 and #976 for those two issues.

JakeQZ (approved):

PR #973, opened 2025-02-23
[CLEANUP] Drop some comments from testcases

oliverklee:
These comments are no longer needed now that the abstract base classes have their own testcases.

### DIFF
--- a/tests/Unit/CSSList/AtRuleBlockListTest.php
+++ b/tests/Unit/CSSList/AtRuleBlockListTest.php
@@ -18,10 +18,6 @@ use Sabberworm\CSS\Renderable;
  */
 final class AtRuleBlockListTest extends TestCase
 {
-    /*
-     * Tests for the implemented interfaces and superclasses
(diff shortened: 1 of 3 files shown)


JakeQZ (commented):
Shouldn't "Tests for functionality implemented in `CSSBlockList`" belong in a `CSSBlockListTest` class, etc.?

Would a better approach be to move the existing tests into more relevant test classes (creating them as necessary) which correspond to the class actually being tested?

JakeQZ (changes requested):
> This PR here is only the pre-patch for this (as explained in the commit message):

OK.  But I'm really struggling with the 'diffs' in this.  Could this PR be broken down into more manageable chunks?

oliverklee:
> Could this PR be broken down into more manageable chunks?

Will do.

#978 is the first part.

JakeQZ (approved):
They were temporarily in place during the refactoring.

PR #977, opened 2025-02-24
[CLEANUP] Avoid Hungarian notation for `start(Position)`

oliverklee:
Part of #756

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -424,21 +424,21 @@ class ParserState
     }
 
     /**
-     * @param int $start
+     * @param int $offset
      * @param int $length
      */
-    private function substr($start, $length): string
+    private function substr($offset, $length): string
     {
         if ($length < 0) {
-            $length = \count($this->characters) - $start + $length;
+            $length = \count($this->characters) - $offset + $length;
         }
-        if ($start + $length > \count($this->characters)) {
-            $length = \count($this->characters) - $start;
+        if ($offset + $length > \count($this->characters)) {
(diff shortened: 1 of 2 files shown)


src/Parsing/ParserState.php
```diff
@@ -424,21 +424,21 @@ public function strlen($sString): int
     }
 
     /**
-     * @param int $iStart
+     * @param int $start
      * @param int $length
      */
-    private function substr($iStart, $length): string
+    private function substr($start, $length): string
```
JakeQZ, line 430: The [equivalant PHP function](https://www.php.net/manual/en/function.substr.php) has this parameter as `$offset`, which I think is better.

src/Value/Value.php
```diff
@@ -78,23 +78,23 @@ public static function parseValue(ParserState $parserState, array $aListDelimite
                 return $aStack[0];
             }
             $aNewStack = [];
-            for ($iStartPosition = 0; $iStartPosition < $iStackLength; ++$iStartPosition) {
-                if ($iStartPosition === ($iStackLength - 1) || $sDelimiter !== $aStack[$iStartPosition + 1]) {
-                    $aNewStack[] = $aStack[$iStartPosition];
+            for ($startPosition = 0; $startPosition < $iStackLength; ++$startPosition) {
```
JakeQZ, line 81: Perhaps `$offset` would be a better name here too.

JakeQZ (changes requested):
I think `$offset` would be better throughout as a variable name.

JakeQZ (approved):

PR #980, opened 2025-02-24
[CLEANUP] Avoid Hungarian notation in `ParserState`

oliverklee:
Part of #756

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -404,11 +404,11 @@ class ParserState
     }
 
     /**
-     * @param int $iAmount
+     * @param int $amount
      */
-    public function backtrack($iAmount): void
+    public function backtrack($amount): void
     {
-        $this->currentPosition -= $iAmount;
+        $this->currentPosition -= $amount;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


src/Parsing/ParserState.php
```diff
@@ -404,11 +404,11 @@ public function streql($string1, $string2, $caseInsensitive = true): bool
     }
 
     /**
-     * @param int $iAmount
+     * @param int $amount
```
JakeQZ, line 407: We could possibly improve this parameter name while renaming it.  Perhaps `$characterCount` or `$numberOfCharacters`.

src/Parsing/ParserState.php
```diff
@@ -490,18 +490,18 @@ private function strsplit($string)
     }
 
     /**
-     * @param string $string
-     * @param string $sNeedle
+     * @param string $haystack
```
JakeQZ, line 493: Good rename to match the PHP docs.

JakeQZ (changes requested):
Could we improve on the naming of the `$amount` parameter, as we are renaming it?

JakeQZ (approved):

PR #983, opened 2025-02-24
[TASK] Drop unnecessary properties from `UnexpectedTokenException`

oliverklee:


JakeQZ (approved):
Good catch.  These were private and only ever used by the constructor (and in some cases not even there).

PR #984, opened 2025-02-24
[TASK] Add a dedicated testcase for the abstract `CSSBlockList`

oliverklee:


tests/Unit/CSSList/CSSBlockListTest.php
```diff
@@ -0,0 +1,47 @@
...
+    }
+
+    /**
+     * @test
+     */
+    public function isCSSList(): void
+    {
+        $subject = new ConcreteCSSBlockList();
+
+        self::assertInstanceOf(ConcreteCSSBlockList::class, $subject);
```
JakeQZ, line 45: Shouldn't this be asserting it's an instance of `CSSList`?

JakeQZ (changes requested):
Looks good, but I think there's a mistake in one of the tests.

tests/Unit/CSSList/CSSBlockListTest.php
oliverklee: Indeed, good catch! Fixed.

JakeQZ: I thought this would also need a `use`, so did not use the 'suggestion' feature.

JakeQZ (approved):

PR #986, opened 2025-02-24
[CLEANUP] Avoid Hungarian notation for `selector`

oliverklee:
Part of #756


JakeQZ (approved):
As per other PR, this looks fine, so if it was intended for review, it's approved for merging now.

Issue #976, opened 2025-02-24
Use strict comparisons


oliverklee:
We should make sure first that the changed code is covered with tests.

JakeQZ:
For reference, one instance is in `DeclarationBlock::parse()`, and was spotted during review of #972.

PR #988, opened 2025-02-25
[CLEANUP] Use `null` for unset value in `DeclarationBlock::parse()`

### LINKED ISSUE #975: Use a  `null` default value in `DeclarationBlock::parse()` instead of `false`
https://github.com/MyIntervals/PHP-CSS-Parser/pull/972#pullrequestreview-2635819716

JakeQZ:
The variable is either a string or it isn't.
The best practice for when it isn't is for it to be `null` or `unset()`.

Using `false` for when it isn't can lead to type-safety issues. Reference:
https://vulke.medium.com/when-should-variables-be-null-false-undefined-or-an-empty-string-in-php-4ebd73c7a954

Also use `===` in a comparison in the affected code.

Resolves #975.
Part of #976.

### DIFF
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -331,12 +331,6 @@ parameters:
 			path: ../src/RuleSet/DeclarationBlock.php
 
 		-
-			message: '#^Loose comparison via "\=\=" is not allowed\.$#'
-			identifier: equal.notAllowed
-			count: 1
-			path: ../src/RuleSet/DeclarationBlock.php
-
-		-
 			message: '#^Parameters should have "string" types as the only types passed to this method$#'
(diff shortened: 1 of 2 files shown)


JakeQZ:
It's possible to use `=== null` and `= null` instead of `isset()` and `unset()`, if that's preferred.  But given PHP offers these functions, why not use them?

oliverklee (approved):

oliverklee:
> It's possible to use `=== null` and `= null` instead of `isset()` and `unset()`, if that's preferred. But given PHP offers these functions, why not use them?

I slightly prefer setting them to `null`.

But what I find most important is that we keep things consistent, i.e., use `isset` and `unset` if it's about set/not set variables, and use `null` checks for cases where a variable is set, but might be null. (This is what we're doing here.)

PR #990, opened 2025-02-25
[CLEANUP] Refactor `::getAllDeclarationBlocks()`

oliverklee:
Part of #994.

### DIFF
--- a/src/CSSList/CSSBlockList.php
+++ b/src/CSSList/CSSBlockList.php
@@ -23,13 +23,23 @@ abstract class CSSBlockList extends CSSList
     /**
      * Gets all `DeclarationBlock` objects recursively, no matter how deeply nested the selectors are.
      *
-     * @return array<int, DeclarationBlock>
+     * @return list<DeclarationBlock>
      */
     public function getAllDeclarationBlocks(): array
     {
-        /** @var array<int, DeclarationBlock> $result */
+        /** @var list<DeclarationBlock> $result */
         $result = [];
-        $this->allDeclarationBlocks($result);
+
+        foreach ($this->contents as $directSibling) {
+            if ($directSibling instanceof DeclarationBlock) {
+                $result[] = $directSibling;
+            } elseif ($directSibling instanceof CSSBlockList) {
+                foreach ($directSibling->getAllDeclarationBlocks() as $grandchild) {
+                    $result[] = $grandchild;
+                }
+            }
+        }
+
         return $result;
     }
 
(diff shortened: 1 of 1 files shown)


oliverklee:
> Can't the method `getAllDeclarationBlocks` simply be moved from `Document` to `CSSBlockList` as is (continuing to use the internal `allDeclarationBlocks()`?

I'm planning on removing `allDeclarationBlocks` as it's redundant with `getAllDeclarationBlocks` and also is hard to read and PHPStan-analyze due to the return parameter.

I've created #994 to list the steps of this plan.

oliverklee:
Okay, I'll mark this as draft for now and move the method up first.

src/CSSList/CSSBlockList.php
```diff
@@ -23,13 +23,23 @@ abstract class CSSBlockList extends CSSList
     /**
      * Gets all `DeclarationBlock` objects recursively, no matter how deeply nested the selectors are.
      *
-     * @return array<int, DeclarationBlock>
+     * @return list<DeclarationBlock>
      */
     public function getAllDeclarationBlocks(): array
     {
-        /** @var array<int, DeclarationBlock> $result */
+        /** @var list<DeclarationBlock> $result */
```
JakeQZ, line 30: Is this still needed?

src/CSSList/CSSBlockList.php
```diff
@@ -23,13 +23,23 @@ abstract class CSSBlockList extends CSSList
...
         $result = [];
-        $this->allDeclarationBlocks($result);
+
+        foreach ($this->contents as $directSibling) {
```
JakeQZ, line 33: Sibling of whom?  Direct (usu. adjacent) sibling (e.g. CSS `+`) normally refers to two of them.  I think the original `$item` was fine.

src/CSSList/CSSBlockList.php
```diff
@@ -23,13 +23,23 @@ abstract class CSSBlockList extends CSSList
...
+            if ($directSibling instanceof DeclarationBlock) {
+                $result[] = $directSibling;
+            } elseif ($directSibling instanceof CSSBlockList) {
+                foreach ($directSibling->getAllDeclarationBlocks() as $grandchild) {
+                    $result[] = $grandchild;
+                }
```
JakeQZ, line 39: `array_merge()` will do the job here.

JakeQZ (changes requested):
Functionally fine, but have made some suggestions for improvement.

src/CSSList/CSSBlockList.php
oliverklee: I've removed it for now. If PHPStan complains (on a higher check level), we can re-add it.

JakeQZ: It may have been needed when the array was passed by reference as an argument to `allDeclarationBlocks()`.

JakeQZ (approved):

PR #991, opened 2025-02-25
[TASK] Add tests for `Document::getAllRuleSets()`

oliverklee:
Part of #757

### DIFF
--- a/tests/Unit/CSSList/DocumentTest.php
+++ b/tests/Unit/CSSList/DocumentTest.php
@@ -208,7 +208,7 @@ final class DocumentTest extends TestCase
     /**
      * @test
      */
-    public function getAllRuleSetsCanReturnMultipleDirectAtRuleSetContents(): void
+    public function getAllRuleSetsReturnsMultipleAtRuleSetsDirectlySetAsContents(): void
     {
         $subject = new Document();
 
--- a/tests/Unit/CSSList/DocumentTest.php
+++ b/tests/Unit/CSSList/DocumentTest.php
@@ -162,7 +162,7 @@ final class DocumentTest extends TestCase
     /**
      * @test
      */
-    public function getAllRuleSetsCanReturnOneDirectDeclarationBlockContent(): void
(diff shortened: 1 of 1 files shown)


tests/Unit/CSSList/DocumentTest.php
```diff
@@ -156,6 +157,142 @@ public function getAllDeclarationBlocksIgnoresCharset(): void
...
+
+        $result = $subject->getAllRuleSets();
+
+        self::assertSame([$atRuleSet], $result);
+    }
+
+    /**
+     * @test
+     */
+    public function getAllRuleSetsCanReturnMultipleDirectDeclarationBlockContents(): void
```
JakeQZ, line 195:
```suggestion
    public function getAllRuleSetsReturnsMultipleDeclarationBlocksDirectlySetAsContents(): void
```

tests/Unit/CSSList/DocumentTest.php
```diff
@@ -156,6 +157,142 @@ public function getAllDeclarationBlocksIgnoresCharset(): void
...
+
+        $result = $subject->getAllRuleSets();
+
+        self::assertSame([$declarationBlock1, $declarationBlock2], $result);
+    }
+
+    /**
+     * @test
+     */
+    public function getAllRuleSetsCanReturnMultipleDirectAtRuleSetContents(): void
```
JakeQZ, line 211:
```suggestion
    public function getAllRuleSetsReturnsMultipleAtRuleSetsDirectlySetAsContents(): void
```

JakeQZ (commented):
The tests look fine.  I've suggested improvements for some of the test method names, where I found them difficult to understand (and, equally, difficult to rewrite).

JakeQZ (approved):
Trying to test that `RuleSet`s are being set does create some awkwardness re test method names.  It would much easier if they were called `RuleCollection`s.  Or we used German - the two meanings of the English word 'set' don't seem to have any correlation.

I think we've done the best we reasonably can :)

PR #993, opened 2025-02-25
[TASK] Deprecate the IE hack in `Rule`

oliverklee:
With modern browsers, we don't need IE hacks anymore.

(Also, this code is not used.)


JakeQZ (approved):
IIRC, these were only needed for versions of IE before IE8, which was released 15 years ago.  Nowadays, many websites (e.g. YouTube) won't even run on IE8.

PR #995, opened 2025-02-25
[TASK] Remove the IE hack in `Rule`

oliverklee:


src/Rule/Rule.php
```diff
@@ -94,7 +89,7 @@ public static function parse(ParserState $parserState): Rule
         if ($parserState->getSettings()->bLenientParsing) {
             while ($parserState->comes('\\')) {
                 $parserState->consume('\\');
-                $rule->addIeHack($parserState->consume());
+                $parserState->consume();
                 $parserState->consumeWhiteSpace();
             }
         }
```
JakeQZ, line 95: I think this whole code block is parsing the IE hack(s).  And the idea is that IE will treat them as valid rules whilst other browsers will discard them.  There's a [mention in the readme that IE hacks are only supported in lenient mode](https://github.com/MyIntervals/PHP-CSS-Parser#contributorsthanks-to).  So, on that basis, I think the entire `if` statement should be removed, so that invalid declarations with IE hacks are discarded rather than kept.

src/Rule/Rule.php
```diff
@@ -220,30 +215,6 @@ public function addValue($value, $type = ' '): void
...
     }
 
-    /**
-     * @param int $modifier
-     */
-    public function addIeHack($modifier): void
-    {
-        $this->ieHack[] = $modifier;
-    }
```
JakeQZ, line 229: In #993, this wasn't deprecated.  This should be done first.

JakeQZ (changes requested):
#993 deprecated this functionality in the 8,x branch, so we can remove it for 9.0 (provided we do actually release 8.8 first).  However, one deprecation was missing from that change.  (And a `private` property was unnecessarily deprecated.)

Also, if we remove support for IE hacks, we should not then allow CSS rules containing them to be parsed as valid.

src/Rule/Rule.php
oliverklee: Done: #1003

JakeQZ (changes requested):
This still isn't quite right, but I'm not sure why.  The rules with IE hacks should be dropped entirely (in lenient mode when an exception is not thrown).

If only the IE hack part of the value is dropped, the result is a property declaration that all browsers would apply.  Whereas with the IE hack, it is only applied in IE.

oliverklee:
> The rules with IE hacks should be dropped entirely (in lenient mode when an exception is not thrown).

The removed code was the only IE-hack-specific code I could find. My best guess ist that the rest is from "the way lenient parsing currently works".

I'd like to keep this PR focused on removing these specific pieces of code, and keep "cleaning up lenient parsing" out of it.

oliverklee:
I have changed the name of the test from `ieHacksAreRemovedInLenientMode` to `ieHacksArePartiallyRemovedInLenientMode` so that it's not misleading anymore.

JakeQZ:
> My best guess ist that the rest is from "the way lenient parsing currently works".

I think you're right.  Rule::parse() will return a valid `Rule` whatever junk comes after it (IE hacks or anything else).  It does not check that there is a semicolon or closing brace following (and will actually, it seems, parse multiple rules without semicolons between them).  These I think are bugs.  The exception in strict mode occurs on attempting to parse the 'next' rule, which is actually the junk or IE hack at the end of the previous.  So I think this is fine.

> I have changed the name of the test from `ieHacksAreRemovedInLenientMode` to `ieHacksArePartiallyRemovedInLenientMode` so that it's not misleading anymore.

I'm not sure about this rename - the IE hacks seem to be all removed - things like `content: "red \\0"` are actual strings, not IE hacks.

JakeQZ (changes requested):
> > The rules with IE hacks should be dropped entirely (in lenient mode when an exception is not thrown).
> 
> The removed code was the only IE-hack-specific code I could find. My best guess ist that the rest is from "the way lenient parsing currently works".

I managed to unearth the change that introduced the IE hacks: https://github.com/MyIntervals/PHP-CSS-Parser/commit/e1ecb78edee861dcfcf7d97dca090af22fae8b87

There are two additional changes I can see that also need to be undone.

1. `Value::parseValue()` has the responsibility of continually parsing until it finds something that terminates the value, like `;`, `}` or `!` (for `!important`).  There it is also testing for `\` for the IE hacks.  So that conditional should be removed.  Then we might find the rules with IE hacks are dropped as they should be (and would be by non-IE browsers).
    - If the above so far is good, the test method in lenient mode can be renamed to indicate that rules with IE hacks will be discarded.
2. `ParserState::parseCharacter` has special-case handling of `\0` and `\9` which should also be removed. 

I am not sure why `Value::parseValue()` is also stopping at `)`.  I have a vague recollection of that also being used in an IE hack.  However, the `Rule` class doesn't appear to support that as an IE hack (if it ever did), and that was already in place before the IE hack support was added.  That's beyond the scope of this PR.

oliverklee:
I have removed the other IE-hacks-related code parts now. I've also removed the tests as we now don't have a special case anymore (and I failed to create a usable expected string to compare with).

JakeQZ:
> I have removed the other IE-hacks-related code parts now. I've also removed the tests as we now don't have a special case anymore (and I failed to create a usable expected string to compare with).

Could you not just run the tests and capture the (unexpected) output, then paste that in as the expected output?

Having the tests in place was a massive help in working out what wasn't quite working, since you had changed the expected result to match the result after the code change.  Having that there saved any actual debugging, as it made it possible to work out what was going on by merely inspecting the code and looking at the output (in the tests).  So, for this PR, I would not like the tests removed, because I would like to see the result of the changes in the tests to confirm it is A-OK.

The tests can be dropped in a subsequent PR.

Does that work for you?

oliverklee:
> Could you not just run the tests and capture the (unexpected) output, then paste that in as the expected output?

I tried … and failed to copy'n'paste the actual output as it contained at least a tab and a non-displayable character. Would you be willing to give it a go (after I've rebase this PR and resolved the merge conflicts)?

JakeQZ:
> > Could you not just run the tests and capture the (unexpected) output, then paste that in as the expected output?
> 
> I tried … and failed to copy'n'paste the actual output as it contained at least a tab and a non-displayable character. Would you be willing to give it a go (after I've rebase this PR and resolved the merge conflicts)?

I'll give it a try later.  But I am using Windows :/

oliverklee:
> I'll give it a try later. But I am using Windows :/

If you get stuck, please let me know. Then I'll try again.

JakeQZ:
I've got a bit further with this.  With IE hacks support removed, `\9` and `\0` are now parsed as escaped characters.  So they become `TAB` and `NUL`, respectively.  Though there seem to be some stray backslashes in the rendered output.  I'll come back to this and see if I can work out why that is happening; probably just put the tests back in with the expected output as it will become (so we can still see what's happening), and create a separate issue for the handling of escaped characters outside of strings (although IIRC, strings in CSS don't need quotes), so we can get this PR completed...

JakeQZ (approved):
I've added #1053 to capture the issues found.

We can now remove those tests.

JakeQZ:
> (although IIRC, strings in CSS don't need quotes)

Actually they do.

Issue #994, opened 2025-02-25
Refactor code in `CSSList`


oliverklee:
The goal is to get rid of redundant code and to reduce functions that have return parameters (which are both hard to understand as well as hard analyze with PHPStan).

- [x] move `getAllDeclarationBlocks` from `Document` to the parent class `CSSBlockList`
- [x] refactor `getAllDeclarationBlocks`
- [x] replace usages of `CSSBlockList::allDeclarationBlocks` with calls to `::getAllDeclarationBlocks` and drop `allDeclarationBlocks`
- [x] move `getAllRuleSets` from `Document` to the parent class `CSSBlockList`
- [x] refactor `getAllRuleSets`
- [x] replace usages of `CSSBlockList::allRuleSets` with calls to `::getAllRuleSets` and drop `allRuleSets`
- [x] move `getAllValues` from `Document` to the parent class `CSSBlockList`
- [x] refactor `getAllValues`
- [x] replace usages of `CSSBlockList::allValues` with calls to `::getAllValues` and drop `allValues`
- [ ] rename `CSSBlockList::allSelectors` to `::getAllSelectors` and change its interface to return the result

PR #996, opened 2025-02-25
[TASK] Move up `getAllDeclarationBlocks` to `CSSBlockList`

oliverklee:
Move this method up without any changes to the method or its callers.

We'll clean this up in later changes.

Part of #994.


oliverklee:
Waiting for #991 to be merged first as there will probably be merge conflicts.

JakeQZ (changes requested):
The names of the moved test methods don't match those finally agreed upon in #991.

JakeQZ:
> I propose we align the names of the tests for `getAllDeclarationBlocks` with those for `getAllRuleSets` in a separate PR.

OK.  But the `getAllRuleSets` tests seem to be being removed in this PR.

oliverklee:
>  But the `getAllRuleSets` tests seem to be being removed in this PR.

Oh, that shouldn't have happened. Hold on …

oliverklee:
> Oh, that shouldn't have happened. Hold on …

Fixed now. Good catch!

JakeQZ (approved):
> This PR here is about `getAllDeclarationBlocks`. #991 is about `getAllRuleSets`.

I originally thought that the tests had been renamed from #991, since I was looking at the removed `getAllRuleSets` tests and failed to spot that the moved tests were those for `getAllDeclarationBlocks`.

I've checked more thoroughly now and confirm that the only difference in the moved tests is the subject now being a `ConcreteCSSBlockList` rather than a `Document`.

PR #999, opened 2025-02-25
[TASK] Add some basic tests for `CSSString`

oliverklee:
Part of #757

### DIFF
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -415,6 +415,12 @@ parameters:
 			path: ../src/RuleSet/RuleSet.php
 
 		-
+			message: '#^Parameters should have "string" types as the only types passed to this method$#'
+			identifier: typePerfect.narrowPublicClassMethodParamType
+			count: 1
+			path: ../src/Value/CSSString.php
+
+		-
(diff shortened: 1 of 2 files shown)


tests/Unit/Value/CSSStringTest.php
```diff
@@ -0,0 +1,84 @@
...
+
+        self::assertInstanceOf(Value::class, $subject);
+    }
+
+    /**
+     * @test
+     */
+    public function getStringReturnsStringProvidedToConstructor(): void
+    {
+        $string = 'coffee';
```
JakeQZ, line 44: :)

config/phpstan-baseline.neon
```diff
@@ -414,6 +414,12 @@ parameters:
 			count: 1
 			path: ../src/RuleSet/RuleSet.php
 
+		-
+			message: '#^Parameters should have "string" types as the only types passed to this method$#'
+			identifier: typePerfect.narrowPublicClassMethodParamType
+			count: 1
+			path: ../src/Value/CSSString.php
+
```
JakeQZ, line 422: I'm guessing this has arisen since we now have a caller (in the `TestCase`).

JakeQZ (approved):

PR #1002, opened 2025-02-26
[CLEANUP] Avoid Hungarian notation in `DeclarationBlock`

oliverklee:
Part of #756


src/RuleSet/DeclarationBlock.php
```diff
@@ -116,16 +116,16 @@ public function setSelectors($mSelector, $list = null): void
...
-        if ($mSelector instanceof Selector) {
-            $mSelector = $mSelector->getSelector();
+        if ($selectorToRemove instanceof Selector) {
+            $selectorToRemove = $selectorToRemove->getSelector();
         }
-        foreach ($this->aSelectors as $key => $selector) {
-            if ($selector->getSelector() === $mSelector) {
-                unset($this->aSelectors[$key]);
+        foreach ($this->selectors as $key => $subSelector) {
+            if ($subSelector->getSelector() === $selectorToRemove) {
```
JakeQZ, line 127: I think `$selector` singular was fine here for the loop variable.  `$subSelector` is a strange name.  (The parameter name change is good.)

JakeQZ (changes requested):
The changes to avoid variable name collisions are good, but I'm a bit unsure about one of the renames.

JakeQZ (approved):

PR #1004, opened 2025-02-26
[CLEANUP] Avoid Hungarian notation in `RuleSet`

oliverklee:
Also improve some names. (This class still in in urgent need of refactoring to improve type safety and make clear naming easier.)

Part of #756


oliverklee:
I'm not sure about the naming at all and would appreciate a critical eye (and suggestions).

src/RuleSet/RuleSet.php
```diff
@@ -101,33 +101,33 @@ public function getLineNo(): int
         return $this->lineNumber;
     }
 
-    public function addRule(Rule $rule, ?Rule $oSibling = null): void
+    public function addRule(Rule $ruleToAdd, ?Rule $sibling = null): void
     {
-        $sRule = $rule->getRule();
-        if (!isset($this->aRules[$sRule])) {
-            $this->aRules[$sRule] = [];
+        $rule = $ruleToAdd->getRule();
```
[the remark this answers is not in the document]
oliverklee: I've also added #1011 for renaming that property.

src/RuleSet/RuleSet.php
```diff
@@ -222,34 +223,34 @@ public function getRulesAssoc($mRule = null)
      * Note: this is different from pre-v.2.0 behaviour of PHP-CSS-Parser, where passing a Rule instance would
      * remove all rules with the same name. To get the old behaviour, use `removeRule($rule->getRule())`.
      *
-     * @param Rule|string|null $mRule
+     * @param Rule|string|null $removalPattern
```
JakeQZ, line 226: I think I would prefer `$searchPattern` again here.  The rules to be removed still have to be searched for.

src/RuleSet/RuleSet.php
```diff
@@ -222,34 +223,34 @@ public function getRulesAssoc($mRule = null)
...
      *        pattern to remove. If $mRule is null, all rules are removed. If the pattern ends in a dash,
```
JakeQZ, line 227: This mentions the parameter by name, and would also need updating.  Though could perhaps be modified to avoid being self-referential.

src/RuleSet/RuleSet.php
```diff
@@ -222,34 +223,34 @@ public function getRulesAssoc($mRule = null)
...
      *        excluded. Passing a Rule behaves matches by identity.
      */
-    public function removeRule($mRule): void
+    public function removeRule($removalPattern): void
     {
-        if ($mRule instanceof Rule) {
-            $sRule = $mRule->getRule();
-            if (!isset($this->aRules[$sRule])) {
+        if ($removalPattern instanceof Rule) {
+            $removalRule = $removalPattern->getRule();
```
JakeQZ, line 234: Maybe `$propertyToRemove` instead of `$removalRule`.

JakeQZ (changes requested):
I may have missed some changes in previous review.  I've suggested a few more naming improvements.

src/RuleSet/RuleSet.php
```diff
@@ -101,33 +101,33 @@ public function getLineNo(): int
...
+        if ($ruleToAdd->getLineNo() === 0 && $ruleToAdd->getColNo() === 0) {
             //this node is added manually, give it the next best line
             $rules = $this->getRules();
-            $pos = \count($rules);
-            if ($pos > 0) {
-                $last = $rules[$pos - 1];
-                $rule->setPosition($last->getLineNo() + 1, 0);
+            $rulesCount = \count($rules);
+            if ($rulesCount > 0) {
+                $last = $rules[$rulesCount - 1];
```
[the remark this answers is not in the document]
oliverklee: Indeed! I've created #1015 to track this.

src/RuleSet/RuleSet.php
```diff
@@ -222,34 +226,34 @@ public function getRulesAssoc($mRule = null)
...
-                if ($rule === $mRule) {
-                    unset($this->aRules[$sRule][$key]);
+            foreach ($this->rules[$propertyToRemove] as $key => $rule) {
+                if ($rule === $searchPattern) {
+                    unset($this->rules[$propertyToRemove][$key]);
                 }
             }
         } else {
-            foreach ($this->aRules as $sName => $aRules) {
+            foreach ($this->rules as $ruleName => $rules) {
```
JakeQZ, line 247: `$propertyName` rather than `$ruleName`

JakeQZ (changes requested):
Might have been better to have this PR in more manageable chunks.  But we are where we are, and we are nearly there.

I found one more variable that could be better-renamed.

src/RuleSet/RuleSet.php
JakeQZ: Actaully I'm not sure.

JakeQZ: If it's not a nested `RuleSet` then 'property name' is correct.  But I'm not sure otherwise.  Perhaps `$propertyOrRuleName` may be better throughout.

oliverklee:
Okay, I'll split this up.

src/RuleSet/RuleSet.php
```diff
@@ -103,31 +106,31 @@ public function getLineNo(): int
 
     public function addRule(Rule $ruleToAdd, ?Rule $sibling = null): void
     {
-        $sRule = $ruleToAdd->getRule();
-        if (!isset($this->rules[$sRule])) {
-            $this->rules[$sRule] = [];
+        $propertyOrRuleName = $ruleToAdd->getRule();
```
JakeQZ, line 109: Having looked into how this works and what it does, I now think just `$propertyName` would be better - see https://github.com/MyIntervals/PHP-CSS-Parser/pull/1051#pullrequestreview-2652886522.

oliverklee:
All the split-off PRs now are merged, and this PR now is quite small and ready for a re-review.

JakeQZ (approved):
Good rename to avoid a variable having two (slightly different) purposes.

Issue #998, opened 2025-02-25
Deprecate/drop `__toString()`


oliverklee:
We already have `render()` and don't need to duplicate this. (In addition, I would expect `__toString()` to render a technical meaningful representation, which it doesn't.)

PR #1006, opened 2025-02-26
[TASK] Deprecate `__toString()`

oliverklee:
Part of #998

### DIFF
--- a/tests/FunctionalDeprecated/Comment/CommentTest.php
+++ b/tests/FunctionalDeprecated/Comment/CommentTest.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\FunctionalDeprecated\Comment;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Comment\Comment;
+
+/**
+ * @covers \Sabberworm\CSS\Comment\Comment
+ */
+final class CommentTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function toStringRendersCommentEnclosedInCommentDelimiters(): void
+    {
+        $comment = 'There is no spoon.';
+        $subject = new Comment();
+
+        $subject->setComment($comment);
+
+        self::assertSame('/*' . $comment . '*/', (string) $subject);
+    }
+}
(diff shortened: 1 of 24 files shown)


tests/FunctionalDeprecated/Comment/CommentTest.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * @covers \Sabberworm\CSS\Comment\Comment
+ */
+final class CommentTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function toStringRendersCommentEnclosedInCommentDelimiters(): void
```
JakeQZ, line 18: Why is this method being moved here to test something that's deprecated?  Shouldn't the introduction of the test class be a separate PR?

JakeQZ (changes requested):
The deprecation notices are all fine, but the moving of a test would seem to belong in a separate PR (if required at all).

tests/FunctionalDeprecated/Comment/CommentTest.php
oliverklee: I'm moving this one test from `Tests/Functional/` to `Tests/FunctionalDeprecated/`, where tests for deprecated functionality should be placed. (This will allow us to have different PHPUnit configuration files later on, allowing calls to deprecated functionality in `Tests/FunctionalDeprecated/`, but not in  `Tests/Functional/`.)

JakeQZ (approved):
My mistake in previous review.  Though I do have a further question in the comments.

oliverklee:
> That said, I'm now wondering why the `CommentTest` class is in `Functional` rather than `Unit`, given that it appears to be a unit test for the `Comment` class - but that's beyond the scope of this PR.

I see the tests for `render` (and for `__toString`) as we're testing the tested class in conjunction with `OutputFormat` and `OutputFormatter` instead of the tested class in isolation.

JakeQZ:
> > That said, I'm now wondering why the `CommentTest` class is in `Functional` rather than `Unit`, given that it appears to be a unit test for the `Comment` class - but that's beyond the scope of this PR.
> 
> I see the tests for `render` (and for `__toString`) as we're testing the tested class in conjunction with `OutputFormat` and `OutputFormatter` instead of the tested class in isolation.

I (originally) saw the functional tests department as mainly a place for testing the various CSS conent that is in `tests/fixtures` that does not test any class directly.  But yes, I see in `CommentTest` we are testing the behaviour in conjunction with `OutputFormat`, which cannot be done in isolation.

PR #1008, opened 2025-02-26
[CLEANUP] Avoid Hungarian notation in `Settings`

oliverklee:
We'll change the direct accesses to use the accessors in a later chance.

Part of #756


JakeQZ (approved):
The properties have been marked as internal, so can now be renamed (provided we release 8.8).

PR #1013, opened 2025-02-27
[FEATURE] Add internal getters to `Settings` and use them

oliverklee:
Also make a type annotation more specific along the way.


JakeQZ (approved):
Neat.  This means the properties can now be made private.

PR #1017, opened 2025-02-27
[TASK] Make `Selector` a `Renderable`

oliverklee:
This is required to be able to drop `__toString` for this class.


JakeQZ (approved):
I see - ``DeclarationBlock::render()` does an implicit cast to string for the selectors.

I'm beginning to accept that `$pattern` can work as a variable name to for a selector string to distinguish it from a `Selector` object.

PR #1018, opened 2025-02-27
[TASK] Deprecate greedy calculation of selector specificity

oliverklee:
This constructor parameter is not used, and having the specificity calculation always done lazily is not a problem.


JakeQZ (approved):
Indeed, it can be done on first-time access without loss of performance.

PR #1022, opened 2025-02-27
[TASK] Deprecate greedy calculation of selector specificity

oliverklee:
This constructor parameter is not used, and having the specificity calculation always done lazily is not a problem.

This is the V8.x backport of #1018.


JakeQZ (approved):

JakeQZ:
> This constructor parameter is not used

I changed the commit message to "This constructor parameter will no longer be used", since it was being used.

PR #1024, opened 2025-02-27
[CLEANUP] Avoid Hungarian notation for `rules`

oliverklee:
Part of #756

### DIFF
--- a/src/RuleSet/RuleSet.php
+++ b/src/RuleSet/RuleSet.php
@@ -27,7 +27,7 @@ abstract class RuleSet implements Renderable, Commentable
     /**
      * @var array<string, Rule>
      */
-    private $aRules = [];
+    private $rules = [];
 
     /**
      * @var int<0, max>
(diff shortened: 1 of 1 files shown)


src/RuleSet/RuleSet.php
```diff
@@ -271,8 +271,8 @@ protected function renderRules(OutputFormat $outputFormat)
         $result = '';
         $isFirst = true;
         $oNextLevel = $outputFormat->nextLevel();
-        foreach ($this->aRules as $aRules) {
-            foreach ($aRules as $rule) {
+        foreach ($this->rules as $rules) {
+            foreach ($rules as $rule) {
```
JakeQZ, line 275: I'm struggling to understand how and why this can be doubly-iterated over - since `Rule` does not appear to be `Iterable`.

JakeQZ (approved):
The change is fine.  I'm struggling to understand the original implementation.

src/RuleSet/RuleSet.php
oliverklee: I don’t understand this either. Either this is broken (and not tested, or never called), or the code somehow still is working fine. 😉

Issue #1016, opened 2025-02-27
Rename `Selector.selector` to `.pattern`


oliverklee:
See the discussion and ideas in #1025.

PR #1025, opened 2025-02-27
[TASK] Add tests for the selector specificity

oliverklee:
Mostly move tests from `ParserTest` that belong into
the unit tests.

Part of #757
Part of #758

### DIFF
--- a/tests/Unit/Property/SelectorTest.php
+++ b/tests/Unit/Property/SelectorTest.php
@@ -35,4 +35,56 @@ final class SelectorTest extends TestCase
 
         self::assertSame($selector, $subject->getSelector());
     }
+
+    /**
+     * @return array<string, array{0: non-empty-string, 1: int<0, max>}>
+     */
+    public static function provideSelectorsAndSpecificities(): array
+    {
+        return [
+            'element' => ['a', 1],
+            'element and descendant with pseudo-selector' => ['ol li::before', 3],
+            'class' => ['.highlighted', 10],
+            'element with class' => ['li.green', 11],
+            'class with pseudo-selector' => ['.help:hover', 20],
+            'ID' => ['#file', 100],
+            'ID and descendant class' => ['#test .help', 110],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param non-empty-string $selector
+     * @param int<0, max> $expectedSpecificity
+     *
+     * @dataProvider provideSelectorsAndSpecificities
+     */
+    public function getSpecificityByDefaultReturnsSpecificityOfSelectorProvidedToConstructor(
+        string $selector,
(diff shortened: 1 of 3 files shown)


JakeQZ (changes requested):
I think there's a naming issue.  [MDN does indeed say](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_selectors):

> In CSS, selectors are patterns used to match, or select, the elements you want to style...

"pattern" is a description of how they work.  "selector" is the (official) name for what they are.  I think the test names should reflect what the 'object' is called (or known as), rather than what it does or how it works.

Wittgenstein might disagree (if I sit on a table, it's now a chair), but he wasn't a computer scientist.

Also, I'd like this PR to be split for more manageable review - the addition of the deprecated test class can be done separately I think.

oliverklee:
I'd like to find a way around that a selector can have a selector (which would mean that we need to deal with a selector's selector) … any ideas?

JakeQZ:
> I'd like to find a way around that a selector can have a selector (which would mean that we need to deal with a selector's selector) … any ideas?

The `Selector` class represents and/or encapsulates a selector, which in the rendered or to-be-parsed CSS is just a string like `a:hover`.  I guess the string property of the class could be named something like `$rawContent` or `$stringRepresentation`, or something better that I can't come up with right now.  Note we alse have `Value.value`, and others (e.g. `Rule.rule`).

oliverklee:
Okay, then I'll keep using `selector` for the tests for now, and we do the renaming in #1016. I'll also split this up into smaller chunks.

oliverklee:
The other two PRs are merged so that now this PR is a lot smaller. Rebased, reworded and repushed.

tests/Unit/Property/SelectorTest.php
```diff
@@ -35,4 +35,56 @@ public function setSelectorOverwritesSelectorProvidedToConstructor(): void
...
+
+    /**
+     * @test
+     *
+     * @param non-empty-string $selector
+     * @param int<0, max> $expectedSpecificity
+     *
+     * @dataProvider provideSelectorsAndSpecificities
+     */
+    public function getSpecificityByReturnsSpecificityOfSelectorLastProvidedViaSetSelector(
```
JakeQZ, line 80:
```suggestion
    public function getSpecificityReturnsSpecificityOfSelectorLastProvidedViaSetSelector(
```

JakeQZ (changes requested):
There's a typo in a test method name.

Specificites cannot be represented by a single number, so we should address that in future...

JakeQZ:
> Specificites cannot be represented by a single number, so we should address that in future...

Added #1042 to capture this.

JakeQZ (approved):

PR #1035, opened 2025-02-28
[TASK] Add tests for `Selector::__toString()`

oliverklee:
Part of #757


JakeQZ (approved):
Observing that this is in the `FunctionalDeprecated` section.

PR #1037, opened 2025-02-28
[TASK] Mark `Selector::isValid()` as `@internal`

oliverklee:


JakeQZ (approved):
It does indeed seem to be used internally, right before constructing a `Selector` - rather suggesting that the `Selector` constructor should be making this check itself instead (beyond the scope of this PR).

JakeQZ:
> ... rather suggesting that the `Selector` constructor should be making this check itself instead (beyond the scope of this PR).

Added #1041 to capture this.

PR #1038, opened 2025-02-28
[TASK] Drop `getLineNo()` from the `Renderable` interface

oliverklee:
The concept of being able to get rendered is logically independent from having a position in the input stream.

Hence, both concepts should not be intermingled in the `Renderable` interface.

(We might want to add a dedicated interface for classes that are linked to specific points in the input stream later, though.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ Please also have a look at our
 
 ### Removed
 
+- Drop `getLineNo()` from the `Renderable` interface (#1038)
 - Remove `OutputFormat::level()` (#874)
 - Remove expansion of shorthand properties (#838)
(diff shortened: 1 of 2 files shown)


JakeQZ (approved):
No deprecation notice required, since classes implementing the original interface will still have the method.  (And PHP is not strict about whether a method is defined in the parameter type expected, as long as it exists in the actual class of object provided.)

JakeQZ:
> intermingled

I changed this in the commit message to "conflated", which is possibly more accurate, though "intermingled" was fine.

oliverklee:
> I changed this in the commit message to "conflated", which is possibly more accurate, though "intermingled" was fine.

Thanks - always glad to work with a native speaker! ❤️

PR #1046, opened 2025-03-01
[TASK] Remove `__toString()`

### LINKED ISSUE #998: Deprecate/drop `__toString()`

oliverklee:
Closes #998.


JakeQZ (approved):
I've seen you rebase this so many times - 23 apparently.  Glad to see it finally completed.

PR #1047, opened 2025-03-01
[TASK] Move up `getAllRuleSets` to `CSSBlockList`

oliverklee:
Move this method up without any changes to the method or its callers.

We'll clean this up in later changes.

Part of #994.


JakeQZ (approved):
Confirmed the moved tests all match up.

PR #1051, opened 2025-03-02
[BUGFIX] Fix a type annotation in `RuleSet`

oliverklee:


JakeQZ (commented):
I figured out why the `Rule`s in `RuleSet` are a nested array.  The top-level key is the property name.  But since it's possible to have multiple rules for the same property, the values are (non-associative) arrays of `Rule`s.  The problem with this structure is that the ordering is lost, and we have #1052.

Anyway, I think the keys are always property names (whatever the `RuleSet` type), and there isn't really a concept of a 'rule name'.

oliverklee:
I've updated the comment accordingly (and also the variables names in #1039).

PR #1059, opened 2025-03-03
[BUGFIX] Render rules in line and column number order

### LINKED ISSUE #1052: CSS properties order
When CSS is parsed, the properties order is not maintained.

Example : 

$css = '.css-109zgpm-MuiButtonBase-root-MuiButton-root{
        background-color:transparent;
        background:#222222;
        background-color:#FFFFFF;
        }';

$parser = (new Parser($css))->parse();

$css = $parser->render();

Output :
.css-109zgpm-MuiButtonBase-root-MuiButton-root {
background-color: transparent;
background-color: #fff;
background: #222;
}

And so the background does not keep the same color override

How can we keep the order?

oliverklee:
Fixes #1052
Closes #1058


tests/Functional/RuleSet/DeclarationBlockTest.php
```diff
@@ -0,0 +1,41 @@
...
+
+        $rule2 = new Rule('background');
+        $rule2->setValue('#222');
+        $declarationBlock->addRule($rule2);
+
+        $rule3 = new Rule('background-color');
+        $rule3->setValue('#fff');
+        $declarationBlock->addRule($rule3);
+
+        $expectedRendering = 'background-color: transparent;background: #222';
```
JakeQZ, line 38: Should this include the final `background-color`, or is it sufficient to assert that it doesn't come immediately after the first one?  I think it is probably sufficient as is.

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@ Please also have a look at our
 
 ### Fixed
 
+- Render rules in line and column number order (#1058)
```
JakeQZ, line 57: This would need to be changed to #1059.

JakeQZ (changes requested):
This seems to work.  I had overlooked a couple of things:
- `RuleSet::addRule()` automatically sets the line number to one more than the previous highest, so they don't need setting explicitly;
- The values can be simply set a strings, and don't need the creation of `Value` objects.

If you change the PR number in the changelog, and review my other comment to double-check, I'd be happy for this to supersede #1058.

tests/Functional/RuleSet/DeclarationBlockTest.php
oliverklee: I'll add it - better make the test more strict.

JakeQZ (approved):

oliverklee:
@JakeQZ Should we backport this? What do you think?

PR #1062, opened 2025-03-04
[BUGFIX] Render rules in line and column number order

oliverklee:
This is the V8.x backport of #1059.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- Render rules in line and column number order (#1059)
 - Create `Size` with correct types in `expandBackgroundShorthand` (#814)
 - Parse `@font-face` `src` property as comma-delimited list (#794)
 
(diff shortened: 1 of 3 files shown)


oliverklee:
This seems to break (or change) shorthand expansion. Also, the results seem be depend on the PHP version (2 failures with PHP 8.4, 17 failures with PHP 5.6).

I'd suggest we don't backport this after all as getting this to work consistently across PHP versions might not be worth the hassle. (And I'm not keen on doing the work.)

What do you think, @JakeQZ?

oliverklee:
… or we remove the failing tests, or we remodel them to not care about the order of things anymore.

JakeQZ:
> This seems to break (or change) shorthand expansion. Also, the results seem be depend on the PHP version (2 failures with PHP 8.4, 17 failures with PHP 5.6).

The tests seem to expect a spefic ordering of the expanded shorthand properties, when in fact it doesn't matter.  Since they all came from the same original property, they probably have the same line number, so when passed to `usort`, the resulting order could be any.  The number of failures with different PHP versions probably depends on the internal implementation of `usort` for the PHP version.

> or we remodel them to not care about the order of things anymore.

That would allow the backport.

> (And I'm not keen on doing the work.)

I'm happy to try to create a pre-PR to change the tests.

JakeQZ (approved):

PR #1065, opened 2025-03-04
Allow rules to be in any order in shorthand tests

JakeQZ:
This is a pre-patch for #1062.

### DIFF
--- a/tests/RuleSet/DeclarationBlockTest.php
+++ b/tests/RuleSet/DeclarationBlockTest.php
@@ -29,7 +29,7 @@ final class DeclarationBlockTest extends TestCase
         foreach ($oDoc->getAllDeclarationBlocks() as $oDeclaration) {
             $oDeclaration->expandBorderShorthand();
         }
-        self::assertSame(trim((string)$oDoc), $sExpected);
+        self::assertDeclarationBlockEquals(trim((string)$oDoc), $sExpected);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


tests/RuleSet/DeclarationBlockTest.php
```diff
@@ -505,4 +505,50 @@ public function canRemoveCommentsFromRulesUsingStrictParsing(
...
+     *
+     * @param string $declarationBlock
+     *
+     * @return string
+     */
+    private static function sortRulesInDeclarationBlock($declarationBlock)
+    {
+        // Match everything between `{` and `}`.
+        return \preg_replace_callback(
+            '/(?<=\{)[^\}]*+(?=\})/',
```
oliverklee, line 534: Let's escape the backslashes to make it clear to the reader that they aren't intended to escape them for the PHP string, but for the regular expression.

```suggestion
            '/(?<=\\{)[^\\}]*+(?=\\})/',
```

oliverklee (changes requested):

tests/RuleSet/DeclarationBlockTest.php
JakeQZ: The backslashes are escaping the `{` and `}`.

JakeQZ: Ah, yes, I see.

oliverklee (approved):

PR #1077, opened 2025-03-04
[CLEANUP] Avoid Hungarian notation for `value`

oliverklee:
Part of #756

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -251,35 +251,35 @@ class ParserState
     }
 
     /**
-     * @param int $mValue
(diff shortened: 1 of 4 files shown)


src/Parsing/ParserState.php
```diff
@@ -251,35 +251,35 @@ public function peek($length = 1, $offset = 0): string
     }
 
     /**
-     * @param int $mValue
```
JakeQZ, line 254: Looks like the type here should be `int|string`, but that's beyond the scope of this PR.

JakeQZ (approved):

PR #1083, opened 2025-03-05
[TASK] Drop special support for vendor prefixes

oliverklee:
In the past, vendor prefixes like `-moz-` or `-webkit-` were used for experimental CSS features in browsers. Nowadays, the browsers use features for this instead. Hence, special support for vendor prefixes is no longer needed.

https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ Please also have a look at our
 
 ### Removed
 
+- Drop special support for vendor prefixes (#1083)
 - Remove the IE hack in `Rule` (#995)
 - Drop `getLineNo()` from the `Renderable` interface (#1038)
 - Remove `OutputFormat::level()` (#874)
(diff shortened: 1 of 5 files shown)


JakeQZ (commented):
With this change, the vendor-prefixed `calc` will be parsed as vanilla `CSSFunction`s, which will work in some cases, but not all - e.g. `-moz-calc(( 50px - 50% ) * 2)` will be rejected at the `(` inside the function argument, but the remainder of the value will still be parsed, resulting in `( 50px - 50px` in this case.

Since the vendor prefixes are still used in legacy CSS, which may need to be parsed (e.g. to automatically update it to use modern constructs), I think we need to first deprecate the vendor-prefixed `calc` in 8.8.  I'll do that now...

JakeQZ (approved):
This can be merged now they are deprecated for v8.8 in #1086.

PR #1085, opened 2025-03-05
[CLEANUP] Avoid Hungarian notation for `size`

oliverklee:
Part of #756

### DIFF
--- a/src/Value/Size.php
+++ b/src/Value/Size.php
@@ -55,7 +55,7 @@ class Size extends PrimitiveValue
     /**
      * @var float
      */
-    private $fSize;
+    private $size;
 
     /**
      * @var string|null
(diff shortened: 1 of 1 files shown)


JakeQZ (changes requested):
As we're renaming this, we can change it to `$number` to match its official name:
- https://developer.mozilla.org/en-US/docs/Web/CSS/length;
- https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/rgb.

oliverklee:
> As we're renaming this, we can change it to `$number` to match its official name:

I'm all for this. However, I'd prefer to do this in a separate change as this involves renaming the setter/getter (if we want to do it properly), which I'd prefer not to do as a drive-by change of a no-changelog-entry cleanup. I've created #1087 for this.

JakeQZ (approved):
> > As we're renaming this, we can change it to `$number` to match its official name:
> 
> I'm all for this. However, I'd prefer to do this in a separate change as this involves renaming the setter/getter (if we want to do it properly), which I'd prefer not to do as a drive-by change of a no-changelog-entry cleanup. I've created #1087 for this.

We can't so easily rename the getter and setter since they are public methods.  To do so (seamlessly) would mean adding the newly named methods and having the old methods chain onto the new ones while the old ones are deprecated.

PR #1089, opened 2025-03-05
[CLEANUP] Avoid Hungarian notation for `tokenLength`

oliverklee:
Part of #756


JakeQZ (changes requested):
Sorry, I missed the original review request for this.

I don't think `$size` is right.  It's actually the length of the unit identifier token string.

oliverklee:
Then maybe `$length`?

JakeQZ (approved):

PR #1098, opened 2025-03-06
[CLEANUP] Avoid Hungarian notation in `OutputFormat` (part 1)

oliverklee:
Part of #756

### DIFF
--- a/src/OutputFormat.php
+++ b/src/OutputFormat.php
@@ -18,7 +18,7 @@ class OutputFormat
      *
      * @var bool
      */
-    private $bRGBHashNotation = true;
+    private $rgbHashNotation = true;
 
     /**
      * Declaration format
@@ -27,7 +27,7 @@ class OutputFormat
      *
      * @var bool
      */
-    private $bSemicolonAfterLastRule = true;
+    private $semicolonAfterLastRule = true;
 
     /**
      * Spacing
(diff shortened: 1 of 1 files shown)


src/OutputFormat.php
```diff
@@ -18,7 +18,7 @@ class OutputFormat
      *
      * @var bool
      */
-    private $bRGBHashNotation = true;
+    private $rgbHashNotation = true;
```
JakeQZ, line 21:
```suggestion
    private $useRgbHashNotation = true;
```

src/OutputFormat.php
```diff
@@ -27,7 +27,7 @@ class OutputFormat
      *
      * @var bool
      */
-    private $bSemicolonAfterLastRule = true;
+    private $semicolonAfterLastRule = true;
```
JakeQZ, line 30:
```suggestion
    private $renderSemicolonAfterLastRule = true;
```

JakeQZ (approved):
I made some suggestions for clearer names for the Boolean properties (to clarify that they are Booleans).  However, this would mean that they are no longer consistent with the getters and setters.  And given that these don't correspond to any formal CSS data type, I don't think it's worth the effort renaming them - unlike in the case of `Size.number`.

PR #1100, opened 2025-03-06
[CLEANUP] Avoid Hungarian notation in `OutputFormat` (part 2)

oliverklee:
Part of #756

### DIFF
--- a/src/OutputFormat.php
+++ b/src/OutputFormat.php
@@ -43,17 +43,17 @@ class OutputFormat
     /**
      * @var string
      */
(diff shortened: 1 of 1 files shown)


oliverklee:
Sigh. This clashes with the magic in `OutputFormat::set()`. We'll need to deprecate and remove that function first.

JakeQZ (approved):

oliverklee:
And before that, we'll need to update the tests to no longer use `set()`/`get()`.

PR #1102, opened 2025-03-06
[CLEANUP] Avoid Hungarian notation in `SpecificityCalculator`

oliverklee:
Part of #756


JakeQZ (approved):
This code was moved from the original class without change when the class was created.

Issue #1103, opened 2025-03-06
Get rid of `OutputFormat::get()` and `::set()`


oliverklee:
- [x] deprecate both in V8.x
- [x] drop the (unused) `get()`
- [x] rework `OutputFormatTest` to not use `set()` anymore
- [ ] drop `set()`

PR #1106, opened 2025-03-07
[CLEANUP] Use the explicit `OutputFormat` setters in the tests

oliverklee:
The `set()` method will be removed soon.

Also unify the tests a bit.

Part of #1103

### DIFF
--- a/tests/OutputFormatTest.php
+++ b/tests/OutputFormatTest.php
@@ -182,6 +182,11 @@ EOT;
      */
     public function spaceRules(): void
     {
+        $outputFormat = OutputFormat::create()
+            ->setSpaceBeforeRules("\n")
+            ->setSpaceBetweenRules("\n")
+            ->setSpaceAfterRules("\n");
+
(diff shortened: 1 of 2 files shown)


tests/OutputFormatTest.php
```diff
@@ -280,6 +310,8 @@ public function ignoreExceptionsOff(): void
     {
         $this->expectException(OutputException::class);
 
+        $outputFormat = OutputFormat::create()->setIgnoreExceptions(false);
```
JakeQZ, line 313: This test uses two separate `OutputFormat`s.

JakeQZ (changes requested):
One test uses two separate `OutputFormat`s.

tests/OutputFormatTest.php
JakeQZ: > it doesn't seem relevant to me to use two instances with identical settings

Me neither - given that it's just testing that an exception is thrown when attempting to render a declaration block without a selector.

PR #1107, opened 2025-03-07
[TASK] Deprecate `OutputFormat::get()` and `::set()`

oliverklee:
Part of #1103

### DIFF
--- a/src/OutputFormat.php
+++ b/src/OutputFormat.php
@@ -268,7 +268,7 @@ class OutputFormat
      *
      * @return self|false
      *
-     * @deprecated since 8.8.0, will be removed in 9.0.0. Use specific magic setters instead.
+     * @deprecated since 8.8.0, will be removed in 9.0.0. Use specific setters instead.
      */
     public function set($aNames, $mValue)
     {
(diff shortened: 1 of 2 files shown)


src/OutputFormat.php
```diff
@@ -247,6 +247,8 @@ public function __construct()
      * @param string $sName
      *
      * @return string|null
+     *
+     * @deprecated since 8.8.0, will be removed in 9.0.0. Use specific magic getters instead.
```
JakeQZ, line 251:
```suggestion
     * @deprecated since 8.8.0, will be removed in 9.0.0. Use specific getters instead.
```

src/OutputFormat.php
```diff
@@ -265,6 +267,8 @@ public function get($sName)
      * @param mixed $mValue
      *
      * @return self|false
+     *
+     * @deprecated since 8.8.0, will be removed in 9.0.0. Use specific magic setters instead.
```
JakeQZ, line 271:
```suggestion
     * @deprecated since 8.8.0, will be removed in 9.0.0. Use specific setters instead.
```

JakeQZ (changes requested):
We're not in the The Magic Circle any more.

oliverklee:
And there goes the magic … :wink:

JakeQZ (approved):

PR #1110, opened 2025-03-07
[TASK ] Remove `OutputFormat::set()`

### LINKED ISSUE #1103: Get rid of `OutputFormat::get()` and `::set()`

oliverklee:
Part of #1103
Closes #1103

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,7 +43,7 @@ Please also have a look at our
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


JakeQZ (approved):
There weren't too many worms in the can.

We can now avoid Hungarian notation for the properties.

PR #1115, opened 2025-03-08
[CLEANUP] Avoid Hungarian notation in `OutputFormat` (part 4)

oliverklee:
Also rename the (internal) getters to match the changed property names.

Part of #756


JakeQZ (approved):
Good rename.

PR #1118, opened 2025-03-08
[CLEANUP] Avoid Hungarian notation in `OutputFormat` (part 6)

oliverklee:
Part of #756


JakeQZ (approved):
I'm wondering if the descriptions should (also) be in the DocBlocks for the setters, since those are the public API, but that's beyond the scope of this PR.

PR #1124, opened 2025-03-08
[CLEANUP] Avoid Hungarian notation for `stopCharacters`

oliverklee:
Part of #756


JakeQZ (approved):
Good rename.

It seems that when this parameter is a string, it must be a single character, not a string of possible stop characters (which might be expected to work, though is beyond the scope of this PR).

PR #1127, opened 2025-03-09
[CLEANUP] Drop empty `OutputFormat` constructor

oliverklee:


JakeQZ (commented):
If a user extends `OutputFormat` (for what reason, I don't know), this could be a breaking change, since calling `parent::construct()` requires a constructor to be explicitly defined.

So I think we might first need to deprecate `OutputFormat` being extensible.  Then when removing the constructor, also make it `final`.

PR #1128, opened 2025-03-09
[TASK] Make `OutputFormat` `final`

oliverklee:
This class is not intended to be extended, and it's intended to be created using the factory methods anyway.

(Also, we most probably won't need to have mocks of this class in the tests as it's mostly a data object.)


JakeQZ (commented):
As per https://github.com/MyIntervals/PHP-CSS-Parser/pull/1127#pullrequestreview-2669436009, should we deprecate the extensibility first?

JakeQZ (approved):

PR #1131, opened 2025-03-09
[TASK] Mark `OutputFormat` as not extendable

oliverklee:
Also mark the constructor as `@internal`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,7 @@ Please also have a look at our
 
 ### Changed
 
-- Mark `OutputFormat` as not extendable and the constructor as `@internal`
-  (#1131) 
+- Mark the `OutputFormat` constructor as `@internal` (#1131) 
 - Mark `OutputFormatter` as `@internal` (#896)
 - Make `Selector` a `Renderable` (#1017)
 - Mark `Selector::isValid()` as `@internal` (#1037)
@@ -37,6 +36,7 @@ Please also have a look at our
 
 ### Deprecated
 
+- Deprecate extending `OutputFormat` (#1131)
 - Deprecate magic method forwarding from `OutputFormat` to `OutputFormatter`
   (#894)
 - Deprecate greedy calculation of selector specificity (#1018)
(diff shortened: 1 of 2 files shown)


src/OutputFormat.php
```diff
@@ -4,6 +4,9 @@
 
 namespace Sabberworm\CSS;
 
+/**
+ * Please note that this class is not intended to be subclassed.
```
JakeQZ, line 8:
```suggestion
 * Extending this class is deprecated in version 8.8.0; it will be made `final` in version 9.0.0.
```

src/OutputFormat.php
```diff
@@ -180,6 +183,9 @@ class OutputFormat
      */
     private $indendationLevel = 0;
 
+    /**
+     * @internal since V8.8.0 Use the factory methods `create()`, `createCompact()`, or `createPretty()` instead.
```
JakeQZ, line 187:
```suggestion
     * @internal since V8.8.0. Use the factory methods `create()`, `createCompact()`, or `createPretty()` instead.
```

CHANGELOG.md
```diff
@@ -21,6 +21,8 @@ Please also have a look at our
 
 ### Changed
 
+- Mark `OutputFormat` as not extendable and the constructor as `@internal`
+  (#1131)
```
JakeQZ, line 25: If my suggestion for the class DocBlock is accepted, this probably warrants a second changelog entry in the 'Deprecated' section.  (But I don't think two separate PRs are needed!)

JakeQZ (changes requested):
There's a missing full-stop in the `@internal` notice.

Also, I think we can be clearer in the class DocBlock that it's going to be made `final`, with extensibility being deprecated.

JakeQZ:
Also, I think these changes could be made directly on the 8.x branch only, as on `main` they will be immediately removed.

oliverklee:
I have just created #1133 for the backport.

As far as having this in `main` as well, I'd be fine either way (i.e., have it in `main` and `v8.x` or only `v8.x`).

JakeQZ (approved):

PR #1132, opened 2025-03-09
[BUGFIX] Fix a typo in `indentationLevel`

oliverklee:


JakeQZ (approved):
Good spot (finally).  We both missed this in #1126.

PR #1134, opened 2025-03-10
[TASK] Use native type declarations for `$lineNumber`

oliverklee:
Part of #811


JakeQZ (changes requested):
There's a duplicated parameter name in one DocBlock.

I made some suggestions, perhaps for a separate PR, where I think `$lineNumber` is necessarily always at least `1`.  Though given that this PR already includes one change to `int<1, max>`, perhaps those should be included here.  I'll leave it to your discretion...

JakeQZ (approved):

PR #1136, opened 2025-03-11
[TASK] Use native type declarations in `ParserState`

oliverklee:
Part of #811

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -33,7 +33,7 @@ class ParserState
     private $characters;
 
     /**
-     * @var int
+     * @var int<0, max>
      */
     private $currentPosition = 0;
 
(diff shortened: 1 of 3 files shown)


src/Parsing/ParserState.php
```diff
@@ -487,13 +459,11 @@ private function strsplit($string)
     }
 
     /**
-     * @param string $haystack
-     * @param string $needle
-     * @param int $offset
+     * @param int<0, max> $offset
```
JakeQZ, line 462: `$offset` can be negative to indicate an offset from the end of the string.  (Note that the version of `substr` provided in this class does not support a negative offset, whereas this method does because it calls on to the built-in PHP functions.)

JakeQZ (changes requested):
`$offset` for `strpos` can be negative.  Otherwise changes look fine.

src/Parsing/ParserState.php
```diff
@@ -487,13 +459,11 @@ private function strsplit($string)
     }
 
     /**
-     * @param string $haystack
-     * @param string $needle
      * @param int $offset
      *
```
JakeQZ, line 463:
```suggestion
```

We don't need the `@param` in the DocBlock any more.

JakeQZ (changes requested):
We can lose the `@param` in the DocBlock for `$offset` now.

JakeQZ (approved):

PR #1142, opened 2025-03-11
[TASK] Drop `atRuleArgs()` from the `AtRule` interface

oliverklee:
The classes implementing this interface use a wide variaty of inconsistent return types when they implement this method.

This is a hard blocker for introducing native return type declaration for this method.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,6 +47,7 @@ Please also have a look at our
 
 ### Removed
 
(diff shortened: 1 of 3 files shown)


JakeQZ (approved):
I can't think of a better solution which doesn't involve a breaking change.

Most implementations return a string, but `CSSNamespace` and `Import` return an array of strings.  Changing one or other to match would be a breaking change, since they are part of the public API.

PR #1143, opened 2025-03-11
[CLEANUP] Make a type annotation more specific

oliverklee:


JakeQZ (approved):
Can do the same for `SET_RULES`.

JakeQZ:
> Can do the same for `SET_RULES`.

Ah, I see in #1144, this is not used.

PR #1144, opened 2025-03-11
[CLEANUP] Drop an unused internal constant

oliverklee:


JakeQZ (approved):
I think this may have been more of a documentary constant, just showing what the other at-rules are.

PR #1145, opened 2025-03-11
[TASK] Add native type declarations for `atRuleName()`

oliverklee:
Part of #811


src/CSSList/KeyFrame.php
```diff
@@ -7,10 +7,14 @@
 use Sabberworm\CSS\OutputFormat;
 use Sabberworm\CSS\Property\AtRule;
 
+/**
+ * Note: When constructing an instance of this class, `setVendorKeyFrame()` must be called directly after construction
+ * for the instance to be valid.
+ */
```
JakeQZ, line 13: This feels messy to me.  The same will apply to `$animationName`.

JakeQZ (commented):
I think `KeyFrame` needs a constructor first.  I'll do that now...

Other changes look fine (and could be a separate PR).

JakeQZ:
> I think `KeyFrame` needs a constructor first. I'll do that now...

#1146

I was about to commit a change adding a constructor, then realized that is not necessary - all that's needed is for the properties to have default non-empty-string values.

JakeQZ (approved):

PR #1162, opened 2025-03-13
[TASK] Add native type declarations in `Size`

oliverklee:
Part of #811


src/Value/Size.php
```diff
@@ -125,9 +121,9 @@ public static function parse(ParserState $parserState, $isColorComponent = false
     }
 
     /**
-     * @return array<int, array<string, string>>
+     * @return array<int<1, max>, array<lowercase-string, non-empty-string>>
```
JakeQZ, line 124: Nice and precise.

JakeQZ (approved):

PR #1164, opened 2025-03-13
[TASK] Add some basic unit tests for `URL`

oliverklee:
Part of #757

### DIFF
--- a/tests/Unit/Value/URLTest.php
+++ b/tests/Unit/Value/URLTest.php
@@ -73,7 +73,7 @@ final class URLTest extends TestCase
     /**
      * @test
      */
-    public function getLineReturnsLineNumberProvidedToConstructor(): void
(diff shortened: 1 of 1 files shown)


tests/Unit/Value/URLTest.php
```diff
@@ -0,0 +1,84 @@
...
+    {
+        $subject = new URL(new CSSString('http://example.com'));
+
+        self::assertSame(0, $subject->getLineNo());
+    }
+
+    /**
+     * @test
+     */
+    public function getLineReturnsLineNumberProvidedToConstructor(): void
```
JakeQZ, line 76:
```suggestion
    public function getLineNoReturnsLineNumberProvidedToConstructor(): void
```

JakeQZ (changes requested):
I think a test method is missing part of the name of the method being tested.

JakeQZ (approved):

PR #1169, opened 2025-03-15
[BUGFIX] Include comments for all rules in declaration block

### LINKED ISSUE #173: bug with parsing multiple comments

JakeQZ:
- `Rule::parse()` will no longer consume anything after the semicolon terminating the rule - it does not belong to that rule;
- The whitespace and comments before a rule will be processed by `RuleSet::parseRuleSet()` and passed as a parameter to `Rule::parse()` -
  - This is only required while 'strict mode' parsing is an option, to avoid having an exception thrown during normal operation (i.e. when `Rule::parse()` encounters normal `}` as opposed to some other junk, which is not distinguished).

Fixes #173.
See also #672, #741.

### DIFF
--- a/src/Rule/Rule.php
+++ b/src/Rule/Rule.php
@@ -68,14 +68,16 @@ class Rule implements Renderable, Commentable
     }
 
     /**
+     * @param list<Comment> $commentsBeforeRule
+     *
      * @throws UnexpectedEOFException
      * @throws UnexpectedTokenException
      *
      * @internal since V8.8.0
      */
-    public static function parse(ParserState $parserState): Rule
+    public static function parse(ParserState $parserState, array $commentsBeforeRule = []): Rule
     {
-        $comments = $parserState->consumeWhiteSpace();
+        $comments = \array_merge($commentsBeforeRule, $parserState->consumeWhiteSpace());
         $rule = new Rule(
             $parserState->parseIdentifier(!$parserState->comes('--')),
             $parserState->currentLine(),
@@ -97,8 +99,6 @@ class Rule implements Renderable, Commentable
         while ($parserState->comes(';')) {
             $parserState->consume(';');
         }
-
-        $parserState->consumeWhiteSpace();
 
         return $rule;
     }
--- a/src/RuleSet/RuleSet.php
+++ b/src/RuleSet/RuleSet.php
@@ -65,11 +65,15 @@ abstract class RuleSet implements Renderable, Commentable
         while ($parserState->comes(';')) {
             $parserState->consume(';');
         }
-        while (!$parserState->comes('}')) {
+        for (;;) {
+            $commentsBeforeRule = $parserState->consumeWhiteSpace();
+            if ($parserState->comes('}')) {
+                break;
+            }
             $rule = null;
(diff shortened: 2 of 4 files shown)


tests/ParserTest.php
```diff
@@ -1103,8 +1103,6 @@ public function flatCommentExtractingOneComment(): void
      */
     public function flatCommentExtractingTwoComments(): void
     {
-        self::markTestSkipped('This is currently broken.');
-
         $parser = new Parser('div {/*Find Me!*/left:10px; /*Find Me Too!*/text-align:left;}');
```
oliverklee, line 1106: I think we also should have some more tests:
- two comments directly after another without whitespace
- two comments directly after another, but separated by whitespace
- three comments

oliverklee (changes requested):
Looks good! I think we should have some more testcases, though.

tests/ParserTest.php
JakeQZ: > I think we also should have some more tests:

I was also thinking the same.  Have added the first two suggestions.

For consistency, I've continued with the original method-naming pattern and location for now.  Though I don't know what `flat` is supposed to mean, and I think these tests belong in functional tests for either `Rule` or `RuleSet` (not sure which), but that is beyond the scope of this PR.

>     * three comments

Not sure what you had in mind here.  If one works and two works, is that not sufficient?  How far do we go (and [how long is](https://dictionary.cambridge.org/dictionary/english/how-long-is-a-piece-of-string) [a piece of string](https://www.youtube.com/watch?v=T1TYiixJ9P8))?

oliverklee (approved):
LGTM.

src/RuleSet/RuleSet.php
```diff
@@ -65,11 +65,15 @@ public static function parseRuleSet(ParserState $parserState, RuleSet $ruleSet):
         while ($parserState->comes(';')) {
             $parserState->consume(';');
         }
-        while (!$parserState->comes('}')) {
+        for (;;) {
```
oliverklee, line 68: Let's communicate the intention behind this line more clearly:

```suggestion
        while (true) {
```

src/Rule/Rule.php
```diff
@@ -68,14 +68,16 @@ public function __construct($rule, int $lineNumber = 0, $columnNumber = 0)
     }
 
     /**
+     * @param list<Comment> $commentsBeforeRule
```
oliverklee, line 71: For this to be consistent on higher PHPStan levels, we also need to change the return type annotation of `ParserState::consumeWhiteSpace()` to be a `list`.

oliverklee (changes requested):
Added two more nitpicks.

tests/ParserTest.php
oliverklee: After thinking about this some more, I think testing for 2 comments should suffice.

src/Rule/Rule.php
oliverklee: Done in #1173.

oliverklee:
@JakeQZ Would you be willing to take care of creating the backport?

PR #1171, opened 2025-03-16
[CLEANUP] Make `Commentable` type annotations more specific

oliverklee:


src/Comment/Commentable.php
```diff
@@ -7,17 +7,17 @@
 interface Commentable
 {
     /**
-     * @param array<array-key, Comment> $comments
+     * @param array<int<0, max>, Comment> $comments
```
JakeQZ, line 10: Can this be a `list` after #1173 is merged?

JakeQZ (commented):
Can the parameter of `addComments` be a `list` after #1173 is merged?

src/Comment/Commentable.php
oliverklee: Probably. I'll recheck once #1173 is merged.

JakeQZ (approved):

PR #1172, opened 2025-03-16
[TASK] Add native type declarations for `Import`

oliverklee:
Part of #811

### DIFF
--- a/src/Property/Import.php
+++ b/src/Property/Import.php
@@ -19,7 +19,7 @@ class Import implements AtRule
     private $location;
 
     /**
-     * @var string
+     * @var string|null
      */
     private $mediaQuery;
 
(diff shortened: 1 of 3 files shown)


src/Property/Import.php
```diff
@@ -19,7 +19,7 @@ class Import implements AtRule
     private $location;
 
     /**
-     * @var string
+     * @var string|null
```
JakeQZ, line 22: Could this be `?string`?

JakeQZ (commented):
Looking at the code, it seems an empty string is used to indicate 'no media query', but perhaps `null` is clearer.  The constructor could perhaps have that as a default value, but maybe as a separate PR.

src/Property/Import.php
oliverklee: No, the `?string` notation only works for native type declarations, but not for PHPDoc annotations.

oliverklee:
> Looking at the code, it seems an empty string is used to indicate 'no media query'

There is some calling code that [explicitly passes `null`](https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/src/CSSList/CSSList.php#L179). We could change the code use `non-empty-string|null` for `$mediaQuery`. What do think?

JakeQZ (approved):
> There is some calling code that [explicitly passes `null`](https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/src/CSSList/CSSList.php#L179). We could change the code use `non-empty-string|null` for `$mediaQuery`. What do think?

I've added #1177 to capture this.

PR #1173, opened 2025-03-16
[CLEANUP] Clean up `ParserState` a bit

oliverklee:


src/Parsing/ParserState.php
```diff
@@ -415,20 +419,22 @@ private function substr(int $offset, int $length): string
             $offset++;
             $length--;
         }
+
         return $result;
     }
 
+    /**
+     * @return ($string is non-empty-string ? non-empty-string : string)
```
JakeQZ, line 427: Didn't know this was possible.

JakeQZ (commented):
LGTM, though I wonder if the return type for `consumeComment` should also be changed as part of this PR - ?

oliverklee:
> LGTM, though I wonder if the return type for `consumeComment` should also be changed as part of this PR - ?

Are you thinking of improving the type annotation, or are you thinking it of changing it from `Comment|false` to `?Comment`?

oliverklee:
Anyway, I have just created #1176 to use `null` as the "nothing" type instead.

JakeQZ:
> Are you thinking of improving the type annotation, or are you thinking it of changing it from `Comment|false` to `?Comment`?

I added a review comment to the code, but it's disappeared (perhaps I forgot to click the 'Add review comment' button).

I was thinking it could be change to `?Comment`, with `null` returned instead of `false` if there isn't a comment to consume.

JakeQZ (approved):
Since #1176 has been added to capture the outstanding issue, this can be merged.

PR #1174, opened 2025-03-16
[BUGFIX] Fix the return type of `Selector::isValid()`

### LINKED ISSUE #1043: Have `Selector::isValid()` actually return a `bool`
Currently, is returns `int<0, max>|false`.

We also need to add test coverage for this method and add a native return type declaration.

oliverklee:
Fixes #1043
Part of #811

### DIFF
--- a/src/Property/Selector.php
+++ b/src/Property/Selector.php
@@ -37,13 +37,13 @@ class Selector implements Renderable
     private $selector;
 
     /**
-     * @return bool
-     *
      * @internal since V8.8.0
      */
-    public static function isValid(string $selector)
+    public static function isValid(string $selector): bool
     {
-        return \preg_match(static::SELECTOR_VALIDATION_RX, $selector);
+        $numberOfMatches = \preg_match(self::SELECTOR_VALIDATION_RX, $selector);
+
+        return $numberOfMatches === 1;
     }
 
     public function __construct(string $selector)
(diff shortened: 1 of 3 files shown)


oliverklee:
Blocked by #1175.

JakeQZ (approved):

src/Property/Selector.php
```diff
@@ -37,13 +37,13 @@ class Selector implements Renderable
...
     /**
-     * @return bool
-     *
      * @internal since V8.8.0
      */
-    public static function isValid(string $selector)
+    public static function isValid(string $selector): bool
     {
-        return \preg_match(static::SELECTOR_VALIDATION_RX, $selector);
+        $numberOfMatches = \preg_match(self::SELECTOR_VALIDATION_RX, $selector);
```
JakeQZ, line 44: This needs to be `static` because `KeyframeSelector` overrides the `SELECTOR_VALIDATION_RX` constant to also allow keyframe animation percentage.

JakeQZ (changes requested):
The failing test is because this breaks keyframes parsing, and is not related to the vendor prefix.  It just so happens that the keyframes test was written with a vendor prefix, and no equivalent test was written without a vendor prefix.

src/Property/Selector.php
oliverklee: Ah, that's why. Fix coming up …

JakeQZ (approved):

PR #1175, opened 2025-03-16
[TASK] Drop support for vendor-prefixed keyframe selectors

oliverklee:


JakeQZ (changes requested):
It just so happens that the keyframes test was written with a vendor prefix, and no equivalent test was written without a vendor prefix.  We shouldn't remove this test (but could change it not to use vendor prefixes, or, perhaps better, add an equivalent test which does not use vendor prefixes).  We don't need to remove support for vendor prefixes here, as it is not actually costing us anything.  See https://github.com/MyIntervals/PHP-CSS-Parser/pull/1174#pullrequestreview-2688763247.

oliverklee:
Okay, closing this then.

PR #1181, opened 2025-03-16
[TASK] Add native type declarations for `CSSList`

oliverklee:
Part of #811

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -41,7 +41,7 @@ abstract class CSSList implements Renderable, Commentable
     protected $comments = [];
 
     /**
-     * @var array<int, RuleSet|CSSList|Import|Charset>
+     * @var array<int<0, max>, RuleSet|CSSList|Import|Charset>
      *
      * @internal since 8.8.0
      */
@@ -283,11 +283,9 @@ abstract class CSSList implements Renderable, Commentable
     /**
      * Splices the list of contents.
      *
-     * @param int $offset
-     * @param int $length
      * @param array<int, RuleSet|CSSList|Import|Charset> $replacement
      */
-    public function splice($offset, $length = null, $replacement = null): void
+    public function splice(int $offset, ?int $length = null, ?array $replacement = null): void
     {
         \array_splice($this->contents, $offset, $length, $replacement);
     }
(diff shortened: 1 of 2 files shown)


src/CSSList/CSSList.php
```diff
@@ -41,7 +41,7 @@ abstract class CSSList implements Renderable, Commentable
     protected $comments = [];
 
     /**
-     * @var array<int, RuleSet|CSSList|Import|Charset>
+     * @var array<int<0, max>, RuleSet|CSSList|Import|Charset>
```
JakeQZ, line 44: Could this be a `list`?

src/CSSList/CSSList.php
```diff
@@ -364,10 +363,10 @@ public function setContents(array $contents): void
     /**
      * Removes a declaration block from the CSS list if it matches all given selectors.
      *
-     * @param DeclarationBlock|array<array-key, Selector>|string $selectors the selectors to match
+     * @param DeclarationBlock|array<Selector>|string $selectors the selectors to match
```
JakeQZ, line 366: I guess `array-key` is redundant so can be removed, but also wonder if the middle type could be a `list`...

src/CSSList/CSSList.php
```diff
@@ -444,17 +443,15 @@ protected function renderListContents(OutputFormat $outputFormat): string
...
-     * @return bool
      */
-    abstract public function isRootList();
+    abstract public function isRootList(): bool;
 
     /**
      * Returns the stored items.
      *
-     * @return array<int, RuleSet|Import|Charset|CSSList>
+     * @return array<int<0, max>, RuleSet|Import|Charset|CSSList>
```
JakeQZ, line 452: Can it be a `list`?

JakeQZ (commented):
I'm wondering of some of the array<int, Type> DocBlock types can actually be list<Type> - or is there some usage preventing that?

src/CSSList/CSSList.php
oliverklee: Unfortunately not: [In `remove()`, we `unset()` elements in the array](https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/src/CSSList/CSSList.php#L324), which creates gaps in the keys.

src/CSSList/CSSList.php
oliverklee: No, as `$this->contents` may have gaps in the keys due to the `unset()` in `::remove()`.

JakeQZ (approved):

PR #1183, opened 2025-03-17
[TASK] Add native type declarations for `CSSBlockList`

oliverklee:
Also add some more type checks to ensure that the corresponding types are actually returned.

Part of #811


src/CSSList/CSSBlockList.php
```diff
@@ -103,18 +101,17 @@ protected function allValues(
                 }
             }
         } else {
-            // Non-List `Value` or `CSSString` (CSS identifier)
-            $result[] = $element;
+            if ($element instanceof Value) {
```
JakeQZ, line 104: Could this just be an `elseif`?

src/CSSList/CSSBlockList.php
```diff
@@ -149,7 +146,7 @@ protected function allSelectors(array &$result, $specificitySearch = null): void
                             $comparatorMatched = $selectorSpecificity === $targetSpecificity;
                             break;
                     }
-                    if ($comparatorMatched) {
+                    if ($comparatorMatched && $selector instanceof Selector) {
```
JakeQZ, line 149: If we did want to do the `instanceof` check, we should do it before calling `getSpecificity()`.  However, I think it's unnecessary because `DeclarationBlock::getSelectors()` always returns an array of `Selector`s - and the types in the PHPDoc for `DeclarationBlock::$selectors` are incorrect.  If we first fix the type hints for that property and its getter, I expect we'd find the `instanceof` test is not needed and PHPStan would be happy without it.

JakeQZ (changes requested):
I think it might be best to first fix the type hints for `DeclarationBlock::$selectors` and `DeclarationBlock::getSelectors()`.

src/CSSList/CSSBlockList.php
oliverklee: I took a quick peek into that - that's quite a can of worms, involving covering `setSelectors` with tests, refactoring it to no longer temporarily storing strings in `DeclarationBlock::selectors`, and more. I'll have a look look at that later and create a ticket.

JakeQZ: Ah, yes, I see: during execution of `setSelectors()`, the elements in `$selectors` may temporarily become strings.  And if an exception is thrown, they'll be left in that state.  So it's not as straightforward as I thought.  When you create the ticket, there should be a note to remove the `instanceof` check added here.  For now, it can be added, but needs to be before the `getSpecificity` method call.

JakeQZ (approved):

PR #1187, opened 2025-03-18
[TASK] Add native type declarations for `CSSNamespace`

oliverklee:
Part of #811


src/Property/CSSNamespace.php
```diff
@@ -35,11 +37,10 @@ class CSSNamespace implements AtRule
...
-     * @param string $url
-     * @param string|null $prefix
+     * @param CSSString|URL $url
      * @param int<0, max> $lineNumber
      */
-    public function __construct($url, $prefix = null, int $lineNumber = 0)
+    public function __construct($url, ?string $prefix = null, int $lineNumber = 0)
     {
         $this->url = $url;
         $this->prefix = $prefix;
```
JakeQZ, line 46: The PHPDoc type annotation for the property should probably also be changed to `string|null` while we're here.

src/Property/CSSNamespace.php
```diff
@@ -101,12 +96,12 @@ public function atRuleName(): string
...
 
     /**
-     * @return array<int, string>
+     * @return array{0: CSSString|URL|string, 1?: CSSString|URL}
      */
     public function atRuleArgs(): array
     {
         $result = [$this->url];
-        if ($this->prefix) {
+        if ($this->prefix !== '') {
```
JakeQZ, line 104: This needs to also check it isn't `null`.  (I suspect this method isn't being fully tested.)

JakeQZ (changes requested):
The type for `$url` is now correct.  But the test in `atRuleArgs()` also needs to check `$prefix` isn't `null`, and the DocBlock for that property should probably be corrected as part of this PR.

src/Property/CSSNamespace.php
oliverklee: Correct - this method is neither tested nor used.

JakeQZ (approved):

PR #1188, opened 2025-03-18
[TASK] Reduce and finetune the scope of `@covers` annotations

oliverklee:
The legacy tests are not very focused. Until we have split them up, try to avoid false positives for code coverage.

Also add `@covers` annotations for the parent classes of the tested classes.

### DIFF
--- a/tests/Comment/CommentTest.php
+++ b/tests/Comment/CommentTest.php
@@ -9,9 +9,7 @@ use Sabberworm\CSS\OutputFormat;
 use Sabberworm\CSS\Tests\ParserTest as TestsParserTest;
 
 /**
- * @covers \Sabberworm\CSS\Comment\Comment
- * @covers \Sabberworm\CSS\OutputFormat
- * @covers \Sabberworm\CSS\OutputFormatter
+ * @coversNothing
  */
 final class CommentTest extends TestCase
 {
--- a/tests/OutputFormatTest.php
+++ b/tests/OutputFormatTest.php
@@ -11,7 +11,7 @@ use Sabberworm\CSS\Parser;
 use Sabberworm\CSS\Parsing\OutputException;
 
 /**
- * @covers \Sabberworm\CSS\OutputFormat
+ * @coversNothing
  */
 final class OutputFormatTest extends TestCase
 {
--- a/tests/RuleSet/LenientParsingTest.php
+++ b/tests/RuleSet/LenientParsingTest.php
@@ -11,16 +11,7 @@ use Sabberworm\CSS\Parsing\UnexpectedTokenException;
 use Sabberworm\CSS\Settings;
 
 /**
- * @covers \Sabberworm\CSS\CSSList\Document
- * @covers \Sabberworm\CSS\Parser
- * @covers \Sabberworm\CSS\RuleSet\DeclarationBlock
- * @covers \Sabberworm\CSS\Rule\Rule
- * @covers \Sabberworm\CSS\Value\CSSString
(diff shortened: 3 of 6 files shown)


tests/Comment/CommentTest.php
```diff
@@ -9,9 +9,7 @@
 use Sabberworm\CSS\Tests\ParserTest as TestsParserTest;
 
 /**
- * @covers \Sabberworm\CSS\Comment\Comment
- * @covers \Sabberworm\CSS\OutputFormat
- * @covers \Sabberworm\CSS\OutputFormatter
+ * @coversNothing
```
JakeQZ, line 12: Doesn't this cover `Comment::render()` and `ParserState::consumeWhitespace()`?

tests/OutputFormatTest.php
```diff
@@ -11,7 +11,7 @@
 use Sabberworm\CSS\Parsing\OutputException;
 
 /**
- * @covers \Sabberworm\CSS\OutputFormat
+ * @coversNothing
```
JakeQZ, line 14: Isn't this covering various `OutputFormat` methods (`create*` and some of the setters)?

tests/RuleSet/LenientParsingTest.php
```diff
@@ -11,16 +11,7 @@
...
- * @covers \Sabberworm\CSS\Parser
- * @covers \Sabberworm\CSS\RuleSet\DeclarationBlock
- * @covers \Sabberworm\CSS\Rule\Rule
- * @covers \Sabberworm\CSS\Value\CSSString
- * @covers \Sabberworm\CSS\Value\CalcFunction
- * @covers \Sabberworm\CSS\Value\Color
- * @covers \Sabberworm\CSS\Value\LineName
- * @covers \Sabberworm\CSS\Value\Size
- * @covers \Sabberworm\CSS\Value\URL
+ * @coversNothing
```
JakeQZ, line 14: Agree it's generally not clear which of the classes this covers - but doesn't it at least cover `Settings` to some extent?

JakeQZ (commented):
Some of the `TestCase`s do seem to at least partially cover some specific classes...

tests/Comment/CommentTest.php
oliverklee: `Comment::render()` is called. As far as I understand the tests (which is not far), they're about other elements rendering comments (or not) depending on the `OutputFormat` settings.

oliverklee:
I'd like to reduce what the legacy tests cover in general, and re-add the `@covers` annotation when we move/rework the tests, and when we're really sure the behavior of which classes is actually tested (instead of only executed).

tests/OutputFormatTest.php
oliverklee: As far as I understand this test, it's not about covering how `OutputFormat` works, but about how other elements get rendered depending on the output format.

tests/RuleSet/LenientParsingTest.php
oliverklee: It _calls_ code in `Settings` (and uses it). It does not test whether `Settings` actually work correctly, though, and the point of the tests is not testing `Settings` (as far as I understand them), but how other code works depending on the settings.

tests/Comment/CommentTest.php
JakeQZ: Yes, it's `outputFormatter::comments()` where the controlling logic is.

JakeQZ (approved):

PR #1191, opened 2025-03-18
[TASK] Deprecate `::atRuleArgs()`

oliverklee:
These methods

- are not used internally
- are redundant to the existing getters
- have wildly different return types across classes, which reduces the usefulness of the general method

### DIFF
--- a/src/Property/Charset.php
+++ b/src/Property/Charset.php
@@ -99,6 +99,8 @@ class Charset implements AtRule, Positionable
 
     /**
      * @return string
+     *
+     * @deprecated since V8.9.0, will be removed in version 9.0.0. Use the specific getter instead.
      */
     public function atRuleArgs()
     {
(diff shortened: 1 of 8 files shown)


src/Property/Charset.php
```diff
@@ -103,6 +103,8 @@ public function atRuleName()
 
     /**
      * @return string
+     *
+     * @deprecated since V8.8.0, will be removed in version 9.0.0. Use the specific getter instead.
```
JakeQZ, line 107: This doesn't have a getter for the `CSSString` object property, only one that returns a native `string`.  I wonder if we should first add a getter for the `CSSString`.

src/RuleSet/AtRuleSet.php
```diff
@@ -45,6 +45,8 @@ public function atRuleName()
 
     /**
      * @return string
+     *
+     * @deprecated since V8.8.0, will be removed in version 9.0.0. Use the specific getter instead.
```
JakeQZ, line 49: This doesn't have a specific getter for the `$arguments` property, so we should add one first.

JakeQZ (commented):
Some of the classes don't have equivalent alternative getters, so we should add these first.

I've made a suggestion for the deprecation note for the `interface`.

Issue #1170, opened 2025-03-15
nested selectors break rendering


ocoste:
When I parse a CSS with nested selector, the rendered CSS bugs

Example : 

$css = "
.presentations_frameEditorItem {
    position: relative;
    a {
        border: 1px solid var(--theme-primary-border);
    }

    &.presentations_isDragging {
        opacity: 0
    }
}
";

$parser = (new Parser($css))->parse();

$css = $parser->render();

--------------------------------------------------------
.presentations_frameEditorItem {
position: relative;
border: 1px solid var(--theme-primary-border);
}

How can we prevent this?

Thanks

oliverklee:
Thanks for reporting this!

This seems to be a missing feature. I've tagged and renamed this issue accordingly.

JakeQZ:
This shouldn't be too tricky to implement.  These are my initial thoughts:

- `RuleSet::$rules` elements would need to be allowed to be either a `Rule` or a `DeclarationBlock`:
  - A (possibly-empty) interface `RuleOrDeclarationBlock` (for want of a better name) can be introduced for type-strictness;
  - This may allow construction of invalid constructs, such as a declaration block inside an at-rule where it is not permitted, but it would be caveat emptor - invalid constructs would be rejected at the parsing phase, and only possible via manipulation;
- `RuleSet::parseRuleSet()` would need an extra Boolean parameter to permit declaration blocks as well as property-value rules:
  - Only `DeclarationBlock::parse()` would pass this as `true`;
  - Determining whether the next item is a rule or a nested declaration block is slightly tricky:
    - The next characters of interest that would determine which it is are `{` (for a declaration block) and `;` or `}` for a rule;
    - But these could be contained within a string in either situation (e.g. `content` property for a rule, or attribute selector), so proper parsing is required;
    - We could attempt to initially parse as a rule, but `a:hover` would pass unless we then checked for a `{` coming later;
    - Or we could attempt to initially parse as a declaration block, which should always fail if there's no `{`;
    - I think the latter seems more reliable;
    - Either way, I think it would involve catching an exception during normal parsing operation, which I'm not keen on, because exceptions should only be used for exceptional situations, but I can't think of a way of easily avoiding that.
- Rendering should take care of itself and need no changes.

@oliverklee, @sabberworm, WDYT?

JakeQZ:
[...]

PR #1194, opened 2025-03-19
[TASK] Use delegation for `DeclarationBlock` -> `RuleSet`

JakeQZ:
... rather than inheritance.

This will allow `DeclarationBlock` to instead extend `CSSBlockList` in order to support
[CSS nesting](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_nesting).

This is a slightly-breaking change, since now `CSSBlockList::getAllRuleSets()` will include the `RuleSet` property of the `DeclarationBlock` instead of the `DeclarationBlock` itself.

Part of #1170.


src/CSSList/CSSList.php
```diff
@@ -41,7 +41,7 @@ abstract class CSSList implements Renderable, Commentable
     protected $comments = [];
 
     /**
-     * @var array<int<0, max>, RuleSet|CSSList|Import|Charset>
+     * @var array<int<0, max>, RuleSet|CSSList|DeclarationBlock|Import|Charset>
```
oliverklee, line 44: I propose we introduce an interface instead, e.g. something like `CSSListItem`. (We could do this as a pre-patch to keep PRs small.)

src/RuleSet/DeclarationBlock.php
```diff
@@ -13,22 +15,48 @@
...
 
+    /**
+     * @var RuleSet
+     */
+    private $ruleSet;
+
+    /**
+     * @var int<0, max>
+     */
+    private $lineNumber;
```
oliverklee, line 44: … and maybe we can introduce the traits for `Commentable` and for things with a line number first.

src/CSSList/CSSList.php
JakeQZ: Doing so would also make Stan happy :)

JakeQZ:
Would be good to resolve #1247 before moving on with this.

JakeQZ:
This is almost ready for review, but some tests need to be added.  Before doing so, I'd appreciate a pre-review from you: @sabberworm, @oliverklee.

I'm not keen on the having the various chaining methods in `DeclarationBlock`, but to avoid breaking changes they are necessary.  A `trait` can be used for testing them in the same way as their counterparts in `RuleSet`.

For `DeclarationBlock` to contain nested rules, it needs to extend `CSSBlockList`.  `RuleSet` cannot be a `trait`; it needs to be a fully-fledged type for methods like `getAllRuleSets()`.

This is the only realistic way forward I see for now, in an 'Agile' sense, which can deliver what is required now, without wholescale refactoring or breaking changes.

config/phpstan-baseline.neon
```diff
@@ -54,6 +54,12 @@ parameters:
 			count: 1
 			path: ../src/RuleSet/DeclarationBlock.php
 
+		-
+			message: '#^Parameters should have "string\|null" types as the only types passed to this method$#'
+			identifier: typePerfect.narrowPublicClassMethodParamType
+			count: 2
+			path: ../src/RuleSet/DeclarationBlock.php
+
```
JakeQZ, line 62: @oliverklee,

These warnings are unfathomable and look like bugs in the tool.  If I remove the call to `getRulesAssoc()` from https://github.com/MyIntervals/PHP-CSS-Parser/blob/1ce18b7810617b9795feb9276d5848d3d9896931/tests/ParserTest.php#L382 one of the warnings disappears.

But `getRulesAssoc()` is defined as accepting only `string` or `null`, and the caller there is passing a `string` constant.  Am I missing something?

oliverklee: This warning comes from the PHPStan extension https://github.com/rectorphp/type-perfect, and this might (or might not) be a bug in that extension. I've just checked with their latest release 2.1.0 (which was released 4 days ago), but the warning still persists. I'll have a look into our code now.

oliverklee: Yes, this looks like a bug in the tool to me. Somehow the tool gets hung up on the parameter being nullable.

oliverklee: Ah, or this might also be just the way this rule works (or is intended to work):
https://github.com/rectorphp/type-perfect?tab=readme-ov-file#4-narrow-param-types

oliverklee: If I understand the rule correctly, it is telling us that the parameter doesn't need to be nullable.

src/CSSList/CSSBlockList.php
```diff
@@ -95,7 +97,7 @@ public function getAllValues(
                     );
                 }
             }
-        } elseif ($element instanceof RuleSet) {
+        } elseif ($element instanceof RuleSet || $element instanceof DeclarationBlock) {
```
oliverklee, line 100: Maybe we can have an interface that both classes implement so we can check for the interface here?

oliverklee (commented):
I like this approach!

config/phpstan-baseline.neon
JakeQZ: Yes, it seems that the message is wrong and should be 'Parameters should have "string" types as the only types passed to this method'.  If I add an additional call to `getRulesAssoc(null)`, this resolves the warning.  But if I call `getRulesAssoc()` without an argument, the warning is still given, despite the default value of the parameter being `null`, so that does look like a bug in the tool.  Perhaps when tests are added, the warnings will disappear.

src/CSSList/CSSBlockList.php
JakeQZ: Done.  Interestingly, this resolves the PHPStan warning (because the parameter type is now required to match the interface).

oliverklee:
Maybe as another spin-off PR (or prepatch), we can introduce the interface and have one class implement it.

JakeQZ:
> Maybe as another spin-off PR (or prepatch), we can introduce the interface and have one class implement it.

#1256.

src/RuleSet/DeclarationBlock.php
```diff
@@ -30,7 +30,7 @@
  *
  * Note that `CSSListItem` extends both `Commentable` and `Renderable`, so those interfaces must also be implemented.
  */
-class DeclarationBlock implements CSSElement, CSSListItem, Positionable
+class DeclarationBlock implements CSSElement, CSSListItem, Positionable, RuleContainer
```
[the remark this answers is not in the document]
JakeQZ: Tests have been added to assert all interfaces are implemented.

JakeQZ:
I think all pre-PRs that can be separated have been completed now.

src/RuleSet/RuleSet.php
```diff
@@ -24,10 +24,9 @@
  * If you want to manipulate a `RuleSet`, use the methods `addRule(Rule $rule)`, `getRules()` and `removeRule($rule)`
  * (which accepts either a `Rule` or a rule name; optionally suffixed by a dash to remove all related rules).
  *
- * Note that `CSSListItem` extends both `Commentable` and `Renderable`,
- * so those interfaces must also be implemented by concrete subclasses.
+ * Note that `CSSListItem` extends both `Commentable` and `Renderable`, so those interfaces must also be implemented.
  */
-abstract class RuleSet implements CSSElement, CSSListItem, Positionable, RuleContainer
+class RuleSet implements CSSElement, CSSListItem, Positionable, RuleContainer
```
oliverklee: I think we can move the changes to `RuleSet` (plus test coverage) to a pre-patch, too.

src/RuleSet/DeclarationBlock.php
```diff
@@ -21,14 +27,33 @@
...
+     */
+    private $ruleSet;
+
+    /**
+     * @param int<1, max>|null $lineNumber
+     */
+    public function __construct(?int $lineNumber = null)
+    {
+        $this->setPosition($lineNumber);
+        $this->ruleSet = new RuleSet($lineNumber);
```
oliverklee, line 53: Nitpick: I usually prefer initializing the dependencies before setting properties in the constructor. That's a personal preference, though, so feel free to ignore it if you prefer the other order.

tests/ParserTest.php
```diff
@@ -1079,7 +1079,7 @@ public function commentExtracting(): void
         self::assertSame('* Number 10 *', $fooBarComments[0]->getComment());
 
         // Media -> declaration -> rule.
```
oliverklee, line 1078: Is this comment still up to date? (Haven't checked, though.)

tests/Unit/RuleSet/DeclarationBlockTest.php
```diff
@@ -5,14 +5,20 @@
...
 use Sabberworm\CSS\CSSList\CSSListItem;
+use Sabberworm\CSS\Position\Positionable;
+use Sabberworm\CSS\Rule\Rule;
 use Sabberworm\CSS\RuleSet\DeclarationBlock;
+use Sabberworm\CSS\RuleSet\RuleSet;
 
 /**
  * @covers \Sabberworm\CSS\RuleSet\DeclarationBlock
  */
 final class DeclarationBlockTest extends TestCase
```
oliverklee, line 22: Are any of the changes to this testcase already possible now, i.e., could we move them to a pre-patch? (Haven't checked in detail yet, though.)

oliverklee (changes requested):

src/RuleSet/DeclarationBlock.php
JakeQZ: The dependency (`Position`) _is_ initialized before the property (`$ruleSet`) here.

Hadn't really thought about it, but from C++ we always initialize the superclass first, which is partially enforced as a language construct.

tests/ParserTest.php
JakeQZ: Actually it seems to fit better following this change.  Though `RuleSet` and `DeclarationBlock` as class names are misnomers.

tests/Unit/RuleSet/DeclarationBlockTest.php
JakeQZ: The `assertInstanceOf` tests can be added separately - #1295 for this.

src/RuleSet/DeclarationBlock.php
oliverklee: Independent of personal preferences, I'd like use to use some terms in compatible ways in this project. :-)

I know the term "dependency" (in the context classes in software architecture) to mean "one class uses/extends/implements another class"). That's also what dependency injection takes care of.

What's your understanding of the term, @JakeQZ ?

JakeQZ: That's my understanding too, I think.  A dependency is another class, interface or trait that is required by the class in hand, either through inheritance, implementing an interface, using a trait, or as an object property.

I think I somewhat misunderstood your initial comment.  I took the view that `setPosition()` was initializing the `Position` dependency (since traits don't have their own constructor), and that assigning the `$ruleSet` property was setting a property, and thus it was in the preferred order.

But I can see that `Position` is already initialized to `null` defaults, and directly assigning a property is more initial than indirectly doing so via a setter.  So am happy to switch the order :)

JakeQZ: > So am happy to switch the order :)

Now done, and rebased.

oliverklee:
This PR needs a rebase.

Apart from that, I think we can move de-abstracting the `RuleSet` class (and creating a dedicated testcase for it by copying some existing tests) into a pre-PR.

JakeQZ:
> This PR needs a rebase.

It is impossible to rebase now, because git does it one commit at a time, and I probably included changes in the wrong commits in a previous rebase.  I tried merge.  Might be able to fix the resulting errors.  Otherwise will have to start over.

tests/Unit/RuleSet/DeclarationBlockTest.php
```diff
@@ -53,112 +54,52 @@ public function implementsPositionable(): void
...
-            'type & id' => ['h2#my-mug'],
-            'pseudo-class' => [':hover'],
-            'type & pseudo-class' => ['a:hover'],
-            '`not`' => [':not(#your-mug)'],
-            '`not` with multiple arguments' => [':not(#your-mug, .their-mug)'],
-            'pseudo-element' => ['::before'],
-            'attribute with `"`' => ['[alt="{}()[]\\"\',"]'],
-            'attribute with `\'`' => ['[alt=\'{}()[]"\\\',\']'],
-        ];
-    }
```
JakeQZ: What happened to this?

JakeQZ (commented):
This needs starting over.  It's a mess now.

JakeQZ:
I tried undoing lastest merge, but it's still a mess.  I can't even get back to where I was now.  Will take some time redoing.

oliverklee:
> I tried undoing lastest merge, but it's still a mess. I can't even get back to where I was now. Will take some time redoing.

In my experience, when I had a PR from which I spun off some pre-PRs, I also sometimes had nasty merge conflicts when rebasing. In those cases, it made things a _lot_ easier if I squashed the existing commits in the PR branch before rebasing.

JakeQZ:
> > I tried undoing lastest merge, but it's still a mess. I can't even get back to where I was now. Will take some time redoing.
> 
> In my experience, when I had a PR from which I spun off some pre-PRs, I also sometimes had nasty merge conflicts when rebasing. In those cases, it made things a _lot_ easier if I squashed the existing commits in the PR branch before rebasing.

I think I have somehow managed to rebase without including any changes from `main`.  So now the diffs are showing that all the changes from `main` are being reverted.  I did `git rebase --abort` when I got in a muddle.

If I could somehow reset the branch to the state it was in immediately after the commit "Reorder construction", then squashing may be possible, but I'm not sure how to do that.

oliverklee:
> If I could somehow reset the branch to the state it was in immediately after the commit "Reorder construction", then squashing may be possible, but I'm not sure how to do that.

[`git reflog`](https://git-scm.com/docs/git-reflog) allows you to list all recent commits and which actions lead to them, including rebases, branch switches etc.  So this might be helpful. (It has saved my work a couple of times.)

JakeQZ:
> If I could somehow reset the branch to the state it was in immediately after the commit "Reorder construction"

```cmd
git reset -hard {SHA-of-last-commit-before-attempted-rebase}
```

... undid the failed rebase/merge.  I think when I tried merging instead (via GitHub), that moved the 'common ancestor' point, but when I reverted the merge, it was not moved back.

> then squashing may be possible, but I'm not sure how to do that.

```cmd
git reset -soft {SHA-of-first-branch-commit}
git commit --amend
```
... did the squashing.

> In my experience, when I had a PR from which I spun off some pre-PRs, I also sometimes had nasty merge conflicts when rebasing. In those cases, it made things a _lot_ easier if I squashed the existing commits in the PR branch before rebasing.

That helped immensely; thanks.  Instead of (presumably) 9 rounds of conflict resolution, including for files with no overall conflict, there was just one round, for the two files that actually had conflicts.  (I'm used to Perforce, which does the merge all at once, and also has a visual editor for 3-way merging.  Keeping a cool head also helps ;))

Rebase now done !)  Thanks <3

oliverklee:
> I'm used to Perforce, which does the merge all at once, and also has a visual editor for 3-way merging. 

I use PhpStorm as my IDE, and IIRC it also have a nice 3-way-merge UI. (Disclaimer: I don't use UIs for git stuff and hence can only report what I hear and see from other people.)

JakeQZ:
I think this is now ready for a final review.

src/RuleSet/DeclarationBlock.php
```diff
@@ -107,7 +132,9 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
             }
         }
         $result->setComments($comments);
-        RuleSet::parseRuleSet($parserState, $result);
+
+        RuleSet::parseRuleSet($parserState, $result->ruleSet);
```
oliverklee: Even though this technically works, I'd find using the getter `getRuleSet()` more clean here.

src/RuleSet/DeclarationBlock.php
```diff
@@ -21,14 +27,33 @@
...
+     * @var RuleSet
+     */
+    private $ruleSet;
+
+    /**
+     * @param int<1, max>|null $lineNumber
+     */
+    public function __construct(?int $lineNumber = null)
+    {
+        $this->ruleSet = new RuleSet($lineNumber);
```
oliverklee, line 53: We also need unit tests that the created `RuleSet` uses has line number from the `DeclarationBlock` constructor, and that it defaults to `null`.

src/RuleSet/DeclarationBlock.php
```diff
@@ -175,6 +202,73 @@ public function getSelectors(): array
         return $this->selectors;
     }
 
+    public function getRuleSet(): RuleSet
```
oliverklee, line 205: We should have unit tests for the added methods.

oliverklee (changes requested):

src/RuleSet/DeclarationBlock.php
JakeQZ: `parseRuleSet()` takes the `RuleSet` argument by reference, in order to fill it in with the the parsed `Rule`s, so the class property needs to be passed.

src/RuleSet/DeclarationBlock.php
JakeQZ: `getRuleSet()` has some tests added for it.  The methods that chain on to the equivalent `RuleSet` method have tests added via the `RuleContainerTest` trait (which is also used in the test case for `RuleSet`).  So I think all the added methods have test coverage.

src/RuleSet/DeclarationBlock.php
JakeQZ: We don't appear to have an existing test that `RuleSet` applies the line number provided, so perhaps another pre-PR is needed...

JakeQZ: #1348 for `RuleSet`.  Once that is approved, I can address this along similar lines.

src/RuleSet/DeclarationBlock.php
oliverklee: As far as I understand it, `getRuleSet()` returns the instance, not a clone of it.

(PHP by default passes object kind-of-by-reference. I explained this on page 17 of my [diploma thesis](https://github.com/oliverklee/pixy-thesis/blob/main/pixy-thesis.pdf) back then.)

src/RuleSet/DeclarationBlock.php
JakeQZ: Have added tests to cover:
1. `getLineNumber()` returning line number passed to `DeclarationBlock` constructor (or `null` by default);
2. `getRuleSet()->getLineNumber()` returning line number passed to `DeclarationBlock` constructor (or `null` by default);

But perhaps `1` could be covered by another pre-PR...

src/RuleSet/DeclarationBlock.php
JakeQZ: You're right.  I was thinking of arrays (even though I knew it was an object - my thinking cap must have been on back-to-front).  Thanks for the thesis link; I will give it a read some time...

src/RuleSet/DeclarationBlock.php
JakeQZ: #1350 for item 1.

JakeQZ:
This issues from the last review should all now be resolved.

PR #1197, opened 2025-03-19
[TASK] Add native type declarations for `CSSFunction`

oliverklee:
Also improve the related type annotations and declarations in other classes in order to keep things consistent and to keep Rector from changing things.

### DIFF
--- a/src/Value/CSSFunction.php
+++ b/src/Value/CSSFunction.php
@@ -17,19 +17,18 @@ use Sabberworm\CSS\Parsing\UnexpectedTokenException;
 class CSSFunction extends ValueList
 {
     /**
-     * @var string
+     * @var non-empty-string
      *
      * @internal since 8.8.0
      */
     protected $name;
 
     /**
-     * @param string $name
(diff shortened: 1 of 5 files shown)


src/Value/CSSFunction.php
```diff
@@ -17,19 +17,18 @@
...
+     * @var non-empty-string
      *
      * @internal since 8.8.0
      */
     protected $name;
 
     /**
-     * @param string $name
-     * @param RuleValueList|array<array-key, Value|string> $arguments
-     * @param string $separator
```
JakeQZ, line 29: Can this be a `non-empty-string` as well?

JakeQZ (commented):
Can `$separator` be a `non-empty-string` or does this break something?

src/Value/CSSFunction.php
oliverklee: It can indeed, and should be. Changed.

JakeQZ (approved):

PR #1198, opened 2025-03-19
[CLEANUP] Improve type annotations in `LineName`

oliverklee:


src/Value/LineName.php
```diff
@@ -12,7 +12,7 @@
 class LineName extends ValueList
 {
     /**
-     * @param array<int, Value|string> $components
+     * @param array<Value|string> $components
```
JakeQZ, line 15: Won't this weaken the type specification, also allowing strings as array keys?

JakeQZ (commented):
I have a query about `array<Type>`.  I had assumed that they keys would default to `array-key`, but can't find anything in the PHPStan docs.

src/Value/LineName.php
oliverklee: Yes, it does. As far as I can see, the array keys are not used, and hence it doesn't matter if string or int array keys are passed in.

JakeQZ: OK.

JakeQZ (approved):

PR #1199, opened 2025-03-19
[CLEANUP] And some more annotations for non-empty strings

oliverklee:


src/Value/CSSString.php
```diff
@@ -83,6 +83,9 @@ public function getString(): string
         return $this->string;
     }
 
+    /**
+     * @return non-empty-string
+     */
     public function render(OutputFormat $outputFormat): string
```
JakeQZ, line 89: We can probably also add non-empty for `OutputFormat::$stringQuotingType` and getter/setter.

PR #1200, opened 2025-03-19
[CLEANUP] Make annotations for `OutputFormat` more specific

oliverklee:


src/OutputFormat.php
```diff
@@ -655,6 +663,8 @@ public function indentWithTabs(int $numberOfTabs = 1): self
     }
 
     /**
+     * @param int<1, max> $numberOfSpaces
```
JakeQZ, line 666: `setIndentation('')` can be called directly to have no indentation.

JakeQZ (approved):

PR #1201, opened 2025-03-19
[TASK] Raise PHPStan to level 4

oliverklee:
Also allow `assertInstanceOf` checks in the tests (as we find those useful).


oliverklee:
Let's get this in before the 9.0 release.

PR #1202, opened 2025-03-19
[TASK] Drop the unused `ParserState::strpos()` method

oliverklee:


JakeQZ (approved):
We can always put it back in if we find we need it to resolve a multibyte character issue.

PR #1206, opened 2025-03-21
[TASK] Add trait providing standard implementation of `Commentable`

### LINKED ISSUE #813: Add `trait` for 'standard' `Commentable` implementation
The various classes that implement `Commentable` all seem to have exactly the same implementation.

This duplication could be avoided by using a `trait` for a 'standard' implementation.

JakeQZ:
Resolves #813.

### DIFF
--- a/src/Comment/CommentContainer.php
+++ b/src/Comment/CommentContainer.php
@@ -1,0 +1,42 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Comment;
+
+/**
+ * Provides a standard reusable implementation of `Commentable`.
+ *
+ * @internal
+ */
+trait CommentContainer
+{
+    /**
+     * @var list<Comment>
+     */
+    protected $comments = [];
+
+    /**
+     * @param list<Comment> $comments
+     */
+    public function addComments(array $comments): void
+    {
+        $this->comments = \array_merge($this->comments, $comments);
+    }
+
+    /**
+     * @return list<Comment>
+     */
+    public function getComments(): array
+    {
+        return $this->comments;
+    }
+
+    /**
+     * @param list<Comment> $comments
+     */
+    public function setComments(array $comments): void
+    {
+        $this->comments = $comments;
+    }
+}
--- a/src/Comment/CommentContainer.php
+++ b/src/Comment/CommentContainer.php
@@ -8,6 +8,8 @@ namespace Sabberworm\CSS\Comment;
  * Provides a standard reusable implementation of `Commentable`.
  *
  * @internal
+ *
+ * @phpstan-require-implements Commentable
  */
 trait CommentContainer
 {
--- a/tests/Unit/Comment/CommentContainerTest.php
+++ b/tests/Unit/Comment/CommentContainerTest.php
@@ -56,7 +56,7 @@ final class CommentContainerTest extends TestCase
      *
      * @dataProvider provideCommentArray
      */
-    public function getCommentsAfterCommentsAddedToVirginContainerReturnsThoseComments(array $comments): void
+    public function addCommentsOnVirginContainerAddsCommentsProvided(array $comments): void
     {
         $this->subject->addComments($comments);
 
@@ -70,7 +70,7 @@ final class CommentContainerTest extends TestCase
      *
      * @dataProvider provideCommentArray
      */
-    public function getCommentsAfterEmptyArrayOfCommentsAddedReturnsOriginalComments(array $comments): void
+    public function addCommentsWithEmptyArrayKeepsOriginalCommentsUnchanged(array $comments): void
     {
         $this->subject->setComments($comments);
 
@@ -148,10 +148,8 @@ final class CommentContainerTest extends TestCase
      *
      * @dataProvider provideTwoDistinctCommentArraysWithSecondNonempty
      */
-    public function getCommentsAfterCommentsAddedIncludesOriginalComments(
-        array $commentsToAdd,
-        array $originalComments
-    ): void {
+    public function addCommentsKeepsOriginalComments(array $commentsToAdd, array $originalComments): void
+    {
         $this->subject->setComments($originalComments);
 
         $this->subject->addComments($commentsToAdd);
@@ -170,10 +168,8 @@ final class CommentContainerTest extends TestCase
      *
      * @dataProvider provideTwoDistinctCommentArraysWithSecondNonempty
      */
-    public function getCommentsAfterCommentsAddedIncludesCommentsAdded(
-        array $originalComments,
-        array $commentsToAdd
-    ): void {
+    public function addCommentsAfterCommentsSetAddsCommentsProvided(array $originalComments, array $commentsToAdd): void
+    {
         $this->subject->setComments($originalComments);
 
         $this->subject->addComments($commentsToAdd);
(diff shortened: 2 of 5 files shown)


src/Comment/CommentContainer.php
```diff
@@ -0,0 +1,42 @@
...
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Comment;
+
+/**
+ * Provides a standard reusable implementation of `Commentable`.
+ *
+ * @internal
+ */
+trait CommentContainer
```
oliverklee, line 12: I haven't had time for a full review yet (customer project, and gave a workshop today) and will do one later.

For the trait, we can add `@phpstan-require-implements Commentable`: https://phpstan.org/writing-php-code/phpdocs-basics#enforcing-implementing-an-interface-for-traits

JakeQZ: Done.

tests/Unit/Comment/CommentContainerTest.php
```diff
@@ -0,0 +1,239 @@
...
+use Sabberworm\CSS\Tests\Unit\Comment\Fixtures\ConcreteCommentContainer;
+use TRegx\DataProvider\DataProviders;
+
+/**
+ * @covers \Sabberworm\CSS\Comment\CommentContainer
+ */
+final class CommentContainerTest extends TestCase
+{
+    /**
+     * @var Commentable
```
oliverklee, line 21:
```suggestion
     * @var ConcreteCommentContainer
```

tests/Unit/Comment/CommentContainerTest.php
```diff
@@ -0,0 +1,239 @@
...
+     * @return array<non-empty-string, array{0: list<Comment>}>
+     */
+    public function provideCommentArray(): array
+    {
+        return [
+            'no comment' => [[]],
+            'one comment' => [[new Comment('Is this really a spoon?')]],
+            'two comments' => [[
+                new Comment('I’m a teapot.'),
+                new Comment('I’m a cafetière.'),
```
oliverklee, line 48: :smile:

tests/Unit/Comment/CommentContainerTest.php
```diff
@@ -0,0 +1,239 @@
...
+    }
+
+    /**
+     * @test
+     *
+     * @param list<Comment> $comments
+     *
+     * @dataProvider provideCommentArray
+     */
+    public function getCommentsAfterEmptyArrayOfCommentsAddedReturnsOriginalComments(array $comments): void
```
oliverklee, line 74: Maybe we can describe the effects instead of focusing on the methods:

```suggestion
    public function setCommentsWithEmptyArrayKeepsOriginalCommentsUnchanged(array $comments): void
```

tests/Unit/Comment/CommentContainerTest.php
```diff
@@ -0,0 +1,239 @@
...
+
+    /**
+     * @test
+     *
+     * @param list<Comment> $originalComments
+     * @param non-empty-list<Comment> $commentsToAdd
+     *
+     * @dataProvider provideTwoDistinctCommentArraysWithSecondNonempty
+     */
+    public function getCommentsAfterCommentsAddedIncludesCommentsAdded(
```
oliverklee, line 174: Love those tests!

src/Comment/CommentContainer.php
```diff
@@ -0,0 +1,44 @@
...
+namespace Sabberworm\CSS\Comment;
+
+/**
+ * Provides a standard reusable implementation of `Commentable`.
+ *
+ * @internal
+ *
+ * @phpstan-require-implements Commentable
+ */
+trait CommentContainer
```
oliverklee, line 14: For traits providing the default implementation of an interface, I like having the two named in a way that makes the relationship obvious - and the naming consistent.

In TYPO3 land, we (try to) use this convention:

`CoffeeInterface` and `CoffeeTrait`

As `Commentable` has no `Interface` suffix, we obviously cannot use this schema without changes.

I'd be perfectly fine with naming the trait `CommetableTrait` (while keeping the interface name without changes), though. What do you think?

oliverklee (changes requested):

tests/Unit/Comment/CommentContainerTest.php
JakeQZ: I didn't want to make the type of the test subject any more specific than it needed to be.  But if you think it should be, I'm happy with that.

tests/Unit/Comment/CommentContainerTest.php
JakeQZ: Your friend @sebastianbergmann said he'd never found a use case for `assertArraySubset()`, which has been [deprecated and removed](https://github.com/sebastianbergmann/phpunit/issues/3494).

@rdohms created an extension package with the removed functionality: https://packagist.org/packages/dms/phpunit-arraysubset-asserts.

But it only ever worked for associative arrays, and that package does not extend the assertion to support indexed arrays (i.e. by only checking the values, not also the keys).  There was an earlier [PR to resolve this](https://github.com/sebastianbergmann/phpunit/pull/2781), but it was rejected.

Lacking an ideal assertion method, I had to implement (small and simple) methods `createContainsContstraint()`, `createContainsContstraints()`, and use `assertThat()` along with `LogicalAnd` in the test method.

Maybe a light topic of conversation when you next meet for board games playing... :)

src/Comment/CommentContainer.php
JakeQZ: I'm really not a fan of having entity types included in their names.

We do not
- name classes like `CoffeeClass` to clarify that it's a class;
- name properties like `$sugarProperty` to clarify that it's a property;
- name variables like `$milkVariable` to clarify that it's a variable.

By extension, we should not
- name interfaces like `CoffeeInterface`:
  - It is clear from `implements Coffee` that an interface is being implemented;
  - `Coffee` might be an interface now, but is also a type, that may be later changed to an abstract (or even concrete) class; if and when that happens, we're stuck with the misnomer `CoffeeInterface`;
- name abstract classes like `AbstractBeverage`:
  - If, later, we decide to make this a concrete class, we are again stuck with a misnomer;
    - In #1194 (albeit embrionic), `RuleSet` is being made a concrete class; if it was originally named `AbstractRuleSet`, and part of the public API (which it is), we would have that naming issue to deal with;
- name traits like `CoffeeTrait`:
  - It should be clear from the `use` declaration that it's a trait.

The internal PHP interfaces, like `Traversable` do not include "Interface" in their names.  The "-able" naming suffix is common in both PHP and C++ to indicate an interface.

> In TYPO3 land, we (try to) use this convention:
> 
> `CoffeeInterface` and `CoffeeTrait`

I would go as far as saying that I recommend you don't, for the reasons given above.

Back to the case in point...

I chose the name `CommentContainer` because that's what it is, in a Wittgensteinian sense.

oliverklee:
There's a leftover `use`, and also one open comment.

tests/Unit/Comment/CommentContainerTest.php
JakeQZ: Apart from `getCommentsInitiallyReturnsEmptyArray()`, all the test methods focus on either `addComments()` (first group) or `setComments()` (last group).  So perhaps they all me renamed like `addCommentsDoesXYZ()` and `setCommentsDoesXYZ` - WDYT?

JakeQZ:
> There's a leftover `use` (which PHP-CS-Fixer is complaining about), and also one open comment.

I've replied to the open comment with a question.  I've fixed the leftover `use` - I'd forgotton to do `git add` before amending the commit.

tests/Unit/Comment/CommentContainerTest.php
oliverklee: Yes, I'd like that!

JakeQZ: Done.

oliverklee (approved):

Issue #1176, opened 2025-03-16
Change `|false` as a fallback type to `|null`


oliverklee:
This will also allow us to have native nullable types more.

PR #1209, opened 2025-03-23
[CLEANUP] Return `null` from `DeclarationBlock::parse()` on failure

JakeQZ:
Also add clarification of meaning of return value from `CSSList::parseListItem()`, where `null` and `false` have different meanings.

Part of #1176.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -106,6 +106,9 @@ abstract class CSSList implements Renderable, Commentable
 
     /**
      * @return AtRuleBlockList|KeyFrame|Charset|CSSNamespace|Import|AtRuleSet|DeclarationBlock|false|null
+     *         If `null` is returned, it means the end of the list has been reached.
+     *         If `false` is returned, it means an invalid item has been encountered,
+     *         but parsing of the next item should proceed.
      *
      * @throws SourceException
      * @throws UnexpectedEOFException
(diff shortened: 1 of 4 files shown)


JakeQZ:
`CSSList::parseList()` seems to distinguish `null` from `false` unexpectedly.  Will need looking into first...

JakeQZ:
This is ready for review now.  (I can't see how to re-request review when a previous review has not been done.)

src/CSSList/CSSList.php
```diff
@@ -148,7 +151,7 @@ private static function parseListItem(ParserState $parserState, CSSList $list)
                 return null;
             }
         } else {
-            return DeclarationBlock::parse($parserState, $list);
+            return DeclarationBlock::parse($parserState, $list) ?? false;
```
oliverklee, line 154: We should cover the two functional changes in this class with (functional) tests.

src/RuleSet/DeclarationBlock.php
```diff
@@ -63,7 +61,7 @@ public static function parse(ParserState $parserState, ?CSSList $list = null)
                 if (!$parserState->comes('}')) {
                     $parserState->consumeUntil('}', false, true);
                 }
-                return false;
+                return null;
```
oliverklee, line 64: We should cover this with a (functional) test.

oliverklee (changes requested):

src/CSSList/CSSList.php
JakeQZ: There is no functional change here.  The return value is still `false` if `DeclarationBlock::parse()` fails.  The code change here is because that now returns `null` on failure rather than `false`, but this method still needs to return `false`.

src/RuleSet/DeclarationBlock.php
JakeQZ: Done.

oliverklee (approved):

PR #1210, opened 2025-03-23
[TASK] Prepare release of version V8.8.0

oliverklee:


JakeQZ (approved):
Looking at the [release checklist](https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/docs/release-checklist.md), there is nothing to do for steps 1 and 3, since the `branch-alias` is already "9.0.x-dev", and there is no milestone set in the Dependabot configuration.

PR #1212, opened 2025-03-24
[TASK] Add (and use) a `CSSListItem` type

JakeQZ:
This allows a single type to be used for the contents of a `CSSList`, instead of a long list of orred types, and helps with static analysis.

Various `assertInstanceOf()` tests are added to the test cases to confirm that the list items are of the type expected.

Some `implements` and `exetends` lists are now alphabetically sorted.


JakeQZ:
Notes:
- I have marked this as blocked by #1209, which is a pre-patch for this;
  - This PR includes that change, so if that PR is accepted, it should be possible to merge this PR without conflict;
- When a type is `A|B`, PHPStan is happy if either `A` or `B` has the method invoked,
  - but if `A` and `B` both implement `C`, and the type is `C`, it requires that `C` defines the method, or an additional type check is performed - hence the need for additional `assertInstanceOf()` checks in the tests;
- I think the new PHPStan error arises because it would now be possible to have `CSSListItem|array` as a native parameter type - but not in PHP 7.2 -
  - It would be nice if PHPStan would allow specification of a target PHP version separately from the version it runs on, so it would not generate errors that can't be fixed for the minimum target PHP version;
    - IIRC, this was not possible with Psalm, but maybe it is possible with PHPStan;
    - Running PHPStan on PHP 7.2 would almost certainly not permit the latest version.

JakeQZ:
Apparently PHP<7.4 does not like implementing an interface that is already implemented.  That was the other note I was going to add - that I kept that redundancy in for clarity and completeness.  Seems we can't :/

JakeQZ:
I've fixed the issues with PHP<7.4, and rebased.  This can be reviewed now.

oliverklee (approved):

PR #1216, opened 2025-03-25
[TASK] Configure the target PHP version for PHPStan

### LINKED ISSUE #1214: Define minimum target PHP version for PHPStan?
From https://github.com/MyIntervals/PHP-CSS-Parser/pull/1212#issuecomment-2746719461

> * I think the new PHPStan error arises because it would now be possible to have `CSSListItem|array` as a native parameter type - but not in PHP 7.2 -
>   * It would be nice if PHPStan would allow specification of a target PHP version separately from the version it runs on, so it would not generate errors that can't be fixed for the minimum target PHP version;
>     * IIRC, this was not possible with Psalm, but maybe it is possible with PHPStan;
>     * Running PHPStan on PHP 7.2 would almost certainly not permit the latest version.

Need to first investigate if this is possible.  If it is, then decide if we also want to receive the warnings applicable to later PHP versions, somewow.  If it's not possible, then there's nothing we can do.

oliverklee:
This will help avoid it suggesting things that are only possile in later PHP versions.

Fixes #1214

### DIFF
--- a/config/phpstan.neon
+++ b/config/phpstan.neon
@@ -6,6 +6,8 @@ parameters:
     # Don't be overly greedy on machines with more CPU's to be a good neighbor especially on CI
     maximumNumberOfProcesses: 5
 
+  phpVersion: 70200
+
   level: 3
 
   paths:


JakeQZ (approved):

JakeQZ:
This isn't working as well as it should.  With PHP 8.3 on the server, it's picking up that `int<0, max>` and `int<1, max>` are not compatible:

```
  Line   src/Position/Position.php (in context of class                               
         Sabberworm\CSS\Tests\Unit\Position\Fixtures\ConcretePosition)                
 ------ ----------------------------------------------------------------------------- 
  62     Parameter #1 $lineNumber (int<1, max>|null) of method                        
         Sabberworm\CSS\Tests\Unit\Position\Fixtures\ConcretePosition::setPosition()  
         should be contravariant with parameter $lineNumber (int<0, max>|null)        
         of method Sabberworm\CSS\Position\Positionable::setPosition()                
         🪪  method.childParameterType                                                
```

But with PHP 7.4 on Windows, this error is not found.  In both cases, PHPStan version is 2.1.2.  Context is #1221 (where I will now correct the `Positionable::setPosition` DocBlock to `int<1, max>|null` for $lineNumber).

Oddly, this doesn't invlove a native declaration, but a DocBlock annotation, which should be independent of PHP version (target or runtime).

Also, the error `Parameters should have "Sabberworm\\CSS\\CSSList\\CSSListItem\|array" types as the only types passed to this method` has not gone away with this change. Native union types require PHP 8.0.

This change involved no changes to the baseline.  Whereas the Equivalent change to Emogrifier did.

I wonder if PHPStan is even picking up the setting at all.  A difference between the config files is that the `parameters` section of the one here begins with a `parallel` setting, whereas the one in Emogrifier begins with the `phpVersion` setting.

JakeQZ:
> This isn't working as well as it should. With PHP 8.3 on the server, it's picking up that `int<0, max>` and `int<1, max>` are not compatible:

Scrub that.  Just ran it again with PHP 7.4, and it picked up the error.  Maybe I didn't run the full set of tests before finally committing (I was playing around with nullability to check that if I'd got it wrong it would be picked up).

> Also, the error `Parameters should have "Sabberworm\\CSS\\CSSList\\CSSListItem\|array" types as the only types passed to this method` has not gone away with this change. Native union types require PHP 8.0.
> 
> This change involved no changes to the baseline. Whereas the equivalent change to Emogrifier did.
> 
> I wonder if PHPStan is even picking up the setting at all. A difference between the config files is that the `parameters` section of the one here begins with a `parallel` setting, whereas the one in Emogrifier begins with the `phpVersion` setting.

This is still a bit of a mystery, however.

PR #1217, opened 2025-03-25
[TASK] Use `CommentContainer` trait to implement `Commentable`

### LINKED ISSUE #813: Add `trait` for 'standard' `Commentable` implementation

JakeQZ:
Closes #813.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -4,8 +4,7 @@ declare(strict_types=1);
 
 namespace Sabberworm\CSS\CSSList;
 
-use Sabberworm\CSS\Comment\Comment;
-use Sabberworm\CSS\Comment\Commentable;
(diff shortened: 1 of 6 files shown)


src/CSSList/CSSList.php
```diff
@@ -4,8 +4,7 @@
 
 namespace Sabberworm\CSS\CSSList;
 
-use Sabberworm\CSS\Comment\Comment;
-use Sabberworm\CSS\Comment\Commentable;
```
JakeQZ, line 8: I don't know why the fixer didn't complain about this unsed `use` (`CSSListItem` extends `Commentable`, as of #1212, but when that change was made, `Commentable` was no longer directly referenced in this file).

oliverklee (approved):

Issue #1207, opened 2025-03-21
Add interface and trait for classes with a `lineNumber`


JakeQZ:
Some things to decide first:
- Should this also include column number (which may not be used in all cases)?
- What should the interface be called (`Positionable`?)?

And some changes that could be made in the course of this:
- Rename `getLineNo()` to `getLineNumber()` as follows:
  - Add both methods to the interface, with `getLineNo` marked as deprecated from 9.0, to be removed in 10.0;
  - The trait can handle forwarding `getLineNo` to `getLineNumber` (or simply access the property), so that no changes to implementing classes are required, provided they can use the trait;
- Use `null` to indicate 'no line number' instead of `0`:
  - Again, the trait should be able to take care of the details, avoiding changes to implementing classes.

oliverklee:
> Should this also include column number (which may not be used in all cases)?

Yes, I'd like that. I propose we make both properties optional and nullable for this.

> What should the interface be called (`Positionable`?)?

I quite like `Positionable`!

> Rename `getLineNo()` to `getLineNumber()` as follows: 

Yes, sounds good.

I propose `getLineNo()` then returns a non-nullable `int` (for backwards compatibality), while `getLineNumber()` returns `int<1, max>|null`.

I'd also like to move setting the position out of the constructors (in a separate PR). To make this possible, I think the interface should have (internal) setters for line number and column number. What do you think?

JakeQZ:
Agree with everything (particularly using `null` instead of `0` to indicate absence), up until the final comment:

> I'd also like to move setting the position out of the constructors (in a separate PR). To make this possible, I think the interface should have (internal) setters for line number and column number. What do you think?

I'd like to do the other bits first, then see how things look.  As part of a separate later PR, we can also add setters to the interface then, if we decide that's what we want to do.

(Traits can have constructors, but they are simply 'copied' to the using class like the other methods, so if the using class has it's own constructor, the trait's constructor will be replaced, and can't be called using `NameOfTrait::__construct()`, [though can with an additional `use`](https://stackoverflow.com/questions/47866408/execute-constructor-in-trait).)

JakeQZ:
> I'd also like to move setting the position out of the constructors (in a separate PR). To make this possible, I think the interface should have (internal) setters for line number and column number. What do you think?

`Rule` has a `setPosition()` method which sets both line and column number.  It's the only class that has a column number (AFAICS).

[...]

PR #1221, opened 2025-03-28
[TASK] Add `Positionable` interface and implementing trait

JakeQZ:
This is for CSS items which have a position in the document.

New methods are added:
- `getLineNumber` to replace `getLineNo`;
- `getColumnNumber` to replace `getColNo`.

These return a nullable `int`, instead of using zero to indicate absence.

The old methods are now deprecated, but defined in the interface and implemented in the trait.

Note that this change only adds the interface and trait.  It does not modify any classes to actually implement or use these.

Part of #1207.


JakeQZ:
Deprecation can be brought forward if it's possible to (easily) merge these changes to the 8.x branch.  Once the follow-up PR (to use the interfaces) is committed, we'll be able to see how possible that is.

JakeQZ:
Changelog entries will be added with the follow-up PR to implement and use the interface and trait.

I've fixed the typo in the commit message title.

src/Position/Position.php
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * @return int<0, max>
+     */
+    public function getColNo(): int
+    {
+        return $this->getColumnNumber() ?? 0;
+    }
+
+    /**
+     * @param int<0, max>|null $lineNumber
```
oliverklee, line 59: I propose we directly make this non-zero (also in the interface). Or am I missing something here?

```suggestion
     * @param int<1, max>|null $lineNumber
```

src/Position/Positionable.php
```diff
@@ -0,0 +1,45 @@
...
+
+    /**
+     * @deprecated in version 9.0.0, will be removed in v10.0. Use `getColumnNumber()` instead.
+     *
+     * @return int<0, max>
+     */
+    public function getColNo(): int;
+
+    /**
+     * @param int<0, max>|null $lineNumber
```
oliverklee, line 39: Ditto.

tests/Unit/Position/PositionTest.php
```diff
@@ -0,0 +1,284 @@
...
+    {
+        $this->subject->setPosition(98, 99);
+
+        $this->subject->setPosition($lineNumber, $columnNumber);
+
+        self::assertSame($lineNumber, $this->subject->getLineNumber());
+        self::assertSame($columnNumber, $this->subject->getColumnNumber());
+    }
+
+    // Testing deprecated functionality
```
oliverklee, line 195: These should reside in `Tests/UnitDeprecated/` instead.

oliverklee (changes requested):

src/Position/Position.php
JakeQZ: Passing `0` is deprecated, but we need to allow it until it is no longer supported.  The DocBlock comment for this parameter of the method in `Positionable` has a deprecation notice to that effect.  I didn't include a deprecation notice in the trait, because it's internal.

oliverklee: Ah, I had missed that `setPosition` already gets called. Then scratch my comments. :-)

tests/Unit/Position/PositionTest.php
JakeQZ: Done.  I could not think of an easy way to share the data providers between the two `TestCase`s - once you extend `TestCase`, it's `final` - so have duplicated them.  The tests for deprecated functionality will be removed in time, so it's no big deal.

src/Position/Position.php
JakeQZ: And it's part of the public API.

oliverklee (approved):

PR #1224, opened 2025-03-29
[TASK] Add to UML diagram that `Selector` implements `Renderable`

oliverklee:
Followup to #1017

### DIFF
--- a/README.md
+++ b/README.md
@@ -719,6 +719,7 @@ classDiagram
     Commentable <|.. RuleSet: realization
     RuleSet <|-- AtRuleSet: inheritance
     AtRule <|.. AtRuleSet: realization
+    Renderable <|.. Selector: realization
     Selector <|-- KeyframeSelector: inheritance
     Renderable <|-- AtRule: inheritance
     Commentable <|-- AtRule: inheritance


JakeQZ (approved):
There might be some other updates to make to the diagram in light of recent changes, e.g. introduction of `CSSListItem` interface.  And should traits be included in the diagram (e.g. `CommentContainer`)?

oliverklee:
>  And should traits be included in the diagram (e.g. `CommentContainer`)?

No, as far as class structures are concerned, traits do not really exists - for example, `instanceof` does not recognize traits. With this perspective, they are an implementation detail to avoid copy'n'pasting code.

oliverklee:
> There might be some other updates to make to the diagram in light of recent changes, e.g. introduction of `CSSListItem` interface.

Yes, indeed. Would you be willing to take care of that?

JakeQZ:
> Yes, indeed. Would you be willing to take care of that?

I have no suitable tool.  All I'd be able to do is commit a change on a branch, and see how it looks on GitHub...

oliverklee:
> All I'd be able to do is commit a change on a branch, and see how it looks on GitHub...

Actually, that's what I do too. 😉

PR #1225, opened 2025-03-30
[TASK] Implement `Positionable`

### LINKED ISSUE #1207: Add interface and trait for classes with a `lineNumber`

JakeQZ:
Closes #1207.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,12 @@ Please also have a look at our
 
 ### Added
 
+- Methods `getLineNumber` and `getColumnNumber` which return a nullable `int`
+  for the following classes:
+  `Comment`, `CSSList`, `SourceException`, `Charset`, `CSSNamespace`, `Import`,
+  `Rule`, `DeclarationBlock`, `RuleSet`, `CSSFunction`, `Value` (#1225)
+- `Positionable` interface for CSS items that may have a position
+  (line and perhaps column number) in the parsed CSS (#1221)
 - Partial support for CSS Color Module Level 4:
     - `rgb` and `rgba`, and `hsl` and `hsla` are now aliases (#797}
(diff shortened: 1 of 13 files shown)


JakeQZ:
If #1221 is accepted as it is now, this PR will be ready for review.  It includes the changes from #1221; there should not be a merge conflict, since they are the same.  I've marked it as 'blocked' (but not 'draft') for the time being...

JakeQZ:
Rebased so the diff doesn't now also include #1221.

JakeQZ:
Coveralls server is broken at the moment, returning HTTP 405 error.  We changed this to not fail the build if coverage went down.  Can we also change it to not fail the build if it fails to run - since it is information only?

JakeQZ:
> Coveralls server is broken at the moment, returning HTTP 405 error.

It says it is down for planned maintenance.

JakeQZ:
> > Coveralls server is broken at the moment, returning HTTP 405 error.
> 
> It says it is down for planned maintenance.

It's now back online, and all checks have passed.

oliverklee (changes requested):
Code looks good.

We also need to update the class diagram, though.

oliverklee (approved):

PR #1227, opened 2025-03-31
[TASK] Migrate to `rawr/phpunit-data-provider`

oliverklee:
The package `rawr/cross-data-providers` that we used has been abandoned and should not be used anymore.


JakeQZ (approved):
Still good.

PR #1231, opened 2025-04-02
[TASK] Add and implement `CSSElement` interface

JakeQZ:
Also add tests to confirm that the supplanted types in the DocBlock actually implement the new interface.

And correct a DocBlock type to also allow `string`, which is currently possible.

cf. #1230

PR #1237, opened 2025-04-08
[TASK] Add `assertInstanceOf` tests for `CSSListItem`

### LINKED ISSUE #1236: Add `assertInstanceOf` tests for newly introduced interfaces
Some new interface types have been added to replace an orred type list for some methods.  Possibly only `CSSListItem` (#1212) is relevant here.  (`Positionable` was also added, but this is not used as a type for any parameter, and only exists to complement the `Position` trait.)

During those changes, tests should have been added to confirm that the classes replaced in the parameter lists actually implement the replacement all-encompassing type.

#1231 has a similar change, and includes relevant tests.

JakeQZ:
These should probably have been added as part of #1212. They confirm that the various types supplanted by `CSSListItem` in the API all implement the new interface.

Resolves #1236.


JakeQZ:
This will have merge conflicts with #1231.

oliverklee (approved):

JakeQZ:
Just needs a final check that conflicts have been resolved correctly, particularly in `RuleSetTest`.

oliverklee (approved):

PR #1240, opened 2025-04-10
[TASK] Move `getAllValues()` to `CSSBlockList`

JakeQZ:
Also add unit tests for this method.

Part of #994.  Relates to #1230.

PR #1242, opened 2025-04-11
[TASK] Deconflate getAllValues() parameters

JakeQZ:
This is the backport of #1231, #1240 and #1241.

cf. #1230

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,12 +17,20 @@ This project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Changed
 
+- Parameters for `getAllValues()` are deconflated, so it now takes three (all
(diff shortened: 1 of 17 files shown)


JakeQZ:
Note that the first commit applies only #1231.  A further commit will fix errors with older PHP versions.  Then #1240 and #1241 will be merged in turn, probably with an additional commit for each to deal with older PHP versions.

JakeQZ:
Reviewing may be easier by viewing each commit separately.  Maybe that backport should have been separate PRs, but as the changes have already been approved on `main`, I don't see much value in that, given the steps taken can still be seen individually by viewing the separate commits.

oliverklee (approved):

PR #1243, opened 2025-04-12
[TASK] Remove original `getAllValues()` API

JakeQZ:
The method still exists with the same (slightly improved) functionality, but the optional arguments have been refactored, and may require changes.

Part of #994.  Closes #1230.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,12 @@ Please also have a look at our
 
 ### Removed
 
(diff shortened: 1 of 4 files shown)


JakeQZ:
#1242 should be merged before this, so that we have the deprecation in place first.

I'd hoped this change would resolve some PHPStan warnings, but that seems not to be the case :/

oliverklee (approved):

PR #1244, opened 2025-04-13
[TASK] Refactor `getAllValues()`

JakeQZ:
Move functionality from `allValues()` directly into to `getAllValues()`, so as to avoid passing array by reference, removing `allValues()`. Avoid making the recursive call for nested items that are not `CSSElement`s.

Part of #994.
Closes #1230.


oliverklee (approved):
LGTM.

PR #1249, opened 2025-05-02
[TASK] Add separate methods for `RuleSet::removeRule()`, with deprecation

JakeQZ:
Passing `string` or `null` to `removeRule()` is deprecated. The new methods handle that functionality.

Relates to #1247.

### DIFF
--- a/src/RuleSet/RuleSet.php
+++ b/src/RuleSet/RuleSet.php
@@ -217,7 +217,7 @@ abstract class RuleSet implements CSSElement, CSSListItem, Positionable
      * @param Rule|string|null $searchPattern
      *        `Rule` to remove.
      *        Passing a `string` or `null` is deprecated in version 8.9.0, and will no longer work from v9.0.
-     *        Use `removeMatchingRules()` instead.
+     *        Use `removeMatchingRules()` or `removeAllRules()` instead.
      */
     public function removeRule($searchPattern): void
     {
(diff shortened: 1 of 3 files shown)


src/RuleSet/RuleSet.php
```diff
@@ -235,18 +229,31 @@ public function removeRule($searchPattern): void
...
-                }
+            $this->removeMatchingRules($searchPattern);
+        }
+    }
+
+    /**
+     * Removes rules by property name or search pattern.
+     *
+     * @param string|null $searchPattern
+     *        pattern to remove. If null, all rules are removed. If the pattern ends in a dash,
```
oliverklee, line 240: As we're introducing new methods anyway, I'd suggest to drop the nullability and a dedicated method `removeAllRules` instead.

oliverklee (changes requested):
The new method(s) also should get tests.

JakeQZ:
> The new method(s) also should get tests.

I was planning to move the existing tests to the new method when the deprecated API was removed.  But you're right, they should be added now, and maybe can be more 'unit'.

JakeQZ:
Suggested changes made...

JakeQZ:
Note that I changed the PR title and OP message.

I was unable to do the same for the OC (original commit) due to subsequent commits.  Could you update the final commit message for `main` accordingly when you squash and merge?

oliverklee (approved):

PR #1253, opened 2025-05-05
[TASK] Drop allowing `Rule` to be passed to `RuleSet::getRules()`

JakeQZ:
... and `getRulesAssoc()`.


JakeQZ:
Curiously, the PHPStan warning "Parameters should have "string" types as the only types passed to this method" is only generated for `removeRule()`, despite all three methods having `Rule|string|null` as the parameter type.  Maybe `remove` in the method name has some special significance (as part of CRUD pattern) which makes it stricter.

JakeQZ:
> Curiously, the PHPStan warning "Parameters should have "string" types as the only types passed to this method" is only generated for `removeRule()`, despite all three methods having `Rule|string|null` as the parameter type. Maybe `remove` in the method name has some special significance (as part of CRUD pattern) which makes it stricter.

I figured out what's going on.  Even after changing `removeRule()` to only accept `Rule`, PHPStan still gave the warning against the method.  This turns out to be because some tests were still calling the method with `string` arguments.  I think PHPStan could do better with its reporting, and include the source of the problem.  Only when I ran the unit tests separately did it become clear, as there were then errors due to incorrect parameter types.

oliverklee (approved):

Issue #974, opened 2025-02-24
Rename `getLineNo()` to `getLineNumber()`


oliverklee:
- [ ] add `getLineNumber` as an alias in V8.x
- [ ] deprecate `getLineNo` in V8.x
- [ ] rename it in `main`

JakeQZ:
Partly done: #1225, #1233.

JakeQZ:
The original deprecation and other related changes may need reworking.

`Rule::addRule()` sets the column number to one less than the sibling.  Which means the default column number must be `0` not `null`, and that column numbers may be negative.  Not sure about line number, but it's used in sorting, so again having a `null` default may be problematic.  I think we need to revert back to using `0` as defaults for both (and not allow `null`), and backport this, with the end result that `getLineNumber` and `getColumnNumber` are just aliases to the pre-existing abbreviatedly-named methods.

oliverklee:
> `Rule::addRule()` sets the column number to one less than the sibling. 

If the sibling has `null` as the column number (which means "no column number known), the right thing (as I understand it) would be to set the column number of the other node to `null` as well. (We'd then need to take `null` into account when sorting as well.)

JakeQZ:
> If the sibling has `null` as the column number (which means "no column number known), the right thing (as I understand it) would be to set the column number of the other node to `null` as well.

Yes, I think that's the best we can do.

> (We'd then need to take `null` into account when sorting as well.)

Yeah, I'll need to take a look at how the sorting works.  It's possible that the change to nullable was slightly breaking (though perhaps `null` is now just getting automatically cast to `0`).

The question arises: how should `Rule`s with `null` position be sorted against `Rule`s with a defined position?

That said, I think (usually) `Rule`s always end up with a defined position, either from parsing, or when added (the `Rule` is assigned a line number one more than the current highest - though need to check what happens when there are no existing `Rule`s).  Except for `setRules()` which could set `Rule`s without position information.

> `Rule::addRule()` sets the column number to one less than the sibling.

This is problematic anyway if adding more than one `Rule`, but beyond the scope of this issue.

Hmm.  This initially seemed like a straightforward change !))

oliverklee:
I have another idea: We could enforce that all `Rules` within a `RuleSet` have a defined position. Maybe like this:

1. in `setRules()`, throw an exception if one of the set rules does not have a position
2. deprecate `setRules()` to be removed (people should use `addRule()` instead
3. remove `setRules()`

[...]

PR #1270, opened 2025-06-03
[BUGFIX] `AddRule` before sibling with different property name

JakeQZ:
Part of #974.


src/RuleSet/RuleSet.php
```diff
@@ -104,12 +104,29 @@ public function addRule(Rule $ruleToAdd, ?Rule $sibling = null): void
...
             $siblingPosition = \array_search($sibling, $this->rules[$propertyName], true);
             if ($siblingPosition !== false) {
+                $siblingIsInSet = true;
                 $position = $siblingPosition;
+            } elseif ($siblingIsInSet = $this->hasRule($sibling)) {
+                // Maintain ordering within `$this->rules[$propertyName]`
+                // by inserting before first `Rule` with a same-or-later position than the sibling.
+                foreach ($this->rules[$propertyName] as $index => $rule) {
+                    if (
+                        $rule->getLineNumber() > $siblingLineNumber ||
```
oliverklee: I propose we create a small, well-named helper method for `Rule` to reduce the complexity of this loop.

src/RuleSet/RuleSet.php
```diff
@@ -299,4 +317,15 @@ protected function renderRules(OutputFormat $outputFormat): string
 
         return $formatter->removeLastSemicolon($result);
     }
+
+    private function hasRule(Rule $rule): bool
+    {
+        foreach ($this->rules as $rulesForAProperty) {
```
oliverklee, line 328: Would it maybe make sense to also store the rules in a second array (not being linked to a property) to make searching for it easier/faster?

JakeQZ: Doing so might save a few opcodes here, though wouldn't improve upon O(n) for the method, and has downsides: removing a `Rule` by property name would then become quite slow, and the memory footprint would be increased.

src/RuleSet/RuleSet.php
JakeQZ: #1272 should help...

src/RuleSet/RuleSet.php
oliverklee: Okay, then let's keep this part as is for now.

src/RuleSet/RuleSet.php
```diff
@@ -104,9 +104,22 @@ public function addRule(Rule $ruleToAdd, ?Rule $sibling = null): void
         $position = \count($this->rules[$propertyName]);
 
         if ($sibling !== null) {
+            $siblingIsInSet = false;
             $siblingPosition = \array_search($sibling, $this->rules[$propertyName], true);
             if ($siblingPosition !== false) {
+                $siblingIsInSet = true;
                 $position = $siblingPosition;
+            } elseif ($siblingIsInSet = $this->hasRule($sibling)) {
```
oliverklee: Please let's separate the assignment and the check into two statements so that the condition is side-effect free.

oliverklee (changes requested):

src/RuleSet/RuleSet.php
JakeQZ: Done.

PR #1273, opened 2025-06-09
[TASK] Add unit tests for `RuleSet::removeRule`

JakeQZ:
This re-uses some data providers, which have been renamed to reflect their more generic usage.
Also, the PHPDoc type has been tightened to `non-empty-list` where applicable.

Part of #974.

### DIFF
--- a/tests/Unit/RuleSet/RuleSetTest.php
+++ b/tests/Unit/RuleSet/RuleSetTest.php
@@ -313,16 +313,32 @@ final class RuleSetTest extends TestCase
      *
      * @dataProvider provideInitialPropertyNamesAndIndexOfOne
      */
-    public function removeRuleRemovesRuleInSetAndKeepsOthers(array $initialPropertyNames, int $indexToRemove): void
+    public function removeRuleRemovesRuleInSet(array $initialPropertyNames, int $indexToRemove): void
     {
         $this->setRulesFromPropertyNames($initialPropertyNames);
         $ruleToRemove = $this->subject->getRules()[$indexToRemove];
 
         $this->subject->removeRule($ruleToRemove);
 
-        $remainingRules = $this->subject->getRules();
-        self::assertNotContains($ruleToRemove, $remainingRules);
-        self::assertCount(\count($initialPropertyNames) - 1, $remainingRules);
+        self::assertNotContains($ruleToRemove, $this->subject->getRules());
+    }
+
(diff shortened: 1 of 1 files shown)


tests/Unit/RuleSet/RuleSetTest.php
```diff
@@ -305,6 +305,46 @@ public function addRuleWithSiblingNotInRuleSetAddsRuleAfterInitialRulesAndSetsVa
...
 
+    /**
+     * @test
+     *
+     * @param non-empty-list<string> $initialPropertyNames
+     * @param int<0, max> $indexToRemove
+     *
+     * @dataProvider provideInitialPropertyNamesAndIndexOfOne
+     */
+    public function removeRuleRemovesRuleInSetAndKeepsOthers(array $initialPropertyNames, int $indexToRemove): void
```
oliverklee: I'd like us to split this into two tests: One that tests that the rule gets removed, and another that tests that the other rules stay. (Generally, an "and" is a code smell that points to a test secretly containing two tests.)

tests/Unit/RuleSet/RuleSetTest.php
```diff
@@ -305,6 +305,46 @@ public function addRuleWithSiblingNotInRuleSetAddsRuleAfterInitialRulesAndSetsVa
...
+    }
+
+    /**
+     * @test
+     *
+     * @param list<string> $initialPropertyNames
+     *
+     * @dataProvider provideInitialPropertyNamesAndAnotherPropertyName
+     */
+    public function removeRuleDoesNothingWithRuleNotInSet(
```
oliverklee: Let's reword this to describe what is happening instead of what isn't happening:

```suggestion
    public function removeRuleWithRuleNotInSetKeepsSetUnchanged
```

oliverklee (changes requested):

tests/Unit/RuleSet/RuleSetTest.php
JakeQZ: Done.  There are several other recently added tests in this class that could be similarly split.

tests/Unit/RuleSet/RuleSetTest.php
JakeQZ: Done.

oliverklee (approved):

PR #1275, opened 2025-06-12
[DOCS] Switch to the new CoC email address

oliverklee:


oliverklee:
This is an email list. At the moment, Jake, John and I have access (full access including admin access).

The main task is handling any CoC violation reports that might come in.

I currently don't see anything to maintain here (as long as we don't add/remove any members), but I might be missing something. What are you referring to?

sabberworm:
I meant maintain as in: respond to reports, make sure the address works and keeps working, etc.

oliverklee:
@sabberworm Would you like to join the list maintainers or the CoC incident responders?

PR #1276, opened 2025-06-13
[TASK] Add unit tests for `RuleSet::setRules`

JakeQZ:
Also rename a data provider to indicate its (now) more generic purpose.

### DIFF
--- a/tests/Unit/RuleSet/RuleSetTest.php
+++ b/tests/Unit/RuleSet/RuleSetTest.php
@@ -737,13 +737,13 @@ final class RuleSetTest extends TestCase
      *
      * @dataProvider providePropertyNames
      */
-    public function setRulesOnVirginSetsRules(array $propertyNamesToSet): void
+    public function setRulesOnVirginSetsRulesWithoutPositionInOrder(array $propertyNamesToSet): void
     {
         $rulesToSet = self::createRulesFromPropertyNames($propertyNamesToSet);
 
         $this->subject->setRules($rulesToSet);
 
-        self::assertArrayHasSameValues($rulesToSet, $this->subject->getRules());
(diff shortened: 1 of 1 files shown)


tests/Unit/RuleSet/RuleSetTest.php
```diff
@@ -730,16 +730,170 @@ public function removeAllRulesRemovesAllRules(array $propertyNamesToRemove): voi
...
             $propertyNames
-        ));
+        );
+    }
+
+    private static function assertArrayHasSameValues(array $expected, array $actual, string $message = ''): void
+    {
+        self::assertCount(\count($expected), $actual, $message);
+        foreach ($expected as $expectedElement) {
+            self::assertContains($expectedElement, $actual, $message);
```
oliverklee: We can shorten this to:

```php
self::assertSame(\array_values($expected), \array_values($actual));
```

oliverklee: I'm also not sure if we then still need the helper method, or if we can inline this line in the test methods then.

JakeQZ: That would then assert that the arrays have the same values in the same order.  I was avoiding testing the ordering (which is a behaviour of `getRules()`), though it is probably the simplest way of checking that the `Rule`s are assigned increasing line/column number from first to last (if they initially do not have a position), which should be tested.

JakeQZ: We don't need the `array_values` call either, since both arrays are `list`s.  I've changed the assertions to simply assert both arrays are the same, and also renamed the first test method to indicate that it's also checking the ordering (it doesn't seem worth having a separate method for checking the ordering, as it would be an exact copy).

JakeQZ:
Part of #974.

PR #1280, opened 2025-06-19
[TASK] Add tests for `RuleSet::getRulesAssoc` with `$searchPattern`

JakeQZ:
Part of #974.


tests/Unit/RuleSet/RuleSetTest.php
```diff
@@ -1142,6 +1142,72 @@ public function getRulesAssocKeysRulesByPropertyName(): void
...
+    public function getRulesAssocWithPatternReturnsAllMatchingPropertyNames(
+        array $propertyNamesToSet,
+        string $searchPattern,
+        array $matchingPropertyNames
+    ): void {
+        $this->setRulesFromPropertyNames($propertyNamesToSet);
+
+        $result = $this->subject->getRulesAssoc($searchPattern);
+
+        if ($matchingPropertyNames === []) {
```
oliverklee: Please let's split up the corresponding data provider in a pre-patch first, which would allow us to have two different tests with meaningful names.

tests/Unit/RuleSet/RuleSetTest.php
```diff
@@ -1142,6 +1142,72 @@ public function getRulesAssocKeysRulesByPropertyName(): void
...
+    public function getRulesAssocWithPatternFiltersNonMatchingRules(
+        array $propertyNamesToSet,
+        string $searchPattern,
+        array $matchingPropertyNames
+    ): void {
+        $this->setRulesFromPropertyNames($propertyNamesToSet);
+
+        $result = $this->subject->getRulesAssoc($searchPattern);
+
+        if ($result === []) {
```
oliverklee: Please let's split up the corresponding data provider in a pre-patch first, which would allow us to have two different tests with meaningful names.

tests/Unit/RuleSet/RuleSetTest.php
JakeQZ: Done in #1281.

tests/Unit/RuleSet/RuleSetTest.php
JakeQZ: Done in #1281.

tests/Unit/RuleSet/RuleSetTest.php
```diff
@@ -1165,6 +1165,84 @@ public function getRulesAssocKeysRulesByPropertyName(): void
...
+        array $propertyNamesToSet,
+        string $searchPattern,
+        array $matchingPropertyNames
+    ): void {
+        $this->setRulesFromPropertyNames($propertyNamesToSet);
+
+        $result = $this->subject->getRulesAssoc($searchPattern);
+
+        foreach ($matchingPropertyNames as $expectedMatchingPropertyName) {
+            self::assertContains($expectedMatchingPropertyName, \array_keys($result));
```
oliverklee: If I understand this code correctly, we can use `assertArrayHasKey` here.

https://docs.phpunit.de/en/9.6/assertions.html#assertarrayhaskey

tests/Unit/RuleSet/RuleSetTest.php
```diff
@@ -1165,6 +1165,84 @@ public function getRulesAssocKeysRulesByPropertyName(): void
...
+    public function getRulesAssocWithPatternFiltersNonMatchingRules(
+        array $propertyNamesToSet,
+        string $searchPattern,
+        array $matchingPropertyNames
+    ): void {
+        $this->setRulesFromPropertyNames($propertyNamesToSet);
+
+        $result = $this->subject->getRulesAssoc($searchPattern);
+
+        foreach ($result as $resultRule) {
```
oliverklee: I have a really hard time wrapping my head around what this test does … maybe some class in our project needs some well-named convenience methods to check for things? What do you think=

tests/Unit/RuleSet/RuleSetTest.php
JakeQZ: Yes.  I didn't think of that.

tests/Unit/RuleSet/RuleSetTest.php
JakeQZ: It is the converse of the previous test, which asserts that all matching `Rule`s are returned.  This test asserts that no non-matching `Rule`s are returned.

Perhaps the test could be better named, or more clearly written.  Or perhaps these two tests could be reduced to a single test that asserts the set of `Rule`s returned is as expected.

For `getRules()`, I also created two separate tests, one asserting that all matching `Rule`s were returned, and one asserting that no non-matching `Rule`s were returned.

For `getRulesAssoc()` there is the added complication that only one `Rule` per property name is returned (as documented in the DocBlock), which means the expected result may not contain all matching `Rule`s if they have the same property name.

Note that `Rule::getRule()` returns the property name and is a misnomer.

JakeQZ: Perhaps `getRulesAssocWithPatternReturnsOnlyMatchingRules()` would be a clearer test method name.

JakeQZ: Basically, this test asserts that every `Rule` in the result has a property name in `$matchingPropertyNames`.  It would be nice to avoid tranposing the meanings of `actual` and `expected` in the assertions, but I could not think of a way of doing so.

In this test we could bypass `Rule::getRule()` and test the array keys in the result instead, which should be the same (that they will be is covered by other tests).

oliverklee: > Basically, this test asserts that every `Rule` in the result has a property name in `$matchingPropertyNames`

Would it work to compare (sorted) arrays instead so we can drop the loop?

> In this test we could bypass `Rule::getRule()` and test the array keys in the result instead

Yes, I think that would be easier to grok. Let's do this!

oliverklee: > Perhaps `getRulesAssocWithPatternReturnsOnlyMatchingRules()` would be a clearer test method name.

I concur.

oliverklee: > Perhaps the test could be better named, or more clearly written. Or perhaps these two tests could be reduced to a single test that asserts the set of `Rules` returned is as expected.

Yes, I'd prefer that.

JakeQZ: > > Or perhaps these two tests could be reduced to a single test that asserts the set of `Rules` returned is as expected.
> 
> Yes, I'd prefer that.

I've modified `getRulesAssocWithPatternReturnsAllMatchingPropertyNames()` to assert that the list of property names returned exactly matches those expected, so it now also ensures no non-matching ones are returned.  (I haven't changed that test method name.)

> Would it work to compare (sorted) arrays instead so we can drop the loop?

Done.
 
> > In this test we could bypass `Rule::getRule()` and test the array keys in the result instead
> 
> Yes, I think that would be easier to grok. Let's do this!

Done.

> > Perhaps `getRulesAssocWithPatternReturnsOnlyMatchingRules()` would be a clearer test method name.
> 
> I concur.

This test has been removed, since it's now covered by the other test.

JakeQZ:
> Or perhaps these two tests could be reduced to a single test that asserts the set of `Rule`s returned is as expected.

There's a similar pair of tests for `getRules()` with a pattern: `getRulesWithPatternReturnsAllMatchingRules()` and `getRulesWithPatternFiltersNonMatchingRules()`.  Perhaps these should be similarly tidied up (combined into a single test).  However, these need to assert that the set of `Rule`s returned is as expected (not just property names, as there may be multiple `Rule`s for a single property).  Arrays of objects can't be reliably sorted, and I'm not aware of a PHPUnit method to assert two arrays have the same values in any order.  However, given that the `Rule`s (without pre-set positions) will be set and returned in the order provided, sorting shouldn't be necessary (and probably isn't for this PR either).

oliverklee (approved):

JakeQZ:
> There's a similar pair of tests for `getRules()` ... Perhaps these should be similarly tidied up (combined into a single test).

#1282

PR #1283, opened 2025-06-23
[TASK] Update `RuleSet::comparePositionable` to use new methods

JakeQZ:
`getLineNo` and `getColNo` are deprecated.
They were originally used in this method to ease backporting and transition to their replacement counterparts: `getLineNumber` and `getColumnNumber`, which differ by returning `null` in the case of 'not set'.

Part of #974


src/RuleSet/RuleSet.php
```diff
@@ -314,13 +314,33 @@ protected function renderRules(OutputFormat $outputFormat): string
...
+     *
+     * @throws \UnexpectedValueException if either argument does not have a valid position, which should never happen
      */
     private static function comparePositionable(Positionable $first, Positionable $second): int
     {
-        if ($first->getLineNo() === $second->getLineNo()) {
-            return $first->getColNo() - $second->getColNo();
+        $firstsLineNumber = $first->getLineNumber();
+        $secondsLineNumber = $second->getLineNumber();
+        if (!isset($firstsLineNumber, $secondsLineNumber)) {
```
oliverklee: Please let's use `is_int` instead as `getLineNumber` will always return a value, even if the value is `null`.

src/RuleSet/RuleSet.php
```diff
@@ -314,13 +314,33 @@ protected function renderRules(OutputFormat $outputFormat): string
...
+                'A Rule without a line number was passed to comparePositionable',
+                1750637683
+            );
         }
-        return $first->getLineNo() - $second->getLineNo();
+
+        if ($firstsLineNumber === $secondsLineNumber) {
+            $firstsColumnNumber = $first->getColumnNumber();
+            $secondsColumnNumber = $second->getColumnNumber();
+            if (!isset($firstsColumnNumber, $secondsColumnNumber)) {
```
oliverklee: Ditto.

src/RuleSet/RuleSet.php
JakeQZ: [`isset()` returns `false` for variables which have been assigned to `null`](https://www.php.net/manual/en/function.isset.php), so the behaviour will be the same.  It also allows testing multiple variables.  I suppose `is_int()` might be more futureproof, in case the interface was changed so that `false` rather then `null` represented 'no line number'.

JakeQZ: I've changed it to use `is_int` for enhanced strictness.

PR #1291, opened 2025-06-26
[TASK] Add `RuleContainerTest` trait, use in `RuleSetTest`

JakeQZ:
The trait provides tests for classes implementing `RuleContainer`. The test methods and data providers have been moved verbatim to the trait from `RuleSetTest`.

Will be needed for #1194.


tests/Unit/RuleSet/RuleContainerTest.php
```diff
@@ -0,0 +1,1198 @@
...
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Rule\Rule;
+use Sabberworm\CSS\RuleSet\RuleContainer;
+use TRegx\PhpUnit\DataProviders\DataProvider;
+
+/**
+ * This trait provides test methods for unit-testing classes that implement `RuleContainer`.
+ * It can be `use`d in a `TestCase` which has a `$subject` property that is an instance of the implementing class
+ * (the class under test), `setUp()` with default values.
+ */
```
oliverklee, line 18: We should add `@phpstan-require-extends TestCase` to enforce that this trait is only used in tests.


https://phpstan.org/writing-php-code/phpdocs-basics#enforcing-class-inheritance-for-interfaces-and-traits

JakeQZ: Done.

oliverklee (approved):

PR #1292, opened 2025-06-28
[BUGFIX] Allow comma-separated arguments in selectors

### LINKED ISSUE #360: Selectors inside :not() not correctly parsed
```css
 .test:not(.a, .b) { display: none; }
```

will be parsed as:

```
Sabberworm\CSS\RuleSet\DeclarationBlock -> getSelectors():
  Sabberworm\CSS\Property\Selector '.test:not(.a',
  Sabberworm\CSS\Property\Selector '.b)'
```

### LINKED ISSUE #1289: Incorrect selector splitting
Selectors like:

```css
#main .button:where(:hover, :focus)
.card :is(.title, p)
```

incorrectly get split up into 2 selectors.

JakeQZ:
Fixes #138.
Fixes #360.
Fixes #1289.

### DIFF
--- a/src/RuleSet/DeclarationBlock.php
+++ b/src/RuleSet/DeclarationBlock.php
@@ -88,6 +88,9 @@ class DeclarationBlock extends RuleSet
                         break;
                 }
             } while (!\in_array($nextCharacter, ['{', '}'], true) || \is_string($stringWrapperCharacter));
+            if ($functionNestingLevel !== 0) {
+                throw new UnexpectedTokenException(')', $nextCharacter);
+            }
             $selectors[] = \implode('', $selectorParts); // add final or only selector
             $result->setSelectors($selectors, $list);
             if ($parserState->comes('{')) {
--- a/tests/Unit/RuleSet/DeclarationBlockTest.php
+++ b/tests/Unit/RuleSet/DeclarationBlockTest.php
@@ -116,6 +116,40 @@ final class DeclarationBlockTest extends TestCase
     }
 
     /**
+     * @return array<non-empty-string, array{0: non-empty-string}>
+     */
+    public static function provideInvalidSelector(): array
+    {
+        // TODO: the `parse` method consumes the first character without inspection,
+        // so the 'lone' test strings are prefixed with a space.
+        return [
+            'lone `(`' => [' ('],
+            'lone `)`' => [' )'],
+            'unclosed `(`' => [':not(#your-mug'],
+            'extra `)`' => [':not(#your-mug))'],
+        ];
+    }
+
+    /**
+     * @test
+     *
+     * @param non-empty-string $selector
+     *
+     * @dataProvider provideInvalidSelector
+     */
+    public function parseSkipsBlockWithInvalidSelector(string $selector): void
+    {
+        static $nextCss = ' .next {}';
+        $css = $selector . ' {}' . $nextCss;
+        $parserState = new ParserState($css, Settings::create());
+
+        $subject = DeclarationBlock::parse($parserState);
+
+        self::assertNull($subject);
+        self::assertTrue($parserState->comes($nextCss));
+    }
+
+    /**
      * @return array<string>
      */
     private static function getSelectorsAsStrings(DeclarationBlock $declarationBlock): array


JakeQZ:
This is rough and ready, though fixes #1289.  I have no idea why an additional fix for comment parsing seems to be required.

I expect judicious review, and that this PR needs to be split because of the curious comment-parsing bug (why is nothing ever easy?).  Anyhow, we're on the road to sorting this.

src/RuleSet/DeclarationBlock.php
```diff
@@ -40,19 +40,48 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
-                    . $parserState->consumeUntil(['{', '}', '\'', '"'], false, false, $comments);
-                if (\in_array($parserState->peek(), ['\'', '"'], true) && \substr(\end($selectorParts), -1) != '\\') {
-                    if (!isset($stringWrapperCharacter)) {
-                        $stringWrapperCharacter = $parserState->peek();
-                    } elseif ($stringWrapperCharacter === $parserState->peek()) {
-                        unset($stringWrapperCharacter);
-                    }
+                    . $parserState->consumeUntil(['{', '}', '\'', '"', '(', ')', ','], false, false, $comments);
+                $nextCharacter = $parserState->peek();
+                switch ($nextCharacter) {
```
oliverklee: Please let's have a cleanup pre-patch that replaces the former `in_array` with the `switch`.

src/RuleSet/DeclarationBlock.php
```diff
@@ -40,19 +40,48 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
+                            $selectorParts = [];
+                            $parserState->consume(1);
+                        }
+                        break;
                 }
-            } while (!\in_array($parserState->peek(), ['{', '}'], true) || isset($stringWrapperCharacter));
-            $result->setSelectors(\implode('', $selectorParts), $list);
+            } while (!\in_array($nextCharacter, ['{', '}'], true) || isset($stringWrapperCharacter));
+            $selectors[] = \implode('', $selectorParts); // add final or only selector
+            $result->setSelectors($selectors, $list);
```
oliverklee, line 95: This cleanup should also go into a separate pre-patch.

src/Parsing/ParserState.php
```diff
@@ -345,6 +345,10 @@ public function consumeUntil(
         $start = $this->currentPosition;
 
         while (!$this->isEnd()) {
+            $comment = $this->consumeComment();
```
oliverklee: We should cover this with a test.

src/Property/Selector.php
```diff
@@ -24,9 +24,9 @@ class Selector implements Renderable
     public const SELECTOR_VALIDATION_RX = '/
         ^(
             (?:
-                [a-zA-Z0-9\\x{00A0}-\\x{FFFF}_^$|*="\'~\\[\\]()\\-\\s\\.:#+>]* # any sequence of valid unescaped characters
-                (?:\\\\.)?                                                     # a single escaped character
-                (?:([\'"]).*?(?<!\\\\)\\2)?                                    # a quoted text like [id="example"]
+                [a-zA-Z0-9\\x{00A0}-\\x{FFFF}_^$|*="\'~\\[\\]()\\-\\s\\.:#+>,]* # any sequence of valid unescaped characters
+                (?:\\\\.)?                                                      # a single escaped character
+                (?:([\'"]).*?(?<!\\\\)\\2)?                                     # a quoted text like [id="example"]
```
oliverklee: We should cover this with a test, and move it to a separate change.

oliverklee (changes requested):

src/Property/Selector.php
JakeQZ: Done in #1293.

src/RuleSet/DeclarationBlock.php
JakeQZ: #1294 for this.

src/RuleSet/DeclarationBlock.php
JakeQZ: That's not really plausible.  Until there is splitting at commas, `$selectorParts` will just be some strings to concatenate.  I guess it's put into an array for later concatenation as an optimization.

src/Parsing/ParserState.php
JakeQZ: Yeah, it's a bit mysterious why the bug this is fixing suddenly reared it's head with this change.  Will need investigating as a separate issue.

CHANGELOG.md
```diff
@@ -110,6 +110,8 @@ Please also have a look at our
 
 ### Fixed
 
+- Selector functions (like `:not`) with comma-separated arguments are now
+  parsed correclty (#1292)
```
oliverklee:
```suggestion
  parsed correctly (#1292)
```

oliverklee:
This PR now needs a rebase.

Apart from that, is it ready for a re-review, or does it need any other changes first?

CHANGELOG.md
JakeQZ: I must be somewhat dyslexic; I never seem to spot these mistakes myself :)

JakeQZ:
> This PR now needs a rebase.
> 
> Apart from that, is it ready for a re-review, or does it need any other changes first?

Have rebased.  The mysterious issue with comment consumption still needs to be resolved...

src/Parsing/ParserState.php
oliverklee: So how should we proceed here? Fix that bug first and then continue here? Or merge this PR first and than fix that bug (and cover it with tests)?

I tend to prefer the first, but am not completely sure.

WDYT?

oliverklee:
About the commit message: The `Fixes #xxxx` comments needs to be on a separate line each ticket ID for the auto-closing to work.

src/Parsing/ParserState.php
JakeQZ: The comment parsing bugfix seems to be needed for this PR (though I don't yet understand why), so it would need to be done first if it really is required for this change to work.

JakeQZ:
> About the commit message: The `Fixes #xxxx` comments needs to be on a separate line each ticket ID for the auto-closing to work.

It the docs are correct, [they can be on the same line, but do need to be written out individually](https://docs.github.com/en/issues/tracking-your-work-with-issues/using-issues/linking-a-pull-request-to-an-issue).  However, I think it's more readable if they are on separate lines, given the format I originally used won't work.  I've amended the original commit.

JakeQZ:
> The mysterious issue with comment consumption still needs to be resolved...

I've found that this arose because this PR changes consumption behaviour to consume the first character after the selector separator (`,`), regardless of what it is, which is a currently a mistake in this PR that needs to be addressed.

The ["Number 4" comment (here)](https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/tests/fixtures/comments.css#L7) is then the next thing `consumeUntil()` reads.

It so happens that there is a space after the comma, which if not pre-consumed, would allow `consumeUntil()` to consume the comment at the end of the loop.  So I think it will be possible to complete this PR without addressing the `consumeUntil()` bug, and construct tests which exhibit the latter.

JakeQZ:
> ... `consumeUntil()` bug, and construct tests which exhibit the latter.

#1320

JakeQZ:
> I've found that this arose because this PR changes consumption behaviour to consume the first character after the selector separator (`,`), regardless of what it is, which is a currently a mistake in this PR that needs to be addressed.

This is now fixed.

I think the `DeclarationBlock::parse` method could be refactored to be broken down into separate methods, but since this a backporting candidate, I'd rather do that later.

I think it would be possible to apply the change regarding `,` handling as a separate pre-PR cleanup, without the fix for bracket handling.  See what you think...

src/RuleSet/DeclarationBlock.php
```diff
@@ -58,9 +69,31 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
+                    case ',':
+                        if (!isset($stringWrapperCharacter) && $functionNestingLevel === 0) {
+                            $selectors[] = \implode('', $selectorParts);
+                            $selectorParts = [];
+                            $parserState->consume(1);
+                            $consumedNextCharacter = true;
+                        }
+                        break;
                 }
             } while (!\in_array($nextCharacter, ['{', '}'], true) || \is_string($stringWrapperCharacter));
```
sabberworm, line 90: The `isset($stringWrapperCharacter)` and `\is_string($stringWrapperCharacter)` are both checking the same thing, right? If so, maybe settle on using one of the two.

sabberworm (approved):

CHANGELOG.md
oliverklee:
```suggestion
  parsed correctly (#1292)
```

src/RuleSet/DeclarationBlock.php
```diff
@@ -40,12 +40,23 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
+            $functionNestingLevel = 0;
+            $consumedNextCharacter = false;
             do {
-                $selectorParts[] = $parserState->consume(1)
-                    . $parserState->consumeUntil(['{', '}', '\'', '"'], false, false, $comments);
+                if (!$consumedNextCharacter) {
+                    $selectorParts[] = $parserState->consume(1);
+                }
+                $selectorParts[] = $parserState->consumeUntil(
+                    ['{', '}', '\'', '"', '(', ')', ','],
```
oliverklee: To make this easier to understand, I propose moving this array either to a class constant or a well-named local variable. WDYT?

src/RuleSet/DeclarationBlock.php
```diff
@@ -58,9 +69,31 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
+                            }
+                            --$functionNestingLevel;
+                        }
+                        break;
+                    case ',':
+                        if (!isset($stringWrapperCharacter) && $functionNestingLevel === 0) {
+                            $selectors[] = \implode('', $selectorParts);
+                            $selectorParts = [];
+                            $parserState->consume(1);
+                            $consumedNextCharacter = true;
```
oliverklee: Unless I'm mistaken, `$consumedNextCharacter` is never read after line 49, and hence we can remove this line and line 59 (where it is written).

src/RuleSet/DeclarationBlock.php
```diff
@@ -58,9 +69,31 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
                             }
                         }
                         break;
+                    case '(':
+                        if (!isset($stringWrapperCharacter)) {
```
oliverklee: We can convert all those to `is_string`.

oliverklee (changes requested):

CHANGELOG.md
JakeQZ: I thought I already fixed this.  Maybe I forgot to do `git pull` to pick up the change applied with 'Apply suggestion', and subsequently overwote it with `git push -f`.

src/RuleSet/DeclarationBlock.php
JakeQZ: Perhaps a local variable, as it's only applicable to this method.  It also matches all the cases in the switch statement.  (`{` and `}` could be handled as switch cases rather than by the `while`, but it's clearer that these are loop end cases the way it's currently written.)

src/RuleSet/DeclarationBlock.php
JakeQZ: It was changed to `is_string` in a pre-PR, but I forgot to update it in this.

src/RuleSet/DeclarationBlock.php
JakeQZ: Yes, I forgot to update this after making the change in the pre-PR.

src/RuleSet/DeclarationBlock.php
JakeQZ: It's read in the next iteration of the loop, to determine whether the stop character still needs consuming.

tests/Unit/RuleSet/DeclarationBlockTest.php
```diff
@@ -48,4 +52,68 @@ public function implementsPositionable(): void
...
+     * @param non-empty-string $selector
+     *
+     * @dataProvider provideSelector
+     */
+    public function parsesAndReturnsSingleSelector(string $selector): void
+    {
+        $subject = DeclarationBlock::parse(new ParserState($selector . '{}', Settings::create()));
+
+        $resultSelectorStrings = \array_map(
+            static function (Selector $selectorObject): string {
```
oliverklee: We have this two times now - so I'd like to propose we move this to a well-named private method.

src/RuleSet/DeclarationBlock.php
JakeQZ: Done

src/RuleSet/DeclarationBlock.php
JakeQZ: Sorted.  It's reassuring to have two reviewers pick up on the same issue :)

src/RuleSet/DeclarationBlock.php
JakeQZ: Done.

tests/Unit/RuleSet/DeclarationBlockTest.php
JakeQZ: Good spot.  As a stickler for code re-use, I can't believe I missed this.  Sorted.

JakeQZ:
I've addressed all comments so far.  But I think there's a bit of a "smell" in the tests which have "parses *and* returns" in a single test.

> I think it would be possible to apply the change regarding `,` handling as a separate pre-PR cleanup, without the fix for bracket handling.

I'm increasingly thinking this would be a good approach, if only to separate the changes into manageable chunks.

oliverklee:
> I'm increasingly thinking this would be a good approach, if only to separate the changes into manageable chunks.

I'm confused - "this" (the good approach) is "having one more pre-patch" or "merging this PR first and then having a follow-up PR"?

JakeQZ:
> > I'm increasingly thinking this would be a good approach, if only to separate the changes into manageable chunks.
> 
> I'm confused - "this" (the good approach) is "having one more pre-patch" or "merging this PR first and then having a follow-up PR"?

Having another pre-PR to move the splitting of separators at commas into `parse()`.  Then this PR will just be the bugfix to not split them when they are inside function arguments.

JakeQZ:
> Having another pre-PR to move the splitting of separators at commas into `parse()`. Then this PR will just be the bugfix to not split them when they are inside function arguments.

The introduction of `$stopCharacters` can also be a separate pre-PR: #1321.

JakeQZ:
> I think there's a bit of a "smell" in the tests which have "parses _and_ returns" in a single test.

I've dealt with the "smell".  The tests are for parsing.  So I've renamed the methods and structured them into 'set up', 'run test', 'check results' on that basis in #1322 as another pre-PR.

JakeQZ:
> I think it would be possible to apply the change regarding `,` handling as a separate pre-PR cleanup, without the fix for bracket handling.

Indeed it is, and this fixes a separate bug: #1323.

JakeQZ:
Pre-PRs are all done.  This is ready for final review.

src/RuleSet/DeclarationBlock.php
```diff
@@ -64,8 +65,21 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
-                    case ',':
+                    case '(':
+                        if (!\is_string($stringWrapperCharacter)) {
+                            ++$functionNestingLevel;
+                        }
+                        break;
+                    case ')':
                         if (!\is_string($stringWrapperCharacter)) {
+                            if ($functionNestingLevel <= 0) {
+                                throw new UnexpectedTokenException('anything but', ')');
```
oliverklee, line 76: Is there any way we can have a test for this case?

JakeQZ: Yes, we can.  And we can also reject unclosed parentheses.  Have committed a change to address this.

There is an issue that the first character of the selector is consumed without inspection, but that is beyond the scope of this PR.

PR #1294, opened 2025-06-28
[CLEANUP] Tidy up `DeclarationBlock::parse()`

JakeQZ:
- Assign the result of `ParserState::peek()` to a local variable, for efficiency;
- Use a switch statement to branch on its value, for extensibility (e.g. #1292);
- Don't unnecessarily test that a quote character is not escaped when not within a string.

### DIFF
--- a/src/RuleSet/DeclarationBlock.php
+++ b/src/RuleSet/DeclarationBlock.php
@@ -44,14 +44,20 @@ class DeclarationBlock extends RuleSet
             do {
                 $selectorParts[] = $parserState->consume(1)
                     . $parserState->consumeUntil(['{', '}', '\'', '"'], false, false, $comments);
-                if (\in_array($parserState->peek(), ['\'', '"'], true) && \substr(\end($selectorParts), -1) != '\\') {
-                    if (!isset($stringWrapperCharacter)) {
-                        $stringWrapperCharacter = $parserState->peek();
-                    } elseif ($stringWrapperCharacter === $parserState->peek()) {
-                        unset($stringWrapperCharacter);
-                    }
+                $nextCharacter = $parserState->peek();
+                switch ($nextCharacter) {
+                    case '\'':
+                    case '"':
+                        if (!isset($stringWrapperCharacter)) {
+                            $stringWrapperCharacter = $nextCharacter;
+                        } elseif ($stringWrapperCharacter === $nextCharacter) {
+                            if (\substr(\end($selectorParts), -1) !== '\\') {
+                                unset($stringWrapperCharacter);
+                            }
+                        }
+                        break;
                 }
-            } while (!\in_array($parserState->peek(), ['{', '}'], true) || isset($stringWrapperCharacter));
+            } while (!\in_array($nextCharacter, ['{', '}'], true) || isset($stringWrapperCharacter));
             $result->setSelectors(\implode('', $selectorParts), $list);
             if ($parserState->comes('{')) {
                 $parserState->consume(1);


src/RuleSet/DeclarationBlock.php
```diff
@@ -44,14 +44,20 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
                     . $parserState->consumeUntil(['{', '}', '\'', '"'], false, false, $comments);
-                if (\in_array($parserState->peek(), ['\'', '"'], true) && \substr(\end($selectorParts), -1) != '\\') {
-                    if (!isset($stringWrapperCharacter)) {
-                        $stringWrapperCharacter = $parserState->peek();
-                    } elseif ($stringWrapperCharacter === $parserState->peek()) {
-                        unset($stringWrapperCharacter);
-                    }
+                $nextCharacter = $parserState->peek();
+                switch ($nextCharacter) {
+                    case '\'':
```
oliverklee, line 50:
```suggestion
                    case '\'':
                        // The fallthrough is intentional.
```

src/RuleSet/DeclarationBlock.php
```diff
@@ -44,14 +44,20 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
+                    case '"':
+                        if (!isset($stringWrapperCharacter)) {
```
oliverklee:
```suggestion
                        if (!\is_string($stringWrapperCharacter)) {
```

src/RuleSet/DeclarationBlock.php
```diff
@@ -44,14 +44,20 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
+                            $stringWrapperCharacter = $nextCharacter;
+                        } elseif ($stringWrapperCharacter === $nextCharacter) {
+                            if (\substr(\end($selectorParts), -1) !== '\\') {
+                                unset($stringWrapperCharacter);
```
oliverklee:
```suggestion
                                $stringWrapperCharacter = null;
```

src/RuleSet/DeclarationBlock.php
```diff
@@ -44,14 +44,20 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
+                            }
+                        }
+                        break;
                 }
-            } while (!\in_array($parserState->peek(), ['{', '}'], true) || isset($stringWrapperCharacter));
+            } while (!\in_array($nextCharacter, ['{', '}'], true) || isset($stringWrapperCharacter));
```
oliverklee:
```suggestion
            } while (!\in_array($nextCharacter, ['{', '}'], true) || \is_string($stringWrapperCharacter));
```

oliverklee (changes requested):
While we're at it, let's move the changed parts to nullable strings instead of set/not-set variables.

src/RuleSet/DeclarationBlock.php
JakeQZ: To me it seems obvious without the comment, but maybe not to others :)

JakeQZ:
> While we're at it, let's move the changed parts to nullable strings instead of set/not-set variables.

I've accepted all suggestions :)

oliverklee:
The red builds look like some PHIVE install hiccup. I've kicked the pipeline.

src/RuleSet/DeclarationBlock.php
```diff
@@ -44,14 +44,21 @@ public static function parse(ParserState $parserState, ?CSSList $list = null): ?
...
-                        $stringWrapperCharacter = $parserState->peek();
-                    } elseif ($stringWrapperCharacter === $parserState->peek()) {
-                        unset($stringWrapperCharacter);
-                    }
+                $nextCharacter = $parserState->peek();
+                switch ($nextCharacter) {
+                    case '\'':
+                        // The fallthrough is intentional.
+                    case '"':
+                        if (!\is_string($stringWrapperCharacter)) {
```
oliverklee, line 53: Ah, we need to initialize `$stringWrapperCharacter` to `null` before the `do` to make this work.

src/RuleSet/DeclarationBlock.php
oliverklee: Yes - the reason why I prefer this comment is that it allows the reader to differentiate between the case of "intentional fallthrough" vs. the case of "the dev forgot the `break`". :-)

oliverklee (changes requested):
This might also fix a PHPStan warning - so we might need to regenerate the baseline.

JakeQZ:
> This might also fix a PHPStan warning - so we might need to regenerate the baseline.

It does, but also `$stringWrapperCharacter` needs to be initialized in order to use `is_string` instead of `isset`.  Should be OK now.

src/RuleSet/DeclarationBlock.php
JakeQZ: If there was some actual code after the first `case`, that would make sense.  But for two (or more) consecutive `case` statements, it seems clear that the pursuant code is covering all those cases.  In that situation, the comment just adds clutter IMO :)

oliverklee: How would the reader then be able to know that the fall-through is intentional, and that the developer did not forget to put in the `break`?

JakeQZ: Because the developer created more-than-adequate unit tests which the reader trusts ;)

Leaving out the `break` for a `case` statement that does something is a common mistake.  Intentional fall-throughs from `case` statments that do something are rare, and these should be commented.  Empty `case` statements (with just an immediate `break`) are rare and only useful to prevent the `default`.  Grouping multiple `case`s with a common handler is common.

Thus, for me, alarm bells ring when seeing

```php
case 'A':
    doSomething();
case 'B':
    doSomethingElse();
    break;
```

but not for 
```php
case 'A':
case 'B':
    doSomething();
    break;
```

I've seen the grouped `case`s put on the same line, e.g.
```php
case 'A': case 'B':
    doSomething();
    break;
```

Perhaps that would make the intention clearer without requiring a comment, but may not sit will with some code formatters or format checkers.

oliverklee: I can understand this - and still would strongly prefer to have the comment even for empty cases. Would you be okay with that?

JakeQZ: That's fine.  It's just not what I'm used to :)

I accepted the suggested change anyway.

PR #1298, opened 2025-06-29
[DOCS] Regenerate the class diagram in the README

oliverklee:
The class diagram is (almost) unchanged to the output from `tasuku43/mermaid-class-diagram`, except for the kept `direction LR`.

This is in preparation for adding a script for auto-creating the class diagram in #1297.

### DIFF
--- a/README.md
+++ b/README.md
@@ -729,80 +729,130 @@ classDiagram
     CSSListItem <|-- AtRule: inheritance
     AtRule <|.. AtRuleBlockList: realization
     CSSBlockList <|-- AtRuleBlockList: inheritance
+    AtRuleBlockList ..> OutputFormat: dependency
     AtRule <|.. AtRuleSet: realization
+    AtRuleSet ..> OutputFormat: dependency
     RuleSet <|-- AtRuleSet: inheritance
     CSSList <|-- CSSBlockList: inheritance
     Renderable <|-- CSSElement: inheritance
+    CSSFunction ..> OutputFormat: dependency
+    CSSFunction ..> ParserState: dependency
     ValueList <|-- CSSFunction: inheritance
+    CSSList ..> AtRuleBlockList: dependency
+    CSSList ..> AtRuleSet: dependency
     CSSElement <|.. CSSList: realization
     CSSListItem <|.. CSSList: realization
+    CSSList ..> CSSNamespace: dependency
     CSSList ..> Charset: dependency
     CSSList ..> Import: dependency
+    CSSList ..> KeyFrame: dependency
+    CSSList ..> OutputFormat: dependency
+    CSSList ..> ParserState: dependency
     Positionable <|.. CSSList: realization
+    CSSList ..> Selector: dependency
+    CSSList ..> SourceException: dependency
+    CSSList ..> UnexpectedTokenException: dependency
     Commentable <|-- CSSListItem: inheritance
     Renderable <|-- CSSListItem: inheritance
     AtRule <|.. CSSNamespace: realization
+    CSSNamespace ..> OutputFormat: dependency
     Positionable <|.. CSSNamespace: realization
+    CSSString ..> OutputFormat: dependency
+    CSSString ..> ParserState: dependency
     PrimitiveValue <|-- CSSString: inheritance
+    CSSString ..> SourceException: dependency
     CSSFunction <|-- CalcFunction: inheritance
+    CalcFunction ..> CalcRuleValueList: dependency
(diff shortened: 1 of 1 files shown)


oliverklee:
Marking this as draft for now. I'll first reorder the classes in the diagram to make this PR here a bit smaller.

JakeQZ:
It would be nice if the diagram had fewer criss-crossing paths.  AI should be good at that, but seems to be vastly overrated :/

JakeQZ (commented):
The previous source for the diagram had entries like

```
URL --> "1" CSSString: url
```

whereas now we have

```
URL ..> CSSString: dependency
```

So we no longer identify the property that has the dependency (or whether it's as a single object, array, or whatever).  Is there any reason for this change?

oliverklee:
> So we no longer identify the property that has the dependency (or whether it's as a single object, array, or whatever). Is there any reason for this change?

The old version of the generator which we used did not recognize associations at all. That's why I manually added those to the diagram, including the name of the corresponding property.

The current version of the generator does recognize associations, even though it only labels them as "dependency", but not with the name of the corresponding property.

JakeQZ (changes requested):
I spent a bit of time reordering the original 'relationships' section, after changing to the now-generated syntax, to conclude that there a quite a few differences there.

I think this needs three separate pre-patches so the diffs can be more clearly seen (along the lines of #1299):
1. Move `class URL` to be ASCII-sorted;
1. Change to use `..>`, `dependency` and remove `"*"`/`"1"`;
1. Reorder the relationships to be consistent with the generated order.

For reference, what I did for reviewing is on the branch at https://github.com/MyIntervals/PHP-CSS-Parser/tree/docs/class-diagram-review and includes all three of the above, which need to be split.  And I may have made mistakes: there was a lot of cutting and pasting

oliverklee:
> I think this needs three separate pre-patches so the diffs can be more clearly seen

Makes sense, will do.

README.md
```diff
@@ -730,76 +730,130 @@ classDiagram
     CSSListItem <|-- AtRule: inheritance
     AtRule <|.. AtRuleBlockList: realization
     CSSBlockList <|-- AtRuleBlockList: inheritance
+    AtRuleBlockList ..> OutputFormat: dependency
     AtRule <|.. AtRuleSet: realization
+    AtRuleSet ..> OutputFormat: dependency
```
JakeQZ, line 734: Looks like it's now also picking up dependencies that are method parameters, not just properties.

README.md
```diff
@@ -730,76 +730,130 @@ classDiagram
...
+    Value ..> CSSFunction: dependency
+    Value ..> ParserState: dependency
     Positionable <|.. Value: realization
+    Value ..> RuleValueList: dependency
+    Value ..> UnexpectedTokenException: dependency
+    ValueList ..> OutputFormat: dependency
     Value <|-- ValueList: inheritance
-
-    CSSList ..> CSSList: dependency
-    CSSList ..> Comment: dependency
```
JakeQZ, line 797: Looks like it's losing dependencies that are class properties via traits.  The `CommentContainer` trait was added for classes that have `Comment`s.  Given that traits are not included in the class diagram, this looks to me like a bug in the generator (for which I think the better fix would be to include traits in their own right, like interfaces and classes).

README.md
```diff
@@ -730,76 +730,130 @@ classDiagram
...
-    CSSList ..> RuleSet: dependency
-    CSSNamespace ..> Comment: dependency
-    Charset ..> Comment: dependency
-    Import ..> Comment: dependency
-    OutputFormat ..> OutputFormat: dependency
-    Rule ..> Comment: dependency
-    RuleSet ..> Comment: dependency
-    ValueList ..> Value: dependency
```
JakeQZ, line 805: `ValueList` extents `Value` and also has `Value`s as properties.  It looks like the latter connection is being lost.

JakeQZ (commented):
It looks like it's adding all dependencies, even if they are only passed as method parameters, and not actually stored in class properties.  This rather clutters the class diagram.  The current one is reasonably easy to understand.  The diagram after this change harder to understand and about three times the size.

Would it be possible to exclude dependencies that are only ever used as method parameters?

There seems to be a bug in that traits are completely ignored.

oliverklee:
Okay, putting this on the backburner until the diagram generator has these two bugfixes/features.

I've created two tickets for that:

- https://github.com/tasuku43/php-mermaid-class-diagram/issues/26
- https://github.com/tasuku43/php-mermaid-class-diagram/issues/27

PR #1299, opened 2025-06-29
[CLEANUP] Order class names in the class diagram alphabetically

oliverklee:


README.md
```diff
@@ -621,109 +621,108 @@ class Sabberworm\CSS\CSSList\Document#4 (2) {
 ```mermaid
 classDiagram
     direction LR
-
```
JakeQZ, line 624: I'd prefer the blank line between the settings and the content to be retained.

README.md
```diff
@@ -621,109 +621,108 @@ class Sabberworm\CSS\CSSList\Document#4 (2) {
...
     }
-    class KeyframeSelector {
+    class CSSNamespace {
     }
-    class AtRule {
-        <<interface>>
+    class CSSString {
+    }
+    class CalcFunction {
+    }
```
JakeQZ, line 655: In my alphabet, 'a' comes before 's'.

JakeQZ (changes requested):
I'm not entirely sure of the rationale behind the reordering, but since there was no specific order prevoiusly (other than to try and get the class diagram to render more clearly, which didn't work), it's fine.

However, I think lexicographical ordering would be better than ASCII, and the blank line after the config settings should be kept.

README.md
oliverklee: Done.

oliverklee:
> I'm not entirely sure of the rationale behind the reordering

The rationale is to make the diff smaller in the [PR that uses the updated autogenerated class diagram](https://github.com/MyIntervals/PHP-CSS-Parser/pull/1298).

README.md
oliverklee: The diagram generator we're using seems to sort an uppercase "S" before a lowercase "a", [it seems](https://github.com/MyIntervals/PHP-CSS-Parser/pull/1298/files).

PR #1300, opened 2025-06-29
[CLEANUP] Avoid spaces before colons in the Mermaid code

oliverklee:


JakeQZ (approved):
Class diagram still works without them :)

PR #1306, opened 2025-07-05
[DOCS] Reorder some line in the class diagram (part 1)

oliverklee:


README.md
```diff
@@ -727,6 +727,14 @@ classDiagram
         <<abstract>>
     }
 
+    Anchor ..> ParserState: dependency
+    CSSListItem <|-- AtRule: inheritance
+    AtRule <|.. AtRuleSet: realization
+    CSSBlockList <|-- AtRuleBlockList: inheritance
+    AtRule <|.. AtRuleBlockList: realization
+    %% end of the generated part
```
JakeQZ: The comment no longer makes sense after the reordering.

oliverklee:
> I think the comments ("start/end of generated part") should be removed with this change, since they no longer apply. (Ultimately it will all be generated, so they will become unnecessary anyway.)

I've temporarily removed them in #1307.

When we autogenerate the whole diagram again, we'll need to have those markers (or something similar) back, as we need to have something for the [generation and insertion script](https://github.com/MyIntervals/PHP-CSS-Parser/pull/1297) to recognize which part to replace.

README.md
oliverklee: Done in #1307, merged and repushed.

JakeQZ (approved):

PR #1310, opened 2025-07-07
[DOCS] Reorder some lines in the class diagram (part 4)

oliverklee:


JakeQZ (approved):
I'm guessing there could be up to 26 of these... ;)

oliverklee:
> I'm guessing there could be up to 26 of these... ;)

Hopefully not quite as many …

PR #1312, opened 2025-07-07
[DOCS] Reorder some lines in the class diagram (part 6)

oliverklee:


oliverklee:
Hold on, I think there's a line missing …

oliverklee:
> I think there's a line missing …

No, it isn't.

PR #1333, opened 2025-07-14
[CLEANUP] Remove superfluous Rector rule

JakeQZ:
Since #1201, this becomes a duplicate.

### DIFF
--- a/config/rector.php
+++ b/config/rector.php
@@ -35,7 +35,4 @@ return RectorConfig::configure()
         PHPUnitSetList::PHPUNIT_80,
         // PHPUnitSetList::PHPUNIT_CODE_QUALITY,
     ])
-    ->withRules([
-        AddVoidReturnTypeWhereNoReturnRector::class,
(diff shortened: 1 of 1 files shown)


oliverklee (approved):

config/rector.php
```diff
@@ -35,7 +35,4 @@
         PHPUnitSetList::PHPUNIT_80,
         // PHPUnitSetList::PHPUNIT_CODE_QUALITY,
     ])
-    ->withRules([
-        AddVoidReturnTypeWhereNoReturnRector::class,
```
oliverklee, line 39: We also need to remove the corresponding `use`. (This is what PHP-CS-Fixer is complaining about.)

oliverklee (changes requested):

config/rector.php
JakeQZ: Done.  Sorry, I forgot to run the full set of CI checks.

PR #1338, opened 2025-07-17
[BUGFIX] Correct DocBlock for `ParserState::consumeUntil()`

JakeQZ:
The special `EOF` constant is actually defined as `null`, so the stop characters may be strings or `null`.

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -328,7 +328,7 @@ class ParserState
     }
 
     /**
-     * @param list<string>|string $stopCharacters
+     * @param list<string|null>|string|null $stopCharacters
      * @param array<int, Comment> $comments
      *
      * @throws UnexpectedEOFException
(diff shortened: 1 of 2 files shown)


src/Parsing/ParserState.php
```diff
@@ -328,7 +328,7 @@ public function isEnd(): bool
     }
 
     /**
-     * @param list<string>|string $stopCharacters
+     * @param list<string|null>|string|null $stopCharacters
```
oliverklee, line 331:
```suggestion
     * @param list<string|self::EOF>|string|[null](self::EOF) $stopCharacters
```

oliverklee: We can make the reasoning behind this a bit more obvious to the reader:

```suggestion
     * @param list<string|self::EOF>|string|self::EOF $stopCharacters
```

PR #1340, opened 2025-07-18
[BUGFIX] Provide the authentication token for PHIVE on CI

### LINKED ISSUE #1339: Composer-Normalize keeps failing to download
This is now happening on almost every other commit.  It's a PITA.  Are there alternatives, or alternative sources?

Image

oliverklee:
This will hopefully avoid the 403 errors when installing the PHIVE packages.

Fixes #1339


JakeQZ (approved):
🤞

PR #1341, opened 2025-07-18
[TASK] Make `RuleSet` concrete

JakeQZ:
... adding internal `render` method.

Precursor to #1194.

### DIFF
--- a/tests/Functional/RuleSet/RuleSetTest.php
+++ b/tests/Functional/RuleSet/RuleSetTest.php
@@ -1,0 +1,119 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Functional\RuleSet;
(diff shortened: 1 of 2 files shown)


JakeQZ:
> the tests ... also test the effects of `OutputFormat`/`OutputFormatter`.

Only insofar as the expected CSS matches that rendered with the default `OutputFormat` - there aren't any tests for varying the spacing.

JakeQZ:
> > the tests ... also test the effects of `OutputFormat`/`OutputFormatter`.
> 
> Only insofar as the expected CSS matches that rendered with the default `OutputFormat` - there aren't any tests for varying the spacing.

I note the equivalent tests for other classes are in the `Functional` department, and also test with `pretty` and `compact` formats.  So this should probably test those too...

PR #1345, opened 2025-07-21
[BUGFIX] Remove trailing semicolon with compact format

### LINKED ISSUE #1342: `OutputFormat::createCompact` should strip trailing semicolons
... from declaration blocks.

The option already exists, but is not currently set by the factory method, which, based on its name, would reasonably be expected to set options for the most minified CSS possible.

Relates to tests added for #1341.

JakeQZ:
Fixes #1342.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,7 +74,7 @@ Please also have a look at our
 
 ### Fixed
 
-- Remove trailing semicolon from declatation blocks with 'compact'
+- Remove trailing semicolon from declaration blocks with 'compact'
   `OutputFormat` (#1345)
 - Parse selector functions (like `:not`) with comma-separated arguments (#1292)
 - Parse quoted attribute selector value containing comma (#1323)


src/OutputFormat.php
```diff
@@ -726,6 +726,7 @@ public static function createCompact(): self
             ->setSpaceAfterRuleName('')
             ->setSpaceBeforeOpeningBrace('')
             ->setSpaceAfterSelectorSeparator('')
+            ->setSemicolonAfterLastRule(false)
```
oliverklee, line 729: There's also a list of unit tests for `OutputFormat` where we should add a testcase for this:

https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/tests/Unit/OutputFormatTest.php#L1040

JakeQZ: Good point.  I've added a test there.

CHANGELOG.md
```diff
@@ -74,6 +74,8 @@ Please also have a look at our
 
 ### Fixed
 
+- Remove trailing semicolon from declatation blocks with 'compact'
```
oliverklee:
```suggestion
- Remove trailing semicolon from declaration blocks with 'compact'
```

oliverklee (approved):

PR #1346, opened 2025-07-21
[CLEANUP] Use `getAllDeclarationBlocks` in `colorParsing` test

JakeQZ:
... instead of `getAllRuleSets`.

This avoids testing if `RuleSet`s are `DeclarationBlock`s, and will be needed for #1194.

### DIFF
--- a/tests/ParserTest.php
+++ b/tests/ParserTest.php
@@ -91,10 +91,7 @@ final class ParserTest extends TestCase
     public function colorParsing(): void
     {
         $document = self::parsedStructureForFile('colortest');
-        foreach ($document->getAllRuleSets() as $ruleSet) {
-            if (!($ruleSet instanceof DeclarationBlock)) {
-                continue;
-            }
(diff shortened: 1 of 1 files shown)


tests/ParserTest.php
```diff
@@ -91,10 +91,7 @@ public function files(): void
     public function colorParsing(): void
     {
         $document = self::parsedStructureForFile('colortest');
-        foreach ($document->getAllRuleSets() as $ruleSet) {
-            if (!($ruleSet instanceof DeclarationBlock)) {
-                continue;
-            }
+        foreach ($document->getAllDeclarationBlocks() as $ruleSet) {
```
oliverklee: Shouldn't we rename `$ruleSet` to `$declarationBlock` then to make its specific type more clear?

JakeQZ: Good point.  I've also renamed a similar local variable in another test method.

PR #1350, opened 2025-07-24
[TASK] Add tests for `DeclarationBlock` constructor

JakeQZ:
The class extends `RuleSet`, but the constructor behaviour needs to be tested for each class.

### DIFF
--- a/tests/Unit/RuleSet/DeclarationBlockTest.php
+++ b/tests/Unit/RuleSet/DeclarationBlockTest.php
@@ -64,11 +64,12 @@ final class DeclarationBlockTest extends TestCase
     }
 
     /**
-     * @return array<non-empty-string, array{0: int<1, max>}>
+     * @return array<non-empty-string, array{0: int<1, max>|null}>
      */
     public function provideLineNumber(): array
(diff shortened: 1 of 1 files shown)


tests/Unit/RuleSet/DeclarationBlockTest.php
```diff
@@ -53,6 +53,43 @@ public function implementsPositionable(): void
...
+        self::assertNull($result);
+    }
+
+    /**
+     * @return array<non-empty-string, array{0: int<1, max>}>
+     */
+    public function provideLineNumber(): array
+    {
+        return [
+            'line 1' => [1],
```
oliverklee, line 73: We also should have `null` (as that's an allowed value, too).

JakeQZ: Added.  Will also do the same for #1348.

JakeQZ: > Will also do the same for #1348.

#1351.

oliverklee (approved):

PR #1353, opened 2025-07-26

oliverklee:


oliverklee:
The last beta of PHP 8.5 [has been released](https://www.php.net/archive/2025.php#2025-09-11-1). So let's get a new version of PHP-CSS-Parser released with support for PHP 8.5. (These updates contains some fixes for PHP 8.5, which is why I'd like to have them for the release.)

PR #1367, opened 2025-08-21
[BUGFIX] Use the safe `file_get_contents` in `quickdump.php`

oliverklee:


bin/quickdump.php
```diff
@@ -3,13 +3,15 @@
...
+
 /**
  * This script is used for generating the examples in the README.
  */
 
 require_once(__DIR__ . '/../vendor/autoload.php');
 
-$sSource = \file_get_contents('php://stdin');
+$sSource = file_get_contents('php://stdin');
 $oParser = new Sabberworm\CSS\Parser($sSource);
```
JakeQZ, line 15: I would have expected a warning from Stan about possibly passing `false` to a method that expects a `string`.

JakeQZ (approved):

bin/quickdump.php
oliverklee: Yeah, me too.

PR #1368, opened 2025-08-21
[BUGFIX] Use the safe regexp functions in `CSSList`

oliverklee:


oliverklee:
> The `use` statement is missing. Is there an automated way that we can catch such mistakes, as some are bound to be made and also missed in review?

Yes - the https://packagist.org/packages/thecodingmachine/phpstan-safe-rule ruleset will catch this. (We'll first need to convert all case to the safe versions, though.)

JakeQZ:
> Yes - the https://packagist.org/packages/thecodingmachine/phpstan-safe-rule ruleset will catch this. (We'll first need to convert all case to the safe versions, though.)

I saw this, but it won't help with the transition period, which is what I'm worried about.

Issue #1168, opened 2025-03-14
Switch the native function calls to `thecodingmachine/safe`


oliverklee:
[This package](https://packagist.org/packages/thecodingmachine/safe) provides versions of native PHP functions that throw an exception in case of an error instead of returning `false`. We also should use the [corresponding set of PHPStan rules](https://packagist.org/packages/thecodingmachine/phpstan-safe-rule) to make sure we use the safe version of functions.

Due to us supporting PHP 7.x as well, we need to use lower versions of these packages:

- thecodingmachine/safe 1.3.3
- thecodingmachine/phpstan-safe-rule 1.2.0 || 1.4.0 (to allow PHPStan 2)

PR #1371, opened 2025-08-24
[BUGFIX] Use the safe regexp functions in `Selector`

oliverklee:
Part of #1168


oliverklee:
Our current regexp(es) is/are throwing `PREG_BACKTRACK_LIMIT_ERROR: Backtrack limit reached` errors (which got covered up so far). So we probably need to fix out regexp(es) first.

I'm out of my depth here with my regexp knowledge, though. @JakeQZ Would you be willing to look into this and create a pre-patch for this?

JakeQZ:
> Our current regexp(es) is/are throwing `PREG_BACKTRACK_LIMIT_ERROR: Backtrack limit reached` errors (which got covered up so far). So we probably need to fix out regexp(es) first.
> 
> I'm out of my depth here with my regexp knowledge, though. @JakeQZ Would you be willing to look into this and create a pre-patch for this?

#1372 should do the job.  There's a link there for explanation.

tests/ParserTest.php
```diff
@@ -73,11 +73,7 @@ public function files(): void
                     continue;
                 }
                 $parser = new Parser(\file_get_contents($directory . '/' . $filename));
-                try {
-                    self::assertNotEquals('', $parser->parse()->render());
-                } catch (\Exception $e) {
-                    self::fail($e);
-                }
+                self::assertNotSame('', $parser->parse()->render());
```
JakeQZ, line 76: The test should fail anyway if an exception is thrown.  `assertNotSame()` is a weaker assertion than `assertNotEquals()`, but `render()` can only return a string.

PR #1372, opened 2025-08-25
[BUGFIX] Improve selector validation performance

JakeQZ:
Avoid [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html) in selector validation regular expression by using possessive quantifier with mutually exclusive alternations.

Also remove outdated description from DocBlock, but add description for extended class summarizing differences.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,8 @@ Please also have a look at our
 
 ### Fixed
 
+- Improve performance of selector validation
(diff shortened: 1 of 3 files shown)


JakeQZ:
> add description for extended class summarizing differences.

The differences are so small; it would be good to avoid duplicating most of the main regular expression, but beyond the scope of this PR.

PR #1355, opened 2025-07-27
[FEATURE] Add support for PHP 8.5

oliverklee:

PR #1374, opened 2025-08-28
feat: Add PHP 8.5 Support

gabrielrbarbosa:
Since php 8.5 beta release is already out we should change this composer.json requirement.


oliverklee:
Thanks, @gabrielrbarbosa!

We already have a PR for this: #1355

Please let's continue there.

PR #1378, opened 2025-09-14
[BUGFIX] Use safe file functions in `ParserTest`

oliverklee:
Part of #1168


JakeQZ (approved):

JakeQZ (changes requested):
This should have a new changelog entry, for the next release.

oliverklee:
> This should have a new changelog entry, for the next release.

For this particular PR, I don't think we need one as this is about testcases only and doesn't affect production code.

JakeQZ (approved):

PR #1380, opened 2025-09-16
[BUGFIX] Use safe preg functions in `Size`

oliverklee:
Also use typesafe comparisons in the affected line.

Part of #1168


JakeQZ (approved):
Having previously introduced the backslash prefix helps make it clear which functions are now using the `Safe` versions :)).

PR #1381, opened 2025-09-16

oliverklee:


oliverklee:
Let's get this in so both projects can be in sync with their [PHP-CS-Fixer configuration](https://github.com/MyIntervals/emogrifier/pull/1507).

PR #1385, opened 2025-09-19
[TASK] Add PHPStan rules for Safe-PHP

### LINKED ISSUE #1168: Switch the native function calls to `thecodingmachine/safe`

oliverklee:
This will prevent unsafe function usage from getting added.

Closes #1168

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,7 @@ updates:
           versions: [ ">= 9.0.0" ]
         - dependency-name: "rector/rector"
         - dependency-name: "thecodingmachine/safe"
+        - dependency-name: "thecodingmachine/phpstan-safe-rule"
       versioning-strategy: "increase"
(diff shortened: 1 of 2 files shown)


JakeQZ (approved):
As a quick test, I commented out one of the `use` statements to confirm that an error is reported.

PR #1390, opened 2025-11-03
allow installation on PHP 8.5

Khuthaily:
This patch widens the PHP version constraint to include PHP 8.5.  This change ensures continued compatibility for downstream packages such as `dompdf/dompdf`, which indirectly depend on this package through `php-svg-lib`.

The reason for this change is to keep the `8.x` branch viable for projects that are not yet ready to migrate to `sabberworm v9`.


oliverklee:
@Khuthaily Thank your for this contribution!

We currently do not support the 8.x branch anymore. So we recommend that projects switch to version 9.x if they want to use the new features (which includes support for PHP 8.5).

Do you know if there is any blocker for the `php-svg-lib` project to either switch to version 9.x of our library, or to at least allow both 8.x and 9.x? (If I understand this correctly, the [existing PR](https://github.com/dompdf/php-svg-lib/pull/132) would solve this.)

oliverklee:
Closing as there has been no feedback for over a week, and the alternative solution https://github.com/dompdf/php-svg-lib/pull/132 aligns more with our maintenance policy.

@Khuthaily Please feel free to comment if you think that php-svg-lib cannot or will not merge the linked PR. Then we can discuss what to do instead.

PR #1391, opened 2025-11-13
[CLEANUP] Reformat the code with the PER-2 configuration

oliverklee:


oliverklee:
I don't understand why these changes cause PHP linting to fail in PHP 7.2 to fail. :exploding_head: 

@JakeQZ Any ideas?

tests/ParserTest.php
```diff
@@ -571,10 +571,10 @@ public function comments(): void
...
         $document = self::parsedStructureForFile('comments');
         $expected = <<<EXPECTED
-@import url("some/url.css") screen;
-.foo, #bar {background-color: #000;}
-@media screen {#foo.bar {position: absolute;}}
-EXPECTED;
+            @import url("some/url.css") screen;
+            .foo, #bar {background-color: #000;}
+            @media screen {#foo.bar {position: absolute;}}
+            EXPECTED;
```
JakeQZ: Indenting this will probably break the test, because it will include the additional whitespace in the expected result.  For readability, it might be better constructing the string by concatenating substrings.

The Lint failure in PHP 7.2 is because the heredoc closing identifier must be at the beginning of the line.  From the [docs](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc):

> The closing identifier may be indented by space or tab, in which case the indentation will be stripped from all lines in the doc string. Prior to PHP 7.3.0, the closing identifier _must_ begin in the first column of the line.

So actually, indenting this will not break the tests for PHP >= 7.3, but will for PHP 7.2.

tests/Comment/CommentTest.php
```diff
@@ -19,7 +19,8 @@ final class CommentTest extends TestCase
     public function keepCommentsInOutput(): void
     {
         $cssDocument = TestsParserTest::parsedStructureForFile('comments');
-        self::assertSame('/** Number 11 **/
+        self::assertSame(
+            '/** Number 11 **/
```
JakeQZ: I can't read this.

oliverklee:
> Could we first have some further fixes to the strings so that they are not cluttering up the view of the code (i.e. put together with concatenation if necessary).

Will do.

JakeQZ (approved):

PR #1394, opened 2025-11-20

oliverklee:


oliverklee:
I'd like to add https://github.com/phpstan/phpstan-deprecation-rules, and for the latest version of that tool, we'll need to update PHPStan first.

composer.json
```diff
@@ -30,15 +30,15 @@
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "1.4.0",
         "phpstan/extension-installer": "1.4.3",
-        "phpstan/phpstan": "1.12.32 || 2.1.32",
-        "phpstan/phpstan-phpunit": "1.4.2 || 2.0.8",
-        "phpstan/phpstan-strict-rules": "1.6.2 || 2.0.7",
+        "phpstan/phpstan": "1.12.33 || 2.1.454",
```
JakeQZ: The latest release is 2.1.45.


```suggestion
        "phpstan/phpstan": "1.12.33 || 2.1.45",
```

oliverklee: I blame fat fingers of mine. ;-)

oliverklee:
Now that we the current PHPStan doesn't lead to changes in the PHPStan baseline, we can let this sit for some more. Marking as draft.

Issue #1324, opened 2025-07-09
Refactor `DeclarationBlock::parse()`


JakeQZ:
Following #1292, this method will have become quite bloated, and may exceed a desired [cyclomatic complexity](https://en.wikipedia.org/wiki/Cyclomatic_complexity) limit.

It could be broken down into helper methods, e.g. `parseSelector` and [`parseSelectorList`](https://developer.mozilla.org/en-US/docs/Web/CSS/Selector_list).

JakeQZ:
Looked at this again.  I think it's as clean and clear as it can be.

The main work of the method is to parse the selector until it finds `{` or `}` not enclosed in quotes.  It then calls on to `parseRuleSet()` to complete the job.

Pre-splitting at commas is not viable for the reasons outlined in https://github.com/MyIntervals/PHP-CSS-Parser/pull/1398#pullrequestreview-3519678182.

That leaves the only refactoring possible being to remove one method call, add another method, and add a call to it.  Which doesn't really help.  Closing.

JakeQZ:
Reopening.  See https://github.com/MyIntervals/PHP-CSS-Parser/pull/1398#issuecomment-3590773005.

`setSelectors()` should include the parsing that `parse()` does, so as to allow selectors like `a[title="this, and that"]` to be passed.

PR #1398, opened 2025-11-28
[CLEANUP] Clean extra whitespace in css selector

8ctopus:
This PR cleans the css selector from extra whitespace:

```php
$input = <<<CSS
p    >
    small {
font-size:.7em
}

CSS;

$document = (new \Sabberworm\CSS\Parser($input))
    ->parse();

echo $document->render(\Sabberworm\CSS\OutputFormat::createPretty());
```

without the PR

```css
p    >
    small {
        font-size: .7em;
}
```

with the PR:

```css
p > small {
        font-size: .7em;
}
```

### DIFF
--- a/tests/Unit/Property/SelectorTest.php
+++ b/tests/Unit/Property/SelectorTest.php
@@ -171,9 +171,7 @@ final class SelectorTest extends TestCase
      */
     public function cleansUpNewLineWithinSelector(): void
     {
-        $selector = "p
->
-small";
+        $selector = "p\n>\nsmall";
 
         $subject = new Selector($selector);
 


JakeQZ (changes requested):
Thanks.  It's a nice idea to tidy up the whitespace in the selectors.

However, it's not as simple as it seems.  We would also need to avoid touching the whitespace within [attribute selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Attribute_selectors).  Although an unlikely real-world example, this change would break a selector like `a[title="extra  space"]`.

So I think this would require a more complex regex that handles quoted strings (single and double quotes) to avoid touching them.  I'm not sure how `a[title=extra  space]` is supposed to be interpreted; that would need checking up on.

If you're willing to give this a go, we would welcome the tidying functionality, with some tests added to `SelectorTest` to make sure it reformats only the whitespace it should.  I also understand if this now seems like too big a job - instead we could convert this PR to a 'would be nice to have' feature request.

JakeQZ:
> If you're willing to give this a go

`DeclarationBlock::parse()` may be the place to look for removing whitespace, though I'm not sure how.

I also note that `setSelectors` would fail for `a[title="this, and that"]`, though the parser would process it fine.

@8ctopus, whether or not you proceed with this PR, I thank you for helping us identify some gremlins :/

8ctopus:
@JakeQZ Give me a few days, I'll see what I can come up with as a solution.

JakeQZ:
> @JakeQZ Give me a few days, I'll see what I can come up with as a solution.

Thanks.  There are some comments in #1324 that may be useful.

8ctopus:
I've come up with a solution for you to look at when you have time. Unfortunately, not elegant, but all tests do pass. I couldn't figure out how to create a regex that only takes two or more spaces that are not within an attribute selector (negative look behind only supports the characters immediately beforehand, therefore not possible to use it).

tests/Unit/Property/SelectorTest.php
```diff
@@ -141,4 +141,29 @@ public function isValidForInvalidSelectorReturnsFalse(string $selector): void
     {
         self::assertFalse(Selector::isValid($selector));
     }
+
+    /**
+     * @test
+     */
+    public function cleanupSpacesWithinSelector(): void
+    {
+        $selector = <<<'CSS'
```
oliverklee: Please don't use heredoc - indented heredoc doesn't work with with PHP 7.2 (though it works with 7.3 upwards), and it makes autoformatting the code more of a hassle.

oliverklee: Instead, could you please switch to a literal, quotes-enclosed string?

src/Property/Selector.php
```diff
@@ -74,7 +75,9 @@ public function getSelector(): string
 
     public function setSelector(string $selector): void
     {
-        $this->selector = \trim($selector);
+        $selector = \trim($selector);
+
+        $this->selector = !str_contains($selector, '[') ? preg_replace('/\s{2,}/', ' ', $selector) : $selector;
```
JakeQZ: The `\s` needs double-escaping (it does technically work single-escaped, but fails a coding style check).

We could also convert any whitespace to an actual space - e.g. a single newline, tab, etc. - and have a test for that too.

And for better performance (and to avoid [catastrophic backtracking](https://www.regular-expressions.info/catastrophic.html) in the unlikely case of a selector string with a million spaces being passed), we can use the [possessive quantifier](https://www.regular-expressions.info/possessive.html) (`+`).


```suggestion
        $this->selector = !str_contains($selector, '[') ? preg_replace('/\\s++/', ' ', $selector) : $selector;
```

tests/Unit/Property/SelectorTest.php
```diff
@@ -141,4 +141,29 @@ public function isValidForInvalidSelectorReturnsFalse(string $selector): void
     {
         self::assertFalse(Selector::isValid($selector));
     }
+
+    /**
+     * @test
+     */
+    public function cleanupSpacesWithinSelector(): void
```
JakeQZ:
```suggestion
    public function cleansUpSpacesWithinSelector(): void
```

tests/Unit/Property/SelectorTest.php
```diff
@@ -141,4 +141,29 @@ public function isValidForInvalidSelectorReturnsFalse(string $selector): void
...
+
+        $subject = new Selector($selector);
+
+        self::assertSame('p > small', $subject->getSelector());
+    }
+
+    /**
+     * @test
+     */
+    public function doNotCleanupSpacesWithinAttributeSelector(): void
```
JakeQZ:
```suggestion
    public function doesNotCleanupSpacesWithinAttributeSelector(): void
```

tests/Unit/Property/SelectorTest.php
8ctopus: Thank you for bringing this to my attention, it's fixed.

8ctopus:
Thank you again for the valuable input, I've made all changes requested, please review when you have the time.

src/Property/Selector.php
```diff
@@ -74,7 +75,9 @@ public function getSelector(): string
 
     public function setSelector(string $selector): void
     {
-        $this->selector = \trim($selector);
+        $selector = \trim($selector);
+
+        $this->selector = !str_contains($selector, '[') ? preg_replace('/\\s++/', ' ', $selector) : $selector;
```
JakeQZ: `str_contains` requires PHP 8, whereas the library supports PHP from 7.2, so I think we'll need to use `strpos` instead (for now).  Also, we're using `\` prefix when calling a global function (for clarity).

EDIT: I just realized I originally got the logic the wrong way round in the suggestion.  So I think it would be good to use a local variable to make it clear what's happening.  Thinking more, a comment explaining why whitespace isn't removed when there's an attribute selector wouldn't go amiss.

```suggestion
        // Whitespace can't be adjusted within an attribute selector, as it would change its meaning
        $hasAttribute = \strpos($selector, '[') !== false;
        $this->selector = $hasAttribute ? $selector : preg_replace('/\\s++/', ' ', $selector);
```

tests/Unit/Property/SelectorTest.php
```diff
@@ -141,4 +141,53 @@ public function isValidForInvalidSelectorReturnsFalse(string $selector): void
...
+    }
+
+    /**
+     * @test
+     */
+    public function cleansUpNewLineWithinSelector(): void
+    {
+        $selector = "p
+>
+small";
```
JakeQZ: To improve readability of the code, it would be better to use `\n` in the string rather than an actual newline (so that we don't have contents of the string indented less than the main code).

```suggestion
        $selector = "p\n>\nsmall";
```

8ctopus:
All changes done, thank you for the feedback.

oliverklee:
@8ctopus Thanks! Now there's a merge conflict in the changelog file. Could you please rebase on `main` (rebase, please don't merge `main` into this branch), resolve the merge conflict and repush? Thanks!

8ctopus:
done

PR #1399, opened 2025-12-01
[BUGFIX] Parse calc split over multiple lines

scq:
`calc()` expressions that are multiline, where one line ends on an operator, currently fail to parse. E.g.

```
.btn {
  --foo: calc(
    var(--bar) +
      (var(--baz) + var(--qux)) * 2
  );
}
```

This is because the code is looking for a space character following the operator, and failing when it sees a newline instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@ Please also have a look at our
 ### Fixed
 
 - Use typesafe versions of PHP functions (#1379, #1380, #1382, #1383, #1384)
+- Fix parsing of `calc` expressions when a newline immediately precedes or
+  follows a + or - operator (#1399)
 
 ### Documentation
 
--- a/src/Value/CalcFunction.php
+++ b/src/Value/CalcFunction.php
@@ -8,6 +8,8 @@ use Sabberworm\CSS\Parsing\ParserState;
 use Sabberworm\CSS\Parsing\UnexpectedEOFException;
 use Sabberworm\CSS\Parsing\UnexpectedTokenException;
 
+use function Safe\preg_match;
+
 class CalcFunction extends CSSFunction
 {
     private const T_OPERAND = 1;
@@ -60,9 +62,8 @@ class CalcFunction extends CSSFunction
                 if (\in_array($parserState->peek(), $operators, true)) {
                     if (($parserState->comes('-') || $parserState->comes('+'))) {
                         if (
-                            $parserState->peek(1, -1) !== ' '
-                            || !($parserState->comes('- ')
-                                || $parserState->comes('+ '))
+                            preg_match('/\\s/', $parserState->peek(1, -1)) !== 1
+                            || preg_match('/\\s/', $parserState->peek(1, 1)) !== 1
                         ) {
                             throw new UnexpectedTokenException(
                                 " {$parserState->peek()} ",
--- a/tests/Unit/Value/CalcFunctionTest.php
+++ b/tests/Unit/Value/CalcFunctionTest.php
@@ -1,0 +1,206 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit\Value;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\OutputFormat;
+use Sabberworm\CSS\Parsing\ParserState;
+use Sabberworm\CSS\Parsing\UnexpectedTokenException;
+use Sabberworm\CSS\Settings;
+use Sabberworm\CSS\Value\CalcFunction;
+use Sabberworm\CSS\Value\CalcRuleValueList;
+use Sabberworm\CSS\Value\Size;
+
+/**
+ * @covers \Sabberworm\CSS\Value\CalcFunction
+ */
+final class CalcFunctionTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function parseSimpleCalc(): void
+    {
+        $css = 'calc(100% - 20px)';
+        $calcFunction = $this->parse($css);
+
+        self::assertInstanceOf(CalcFunction::class, $calcFunction);
+        self::assertSame('calc', $calcFunction->getName());
+
+        $args = $calcFunction->getArguments();
+        self::assertCount(1, $args);
+        self::assertInstanceOf(CalcRuleValueList::class, $args[0]);
+
+        /** @var CalcRuleValueList $value */
+        $value = $args[0];
+        $components = $value->getListComponents();
+        self::assertCount(3, $components); // 100%, -, 20px
+
+        self::assertInstanceOf(Size::class, $components[0]);
+        self::assertSame(100.0, $components[0]->getSize());
+        self::assertSame('%', $components[0]->getUnit());
+
+        self::assertSame('-', $components[1]);
+
+        self::assertInstanceOf(Size::class, $components[2]);
+        self::assertSame(20.0, $components[2]->getSize());
+        self::assertSame('px', $components[2]->getUnit());
+    }
+    /**
+     * @test
+     */
+    public function parseNestedCalc(): void
+    {
+        $css = 'calc(100% - calc(20px + 1em))';
+        $calcFunction = $this->parse($css);
+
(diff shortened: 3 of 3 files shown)


src/Value/CalcFunction.php
```diff
@@ -59,10 +59,10 @@ public static function parse(ParserState $parserState, bool $ignoreCase = false)
             } else {
                 if (\in_array($parserState->peek(), $operators, true)) {
                     if (($parserState->comes('-') || $parserState->comes('+'))) {
+                        $nextChar = $parserState->peek(1, 1);
                         if (
                             $parserState->peek(1, -1) !== ' '
-                            || !($parserState->comes('- ')
-                                || $parserState->comes('+ '))
+                            || !($nextChar === ' ' || $nextChar === "\n" || $nextChar === "\r")
```
JakeQZ: `preg_match('/\\s/', $nextChar) !== 1` would cover all possible whitespace (including vertical and horizontal tabs).

We're using the 'Safe' library, so `use function Safe\preg_match;` would need to be added.

src/Value/CalcFunction.php
```diff
@@ -59,10 +59,10 @@ public static function parse(ParserState $parserState, bool $ignoreCase = false)
             } else {
                 if (\in_array($parserState->peek(), $operators, true)) {
                     if (($parserState->comes('-') || $parserState->comes('+'))) {
+                        $nextChar = $parserState->peek(1, 1);
                         if (
                             $parserState->peek(1, -1) !== ' '
```
JakeQZ: As part of this change, could we also allow any type of whitespace before the operator?

tests/ParserTest.php
```diff
@@ -1237,4 +1237,14 @@ public function escapedSpecialCaseTokens(): void
         self::assertInstanceOf(URL::class, $urlRule->getValue());
         self::assertInstanceOf(CalcFunction::class, $calcRule->getValue());
     }
+
+    /**
+     * @test
+     */
+    public function multilineCalc()
```
JakeQZ: The static analysis checks want there to be a return type provided.

```suggestion
    public function multilineCalc(): void
```

JakeQZ (changes requested):
Hi @scq,

Thanks for this.  I concur that the spec allows any type of whitespace around the `+` and `-` operators (requiring at least one whitespace character).

To cover all types of whitespace, we can use `preg_match` with `\s` (via the 'Safe' library).

As part of this change, it would be good to also allow all types of whitespace before the operator (as well as after) - with tests covering that too.

We're aiming to move the tests away from functional tests to dedicated unit tests.  I note there is not yet a `CalcFunctionTest`.  Adding such a `TestCase` for the tests related to this change would be good.

Also, could you add an entry to the `CHANGELOG.md` file (in the root folder of the repository)?

src/Value/CalcFunction.php
scq: Ooh, good idea, switched that over.

src/Value/CalcFunction.php
scq: Good call, made that change.

tests/ParserTest.php
scq: Removed this test, replaced with a more in depth `CalcFunctionTest` unit test.

scq:
Sweet, thanks for the review! Made those changes.

JakeQZ:
> [Coverage Status](https://coveralls.io/builds/76745971)
> 
> coverage: 62.467% (+2.8%) from 59.658% when pulling **[3de58cd](https://github.com/MyIntervals/PHP-CSS-Parser/commit/3de58cd673027511f9a782514e5395d52aa50ed8) on totara:calc-parsing-fix** into **[fe29992](https://github.com/MyIntervals/PHP-CSS-Parser/commit/fe29992d5ef3c74e8c9927549cbf82072eea9d6c) on MyIntervals:main**.

That's impressive too :)

tests/Unit/Value/CalcFunctionTest.php
```diff
@@ -0,0 +1,206 @@
...
+        $calcFunction = $this->parse($css);
+
+        /** @var CalcRuleValueList $value */
+        $value = $calcFunction->getArguments()[0];
+        $components = $value->getListComponents();
+
+        self::assertCount(3, $components);
+        self::assertSame('-', $components[1]);
+
+        /** @var CalcFunction */
```
JakeQZ, line 66: Just spotted that this is missing a variable name, so is probably not needed.

scq: Ah oops, my bad! Want me to make a PR for that one?

JakeQZ (commented):
Found a minor issue...

tests/Unit/Value/CalcFunctionTest.php
JakeQZ: If it's easy, would be good to tidy that up.

I don't know if the other `@var`s are all needed - worth checking.

scq: Sweet, opened #1402. There were a couple that PHPStan needed, but checks passed without the others.

tests/Unit/Value/CalcFunctionTest.php
```diff
@@ -0,0 +1,206 @@
...
+        $this->expectException(UnexpectedTokenException::class);
+        $this->expectExceptionMessage('calc');
+        CalcFunction::parse($parserState);
+    }
+
+    /**
+     * Parse provided CSS as a CalcFunction
+     *
+     * @param string $css
+     * @return CalcFunction
```
JakeQZ, line 196: These PHPDoc annotations are actually redundant, since the types are in the method signature.

tests/Unit/Value/CalcFunctionTest.php
```diff
@@ -0,0 +1,206 @@
...
+        self::assertInstanceOf(Size::class, $components[1]); // 100%
+        self::assertSame('-', $components[2]);
+        self::assertInstanceOf(Size::class, $components[3]); // 20px
+        self::assertSame(')', $components[4]);
+        self::assertSame('*', $components[5]);
+        self::assertInstanceOf(Size::class, $components[6]); // 2
+    }
+
+    /**
+     * @return array<string, array{0: string, 1: string}>
```
JakeQZ, line 102: These can all be `non-empty-string` for tightness.

tests/Unit/Value/CalcFunctionTest.php
```diff
@@ -0,0 +1,206 @@
...
+     */
+    public function parseValidOperators(string $css, string $rendered): void
+    {
+        $calcFunction = $this->parse($css);
+        $output = $calcFunction->render(OutputFormat::create());
+        self::assertSame($rendered, $output);
+    }
+
+    /**
+     * @return array<string, array{0: string, 1: string}>
```
JakeQZ, line 129: And these.

tests/Unit/Value/CalcFunctionTest.php
```diff
@@ -0,0 +1,206 @@
...
+     */
+    public function parseMultiline(string $css, string $rendered): void
+    {
+        $calcFunction = $this->parse($css);
+        $output = $calcFunction->render(OutputFormat::create());
+        self::assertSame($rendered, $output);
+    }
+
+    /**
+     * @return array<string, array{0: string}>
```
JakeQZ, line 157: And these.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ Please also have a look at our
 ### Fixed
 
 - Use typesafe versions of PHP functions (#1379, #1380, #1382, #1383, #1384)
+- Fix parsing of `calc` expressions when a newline immediately precedes or
+  follows a + or - operator (#1399)
```
JakeQZ, line 23: I forgot to mention that we put the changelog entries in reverse chronological order (though I'm not sure why).

tests/Unit/Value/CalcFunctionTest.php
scq: Ah nice, opened #1405 for that and the `parse()` comment

PR #1400, opened 2025-12-01
[FEATURE] Parse container queries

scq:
Adds support for CSS container queries.

https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Containment/Container_queries

```
@container (width > 700px) {
  .card h2 {
    font-size: 2em;
  }
}
```

### DIFF
--- a/tests/Unit/Property/AtRuleTest.php
+++ b/tests/Unit/Property/AtRuleTest.php
@@ -1,0 +1,22 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit\Property;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Property\AtRule;
+
+final class AtRuleTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function blockRulesConstantIsCorrect(): void
+    {
+        self::assertEqualsCanonicalizing(
+            ['media', 'document', 'supports', 'region-style', 'font-feature-values', 'container'],
+            explode('/', AtRule::BLOCK_RULES)
+        );
+    }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Please also have a look at our
 
 ### Added
 
+- Support for CSS container queries (#1400)
+
 ### Changed
 
 ### Deprecated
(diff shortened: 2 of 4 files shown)


JakeQZ (changes requested):
Looks good.

Could you also add a test to confirm the feature - ideally introducing an `AtRuleTest` `TestCase`?  And add an entry to `CHANGELOG.md`?

Thanks.

scq:
Thanks! Added a test here, it's a little difficult to test being an interface so it just checks the value of the property, but open to any better ideas you have.

tests/Unit/Property/AtRuleTest.php
```diff
@@ -0,0 +1,25 @@
...
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit\Property;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Property\AtRule;
+
+/**
+ * @covers \Sabberworm\CSS\Property\AtRule
+ */
```
JakeQZ: PHPStan is complaining that `@covers` can't be applied to an interface, so I think the DocBlock will need removing.

tests/Unit/Property/AtRuleTest.php
```diff
@@ -0,0 +1,25 @@
...
+final class AtRuleTest extends TestCase
+{
+    /**
+     * @test
+     */
+    public function blockRulesConstantIsCorrect(): void
+    {
+        self::assertEqualsCanonicalizing(
```
JakeQZ, line 17: Good choice of assert method - it allows the components to be in any order.

JakeQZ (changes requested):
Thanks again.

> it's a little difficult to test being an interface so it just checks the value of the property

My apologies.  When looking at the diffs without full context, I was thinking that `AtRule` was a class.

The test you have added I think a is good to check to have for the internal behaviour (and fulfills the criteria: fails without the change; passes with).  Though it seems PHPStan does not like a `@covers` annotation for an interface.

> but open to any better ideas you have.

I've now done a bit of digging around the code base and existing tests and found that we have `tests/CSSList/AtRuleBlockListTest.php`.  This has a test method `parsesSyntacticallyCorrectAtRuleInStrictMode` which checks that the at-rule is not discarded.

So I think to test the external beviour, it would good to have the `provideSyntacticallyCorrectAtRule()` data provider also provide a valid `@contianer` rule, which I expect would fail without the change.  I think that would be sufficient (it's all we have for the other at-rules).

@oliverklee, do you have any other comments?

scq:
Aha! Thanks, good find --  I've added a `@container` entry to the `provideSyntacticallyCorrectAtRule` data provider.

PR #1403, opened 2025-12-02
[DOCS] Polish the changelog and add code coverage metadata

oliverklee:


JakeQZ:
> * Added `@covers` annotation to `AtRuleTest` for proper code coverage tracking

This was originally put in with the PR, but then removed because of the PHPStan warning.

I don't think `@covers` makes sense for an interface, because it has no code that can be covered.

PR #1404, opened 2025-12-02
[DOCS] Polish the changelog

oliverklee:


JakeQZ (approved):
I also noticed later that the operators should probably have backticks.

And that the new changelog entry should be above the previous one - https://github.com/MyIntervals/PHP-CSS-Parser/pull/1399#discussion_r2580608714.

PR #1406, opened 2025-12-04
[BUGFIX] Allow order-insensitve `removeDeclarationBlockBySelector`

### LINKED ISSUE #1330: `removeDeclarationBlockBySelector()` is unlikely to work in most cases
This is a method of `CSSList` (which `Document` extends).

If it is not passed an array of `Selector` objects (directly, or indirectly via a `DeclarationBlock`), then it will construct a new array of `Selector` objects for use in comparison using `==`, comparing the arrays.

AFAIK, any comparison involving objects, even if loose, requires the objects to be the same instances for a match.

The only current tests for it inovle passing an existing `DelcarationBlock`; there are no tests that pass selectors as strings.

This may be a candidate for removal.  @sabberworm, @oliverklee, WDYT?

JakeQZ:
Also internally avoid comparison with `==`.

Resolves #1330

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -429,18 +429,18 @@ abstract class CSSList implements CSSElement, CSSListItem, Positionable
     }
 
     /**
-     * @param list<Selector> $lhs
-     * @param list<Selector> $rhs
+     * @param list<Selector> $selectors1
+     * @param list<Selector> $selectors2
      */
-    private static function selectorsMatch(array $lhs, array $rhs): bool
+    private static function selectorsMatch(array $selectors1, array $selectors2): bool
     {
-        $lhsStrings = self::getSelectorStrings($lhs);
-        $rhsStrings = self::getSelectorStrings($rhs);
+        $selectorStrings1 = self::getSelectorStrings($selectors1);
+        $selectorStrings2 = self::getSelectorStrings($selectors2);
 
-        \sort($lhsStrings);
-        \sort($rhsStrings);
+        \sort($selectorStrings1);
+        \sort($selectorStrings2);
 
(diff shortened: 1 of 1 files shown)


src/CSSList/CSSList.php
```diff
@@ -427,4 +427,34 @@ public function getContents(): array
...
         return $this->contents;
     }
+
+    /**
+     * @param list<Selector> $lhs
+     * @param list<Selector> $rhs
+     */
+    private static function selectorsMatch(array $lhs, array $rhs): bool
+    {
+        $lhsStrings = self::getSelectorStrings($lhs);
```
oliverklee: I had to think for a moment about what "lhs" and "rhs" might mean. As the left/right side doesn't seem to be relevant to me, maybe we can use names like `$selectors1` and `$selectors2` instead. WDYT?

JakeQZ: I wasn't sure how to name them, as the names are pretty-much irrelevant.  I had thought about just `$a` and `$b` before going for mathematical convention for an equation.  Agree with `$selectors1` and `$selectors2`.  At least it says what they are FAOD.

oliverklee (approved):
LGTM.

src/CSSList/CSSList.php
JakeQZ: I wrote a reply here that seems to have been lost.

I had thought that just `$a` and `$b` would do, as the names have little significance, but finally opted for the mathematical terminology (LHS = left hand side), as it's an equation.  Agree that `$selectors1` etc. is better: that, at least, FAOD, conveys what the parameter is.

JakeQZ: My original reply has suddenly reappeared.  Got to love GitHub.

PR #1411, opened 2025-12-04
[TASK] Avoid heredoc in another test

oliverklee:
This allows the testcase being autoformatted
without breaking the tests for PHP 7.2.

https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc


oliverklee:
If I change the indentation, some tests will fail. I'll try concatenated strings next.

JakeQZ (approved):
Some of the tests are unnecessarily whitespace-dependent :/

Issue #1412, opened 2025-12-06
Remove 'heredoc'


JakeQZ:
'[heredoc](https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.heredoc)' is a hangover from when PHP was just a simple scripting language.

Now it is a fully-fledged OO language, the syntax breaks readability in classes, etc., and should not be used (IOO).

- [ ] Replace all 'heredoc'
- [ ] Add code sniffer rule (or suchlike) to not allow it

oliverklee:
I have checked the PHP_CodeSniffer and PHP-CS-Fixer rules and was unable to find any rules for this. Maybe we need to write a small Composer script to grep for the `<<<` heredoc marker.

JakeQZ:
> I have checked the PHP_CodeSniffer and PHP-CS-Fixer rules and was unable to find any rules for this. Maybe we need to write a small Composer script to grep for the `<<<` heredoc marker.

I found [HeredocSniff](https://github.com/squizlabs/PHP_CodeSniffer/blob/master/src/Standards/Squiz/Sniffs/PHP/HeredocSniff.php) in the current codebase.  Looks like it would do the job.  Doesn't seem to be documented, though I've established that the rule is `Squiz.PHP.Heredoc`, and it works.  And that we don't have any remaining heredoc (or nowdoc) to replace.

Would be nice to also have a rule to disallow multiline strings that have content indented less than the code...

JakeQZ:
> I've established that the rule is `Squiz.PHP.Heredoc`, and it works.

#1415

JakeQZ:
[Two out of three ain't bad](https://tidal.com/track/61126692/u).  There is no sniffer or fixer rule I'm aware of or can find to deal with multiline strings, and I'm not planning on writing one at this time.  Thus closing, as there's nothing more that can be done here now.

PR #1415, opened 2025-12-08
[DEVELOPER] Disallow 'heredoc' and 'nowdoc' for strings

JakeQZ:
Part of #1412.

### DIFF
--- a/config/phpcs.xml
+++ b/config/phpcs.xml
@@ -9,4 +9,5 @@
       <property name="absoluteLineLimit" value="120"/>
     </properties>
   </rule>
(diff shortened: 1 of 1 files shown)


oliverklee:
Would you be willing to port this to Emogrifier as well?

JakeQZ:
> Would you be willing to port this to Emogrifier as well?

https://github.com/MyIntervals/emogrifier/pull/1534

Wondering if there's a way of sharing the config file between projects...

oliverklee:
> Wondering if there's a way of sharing the config file between projects...

We'd probably need to set up (and maintain) a third project with the config files, and then add that package as a development dependency.

PR #1416, opened 2025-12-09
[CLEANUP] Extract method `DeclarationBlock::parseSelectors`

JakeQZ:
This now does the selector-parsing that `parse()` used to do itself.

Should help with #1324 and
https://github.com/MyIntervals/PHP-CSS-Parser/pull/1398#issuecomment-3590773005

### DIFF
--- a/src/RuleSet/DeclarationBlock.php
+++ b/src/RuleSet/DeclarationBlock.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Sabberworm\CSS\RuleSet;
 
+use Sabberworm\CSS\Comment\Comment;
 use Sabberworm\CSS\Comment\CommentContainer;
 use Sabberworm\CSS\CSSElement;
 use Sabberworm\CSS\CSSList\CSSList;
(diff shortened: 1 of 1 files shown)


src/RuleSet/DeclarationBlock.php
```diff
@@ -303,4 +253,71 @@ public function render(OutputFormat $outputFormat): string
 
         return $result;
     }
+
+    /**
+     * @param array<int, Comment> $comments
```
JakeQZ, line 258: I couldn't get Stan to be happy with `list<Comment>`...

oliverklee: My best guess is that that's because `ParserState::consumeUntil` has `$comments` annotated as `@param array<int, Comment> $comments` as well, and it's passed by reference.

We probably can change both type annotations to `list` (in a separate PR).

oliverklee (approved):
:broom:

src/RuleSet/DeclarationBlock.php
JakeQZ: Thanks.  That is indeed the case.

> We probably can change both type annotations to `list` (in a separate PR).

#1418.

PR #1418, opened 2025-12-09
[CLEANUP] Tighten `consumeUntil` `$comments` parameter type

JakeQZ:


JakeQZ:
Seems that `DeclarationBlock::parseSelectors()` is the only place where the `$comments` parameter of `comsumeUntil()` is used.

PR #1419, opened 2025-12-09
[BUGFIX] Allow commas in attributes in `setSelectors`

JakeQZ:
This fixes an issue noted in #1324.

### DIFF
--- a/src/RuleSet/DeclarationBlock.php
+++ b/src/RuleSet/DeclarationBlock.php
@@ -101,11 +101,22 @@ class DeclarationBlock implements CSSElement, CSSListItem, Positionable, RuleCon
         } else {
             // A string of comma-separated selectors requires parsing.
             // Parse as if it's the opening part of a rule.
-            $parserState = new ParserState($selectors . '{', Settings::create());
-            $selectorsToSet = self::parseSelectors($parserState);
-            $parserState->consume('{'); // throw exception if this is not next
-            if (!$parserState->isEnd()) {
-                throw new UnexpectedTokenException('EOF', $parserState->peek(5));
+            try {
+                $parserState = new ParserState($selectors . '{', Settings::create());
+                $selectorsToSet = self::parseSelectors($parserState);
+                $parserState->consume('{'); // throw exception if this is not next
+                if (!$parserState->isEnd()) {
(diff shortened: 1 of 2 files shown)


src/RuleSet/DeclarationBlock.php
```diff
@@ -98,7 +99,14 @@ public function setSelectors($selectors, ?CSSList $list = null): void
...
             $selectorsToSet = $selectors;
         } else {
-            $selectorsToSet = \explode(',', $selectors);
+            // A string of comma-separated selectors requires parsing.
+            // Parse as if it's the opening part of a rule.
+            $parserState = new ParserState($selectors . '{', Settings::create());
+            $selectorsToSet = self::parseSelectors($parserState);
+            $parserState->consume('{'); // throw exception if this is not next
+            if (!$parserState->isEnd()) {
+                throw new UnexpectedTokenException('EOF', $parserState->peek(5));
```
oliverklee: If possible, we also should pass the line number to the `UnexpectedTokenException`.

tests/Unit/RuleSet/DeclarationBlockTest.php
```diff
@@ -293,4 +294,58 @@ public function setSelectorsIgnoresKeys(): void
...
+    /**
+     * @test
+     *
+     * @param non-empty-string $selector
+     *
+     * @dataProvider provideInvalidSelector
+     */
+    public function setSelectorsThrowsExceptionWithInvalidSelector(string $selector): void
+    {
+        $this->expectException(UnexpectedTokenException::class);
```
oliverklee, line 359: We should also check for the exception message.

src/RuleSet/DeclarationBlock.php
JakeQZ: No line number is available, as this is a manipulation method.  The other `throw`s here don't pass a line number.

tests/Unit/RuleSet/DeclarationBlockTest.php
JakeQZ: I looked at the messages and found they were unhelpful, so have fixed that.

I also found that the exceptions thrown by the code when `{` or end of string were not as expected were not being tested, so have added more test data for that.

PR #1420, opened 2025-12-11
[CLEANUP] Extract method `DeclarationBlock::parseSelector`

### LINKED ISSUE #1324: Refactor `DeclarationBlock::parse()`

JakeQZ:
Closes #1324

### DIFF
--- a/src/RuleSet/DeclarationBlock.php
+++ b/src/RuleSet/DeclarationBlock.php
@@ -336,6 +336,7 @@ class DeclarationBlock implements CSSElement, CSSListItem, Positionable, RuleCon
                     }
                     break;
                 case '{':
+                    // The fallthrough is intentional.
                 case '}':
                     if (!\is_string($stringWrapperCharacter)) {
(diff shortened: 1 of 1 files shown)


src/RuleSet/DeclarationBlock.php
```diff
@@ -321,22 +335,24 @@ private static function parseSelectors(ParserState $parserState, array &$comment
                         --$functionNestingLevel;
                     }
                     break;
+                case '{':
```
oliverklee, line 338:
```suggestion
                case '{':
                    // The fallthrough is intentional.
```

PR #1424, opened 2025-12-12
[BUGFIX] Parse comment(s) immediately preceding selector (part 2)

JakeQZ:


src/RuleSet/DeclarationBlock.php
```diff
@@ -348,12 +347,18 @@ private static function parseSelector(ParserState $parserState, array &$comments
...
         }
 
         if ($functionNestingLevel !== 0) {
             throw new UnexpectedTokenException(')', $nextCharacter);
         }
 
-        return \implode('', $selectorParts);
+        $selector = \implode('', $selectorParts);
+        if ($selector === '') {
+            throw new UnexpectedTokenException('selector', $nextCharacter, 'literal', $parserState->currentLine());
```
oliverklee, line 359: Is there a way for us to cover this exception with a test?

JakeQZ: It's sort-of covered by `Sabberworm\CSS\Tests\Functional\RuleSet\DeclarationBlock::parseReturnsNullForInvalidDeclarationBlock` - in the sense that that test fails without this exception throwing.  I'll have another look...

JakeQZ: I've added some more tests for rejecting invalid selectors.

I've also made a minor change to reject selectors comprising only whitespace, as some of the tests I added would fail otherwise.  It may be possible to apply that change as another pre-PR, but it's complicated by the fact that without the main fix (to not consume the first selector character without inspection), various tests would fail...

JakeQZ: > I've also made a minor change to reject selectors comprising only whitespace ... It may be possible to apply that change as another pre-PR

#1433

oliverklee:
I've merged the other PR. This PR now needs a rebase.

PR #1427, opened 2025-12-14
[TASK] Guard against infinite loop in `parseList`

JakeQZ:
This can't be tested because there are currently no cases that would fail the test without the change.

It is there as a preventative double-lock measure to make sure any future code changes do not introduce an infinite loop.

It's a slight unoptimization,
as it adds what should be a redundant runtime check, but at miniscule performance cost versus the value of preventing a waste of CPU power and memory in the case that something goes wrong.


JakeQZ:
Longer term it would probably be good to refactor the logic in `parseList`.  This is a sticking-plaster for now.

PR #1429, opened 2025-12-14
[CLEANUP] Avoid multiline strings in `OutputFormatTest`

oliverklee:
Also avoid literal tabs.

### DIFF
--- a/tests/OutputFormatTest.php
+++ b/tests/OutputFormatTest.php
@@ -51,8 +51,8 @@ final class OutputFormatTest extends TestCase
     public function plain(): void
     {
         self::assertSame(
-            '.main, .test {font: italic normal bold 16px/1.2 "Helvetica",Verdana,sans-serif;background: white;}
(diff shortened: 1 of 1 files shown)


tests/OutputFormatTest.php
```diff
@@ -51,8 +51,8 @@ protected function setUp(): void
     public function plain(): void
     {
         self::assertSame(
-            '.main, .test {font: italic normal bold 16px/1.2 "Helvetica",Verdana,sans-serif;background: white;}
-@media screen {.main {background-size: 100% 100%;font-size: 1.3em;background-color: #fff;}}',
+            ".main, .test {font: italic normal bold 16px/1.2 \"Helvetica\",Verdana,sans-serif;background: white;}\n"
```
JakeQZ, line 54: It's debatable whether such a string should be single-quoted, with `"\n"` concatenated.  Either way is fine I think.

PR #1437, opened 2025-12-18
[FEATURE] Parse simple expressions

JakeQZ:
Part of #389

### DIFF
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -49,6 +49,12 @@ parameters:
 			path: ../src/Value/Color.php
 
 		-
+			message: '#^Parameter \#2 \$arguments of class Sabberworm\\CSS\\Value\\Expression constructor expects array\<Sabberworm\\CSS\\Value\\Value\|string\>\|Sabberworm\\CSS\\Value\\RuleValueList, Sabberworm\\CSS\\Value\\Value\|string given\.$#'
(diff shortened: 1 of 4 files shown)


JakeQZ:
I can't find any specification reference that indicates this should be allowed or possible, though I'm sure I've seen it before.

Unless anyone can find it, I suggest this PR be closed as invalid.

oliverklee:
> I can't find any specification reference that indicates this should be allowed or possible, though I'm sure I've seen it before.
>
> Unless anyone can find it, I suggest this PR be closed as invalid.

I also can only find the [`calc()` syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc).

I agree that we should close this as invalid.

Closing it now.

JakeQZ:
> I also can only find the [`calc()` syntax](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/calc).

I just put the test CSS into Firefox.  If it's fixed so that the values are well formed (with a length and a unit), the expressions do work inside the `max` function, but not as standalones.

So there is something to support - expressions within CSS function values.  I don't know if all functions support expressions - or just specific length calculation functions like `max`...

JakeQZ:
> So there is something to support - expressions within CSS function values. I don't know if all functions support expressions - or just specific length calculation functions like `max`...

#1438 captures this.

Issue #1443, opened 2025-12-25
escaping data:image when not required


8ctopus:
I've noticed that the following CSS gets escaped when it doesn't need to:

```php
$input = <<<CSS
a {
  background-image: url("data:image/svg+xml,%3Csvg width='24' height='24' viewBox='0 0 24 24' fill='none' xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd' d='M14.3145 2 3V11.9987H17.5687L18 8.20761H14.3145L14.32 6.31012C14.32 5.32134 14.4207 4.79153 15.9426 4.79153H17.977V1H14.7223C10.8129 1 9.43687 2.83909 9.43687 5.93187V8.20804H7V11.9991H9.43687V23H14.3145Z' fill='black'/%3E%3C/svg%3E%0A");
}

CSS;

$document = (new \Sabberworm\CSS\Parser($input))
    ->parse();

$format = OutputFormat::createPretty();

echo $document->render($format);
```

result:

```css
a {
        background-image: url("data:image/svg+xml,%3Csvg width=\'24\' height=\'24\' viewBox=\'0 0 24 24\' fill
=\'none\' xmlns=\'http://www.w3.org/2000/svg\'%3E%3Cpath fill-rule=\'evenodd\' clip-rule=\'evenodd\' d=\'M14.3
145 2 3V11.9987H17.5687L18 8.20761H14.3145L14.32 6.31012C14.32 5.32134 14.4207 4.79153 15.9426 4.79153H17.977V
1H14.7223C10.8129 1 9.43687 2.83909 9.43687 5.93187V8.20804H7V11.9991H9.43687V23H14.3145Z\' fill=\'black\'/%3E
%3C/svg%3E%0A");
}
```

This happens in the following code:

https://github.com/MyIntervals/PHP-CSS-Parser/blob/0ae1fde22bc2f66363ebdd77095e6f1d35118bc8/src/Value/CSSString.php#L91-L96

Would it be possible to check the quoting type and only escape when actually required?

PR #1444, opened 2025-12-25
[BUGFIX] Do not escape characters that do not need escaping in CSS string

8ctopus:
Add draft to attempt to resolve https://github.com/MyIntervals/PHP-CSS-Parser/issues/1443

There are no specific tests yet, and there are improvements that can be made, I just want to get your approval before I polish things up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -26,6 +26,7 @@ Please also have a look at our
 
 ### Fixed
 
+- Do not escape characters that do not need escaping in CSS string (#1444)
 - Reject selector comprising only whitespace (#1433)
 - Improve recovery parsing when a rogue `}` is encountered (#1425, #1426)
 - Parse comment(s) immediately preceding a selector (#1421, #1424)
@@ -38,12 +39,6 @@ Please also have a look at our
 - Use typesafe versions of PHP functions (#1379, #1380, #1382, #1383, #1384)
 
 ### Documentation
-
-## 9.2.0:
-
-### Fixed
-
-- Do not escape characters that do not need escaping in CSS string (#1444)
 
 ## 9.1.0: Add support for PHP 8.5
 


JakeQZ:
> Wouldn't it be simpler (and more efficient) to use `addslashes()` as before, then do a `str_replace()` to replace `\'` with `'` or `\"` with `"` depending on the string quoting type?

PS. I agree with adding a separate method to do this.

8ctopus:
Updated according to your feedback.

oliverklee:
@8ctopus Thanks! Could you please rebase?

8ctopus:
done

oliverklee:
Thanks! GitHub is still showing merge conflicts. Could you please have a look?

8ctopus:
now should be good.

oliverklee:
Thanks! Now GitHub was able to run the CI jobs.

The PHP-CS-Fixer job suggests some improvements. Could you please apply those to get the build green? Thanks!

8ctopus:
This time all tests are green, I also added a test to check that we do not escape when not necessary.

oliverklee:
Thanks!

tests/fixtures/unicode.css
```diff
@@ -5,7 +5,7 @@
 .test-6 { content: "\00A5" } /* Same as "¥" */
 .test-7 { content: '\a' } /* Same as "\A" (Newline) */
 .test-8 { content: "\"\22" } /* Same as "\"\"" */
-.test-9 { content: "\"\27" } /* Same as ""\"\'"" */
+.test-9 { content: "\"\27" } /* Same as ""\"'"" */
```
JakeQZ: There seems to be an extra double-quote around the string in the comment, which was previously incorrect.

As we're changing this, we should probably fix that too:

```suggestion
.test-9 { content: "\"\27" } /* Same as "\"'" */
```

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,26 @@ public function getArrayRepresentationIncludesContents(): void
...
+    /**
+     * @test
+     */
+    public function doNotEscapeCharactersThatDoNotNeedToBeEscaped(): void
+    {
+        $input = "data:image/svg+xml,%3Csvg width='24' height='24' viewBox='0 0 24 24' fill='none'" .
+        " xmlns='http://www.w3.org/2000/svg'%3E%3Cpath fill-rule='evenodd' clip-rule='evenodd'" .
+        " d='M14.3145 2 3V11.9987H17.5687L18 8.20761H14.3145L14.32 6.31012C14.32 5.32134 14.4207" .
+        ' 4.79153 15.9426 4.79153H17.977V1H14.7223C10.8129 1 9.43687 2.83909 9.43687 ' .
+        " 5.93187V8.20804H7V11.9991H9.43687V23H14.3145Z' fill='black'/%3E%3C/svg%3E%0A";
```
JakeQZ: Autoformatting may apply some indentation, but I suggest leaving it as is for now.  @oliverklee can run the autoformatter he uses as a post-PR (I think it's part of PHPStorm, which I don't have, and even if you have it, you may not have the same configuration set-up).

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,26 @@ public function getArrayRepresentationIncludesContents(): void
...
+
+        $outputFormat = OutputFormat::createPretty();
+
+        self::assertSame("\"{$input}\"", (new CSSString($input))->render($outputFormat));
```
JakeQZ, line 125: I think we are generally favouring using string contactenation rather than string interpolation, but I'm not sure of the rationale.  All I can gather from [some threads on StackOverflow](https://stackoverflow.com/questions/15985675/is-there-any-reason-for-not-using-string-interpolation-in-php) is that it is a matter of personal preference.  @oliverklee, are you able to expand on this?

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,26 @@ public function getArrayRepresentationIncludesContents(): void
...
+
+        $outputFormat->setStringQuotingType("'");
+
+        $expected = str_replace("'", "\\'", $input);
+
+        self::assertSame("'{$expected}'", (new CSSString($input))->render($outputFormat));
```
JakeQZ, line 131: Ditto.

src/Value/CSSString.php
```diff
@@ -111,4 +111,13 @@ public function getArrayRepresentation(): array
             'contents' => $this->string,
         ];
     }
+
+    private function escape(string $str, string $quote): string
```
JakeQZ: I think we would prefer a full word (`$string`) as the parameter name rather than an abbreviation (`$str`).

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,26 @@ public function getArrayRepresentationIncludesContents(): void
         self::assertArrayHasKey('contents', $result);
         self::assertSame($contents, $result['contents']);
     }
+
+    /**
+     * @test
+     */
+    public function doNotEscapeCharactersThatDoNotNeedToBeEscaped(): void
```
JakeQZ: Could we have the test name in the present rather than imperative tense?

```suggestion
    public function doesNotEscapeCharactersThatDoNotNeedToBeEscaped(): void
```

8ctopus: done

src/Value/CSSString.php
8ctopus: done

tests/fixtures/unicode.css
8ctopus: done

8ctopus:
> However, looking at the tests, there don't seem to be any covering that double-quotes within a single-quoted string are no longer escaped, e.g. if string quoting type is ', the quotes in "Hello World" don't need escaping and the result should be '"Hello World"'. Could you add a test for that?

done

tests/Unit/Value/CSSStringTest.php
8ctopus: If you regex the project with `/\{\$.*\}/` you will see that interpolation is used 19 times, 4 times being my newly added code.

oliverklee:
> I'm not sure whether we have a project preference for string concatenation vs interpolation,

For new/updated code, we prefer string concatenation over interpolation (without having a hard-and-fast rule, though).

src/Value/CSSString.php
```diff
@@ -111,4 +111,13 @@ public function getArrayRepresentation(): array
             'contents' => $this->string,
         ];
     }
+
+    private function escape(string $string, string $quote): string
+    {
+        $string = \addslashes($string);
```
oliverklee: Instead of first escaping unnecessarily and then undoing part of the work, I'd prefer to only escape what we need to escape in the first place, i.e., directly use `str_replace` and avoid `addslashes`.

tests/Unit/Value/CSSStringTest.php
oliverklee: Yes, for new code, please use string concatenation instead of interpolation.

src/Value/CSSString.php
JakeQZ: I suggested the current approach in my original review - see https://github.com/MyIntervals/PHP-CSS-Parser/pull/1444#pullrequestreview-3616544899

oliverklee: Hm … couldn't we use just the `str_replace` and skip the `addslashes` (without using a loop)?

oliverklee: … or use [`addcslashes`](https://www.php.net/manual/en/function.addcslashes.php) and provide the characters to escape?

JakeQZ: We could probably use `addcslashes` to escape the backslash and quote.

The only other character that `addslashes` escapes is the null character (U+0000), which I don't think we need to worry about (there are no tests for this, it is highly unlikely to be used in the real world, and it was probably just an inconsequential side-effect of `addslashes` that that would be the behaviour - other control characters aren't handled, apart from newline which is currently `str_replace`d separately by the calling method).

Perhaps we could move the `str_replace` for newline to the new method as part of this change.

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,44 @@ public function getArrayRepresentationIncludesContents(): void
...
+     */
+    public function doesNotEscapeCharactersThatDoNotNeedToBeEscaped2(): void
+    {
+        $outputFormat = OutputFormat::createPretty();
+
+        $input = "'Hello World'";
+
+        self::assertSame("\"{$input}\"", (new CSSString($input))->render($outputFormat));
+
+        $input = '"Hello World"';
```
JakeQZ: The first test checks that single quote (`'`) is escaped when it needs to be but not when it doesn't.

I'd suggest the second test does likewise for double quote (`"`), i.e. focuses only on the double quote and includes confirmation that it is still escaped when it needs to be:

```suggestion
        $input = '"Hello World"';
        
        $outputFormat = OutputFormat::createPretty();

        self::assertSame('"\\"Hello World\\""', (new CSSString($input))->render($outputFormat));

```

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,44 @@ public function getArrayRepresentationIncludesContents(): void
...
+
+        $expected = str_replace("'", "\\'", $input);
+
+        self::assertSame("'{$expected}'", (new CSSString($input))->render($outputFormat));
+    }
+
+    /**
+     * @test
+     */
+    public function doesNotEscapeCharactersThatDoNotNeedToBeEscaped2(): void
```
JakeQZ: Now there are two test methods, rather than appending the number `2` to the name of the second, they could be renamed to indicate their character speciality:


```suggestion
    public function doesNotEscapeDoubleQuotesThatDoNotNeedToBeEscaped(): void
```

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,44 @@ public function getArrayRepresentationIncludesContents(): void
         self::assertArrayHasKey('contents', $result);
         self::assertSame($contents, $result['contents']);
     }
+
+    /**
+     * @test
+     */
+    public function doesNotEscapeCharactersThatDoNotNeedToBeEscaped(): void
```
JakeQZ: ...and


```suggestion
    public function doesNotEscapeSingleQuotesThatDoNotNeedToBeEscaped(): void
```

JakeQZ (changes requested):
> > looking at the tests, there don't seem to be any covering that double-quotes within a single-quoted string are no longer escaped
> 
> done

Thanks for your perseverence <3

I've made a suggestion for what the additional test should cover (to be consistent with the first test for single quotes), and for the test method naming.  (The comments appear in order of writing in the Conversation tab, and should be read in that order, which is not in the order of the code.  Hope that makes sense and they make sense.)

The code change suggested includes avoiding string interpolation in the one instance covered, though that would still need to resolved in the other three instances as per @oliverklee's comments.

tests/Unit/Value/CSSStringTest.php
JakeQZ: > Yes, for new code, please use string concatenation instead of interpolation.

I've opened #1464, because I'm not sure that's best.

Apologies for the conflicting reviewer requests.

I'd be happy to merge with string interpolation for now.  We can revisit this later pending the outcome of that discussion.

@oliverklee, would you be happy to merge this with the strings as are, pending further revisitation?

oliverklee: Yes, I'd be fine to keep the string interpolation when we merge this PR. I'd still like to get rid of the addslash-first-and-then-partially-undo part, though.

src/Value/CSSString.php
oliverklee: Done.

oliverklee (approved):

tests/Unit/Value/CSSStringTest.php
```diff
@@ -107,4 +108,43 @@ public function getArrayRepresentationIncludesContents(): void
...
+    public function doesNotEscapeDoubleQuotesThatDoNotNeedToBeEscaped(): void
+    {
+        $input = '"Hello World"';
+
+        $outputFormat = OutputFormat::createPretty();
+
+        self::assertSame('"\\"Hello World\\""', (new CSSString($input))->render($outputFormat));
+
+
+        $outputFormat->setStringQuotingType("'");
```
JakeQZ, line 145:
```suggestion
        self::assertSame('"\\"Hello World\\""', (new CSSString($input))->render($outputFormat));

        $outputFormat->setStringQuotingType("'");
```

src/Value/CSSString.php
```diff
@@ -93,7 +93,7 @@ public function getString(): string
      */
     public function render(OutputFormat $outputFormat): string
     {
-        $string = \addslashes($this->string);
+        $string = $this->escape($this->string, $outputFormat->getStringQuotingType());
         $string = \str_replace("\n", '\\A', $string);
```
JakeQZ, line 97: We should move this to the new `escape` method as a post-PR.  Or maybe it can be done as part of this PR, as I've just spotted an issue with the changelog that also needs resolving.

CHANGELOG.md
```diff
@@ -39,6 +39,12 @@ Please also have a look at our
 
 ### Documentation
 
+## 9.2.0:
+
+### Fixed
+
+- Do not escape characters that do not need escaping in CSS string (#1444)
+
```
JakeQZ: I've just spotted that this is in the wrong place (it should be added at the top of the 'Fixed' section above), and that the '9.2.0' heading shouldn't be created yet - we do that when we actually cut the release.

oliverklee: Done.

src/Value/CSSString.php
oliverklee: I'd prefer to have this as a post-PR cleanup as I still don't fully understand the `\\A` thing.

oliverklee: #1466

JakeQZ (approved):

PR #1447, opened 2025-12-31
[TASK] Add `getArrayRepresentation()` to `Renderable`

JakeQZ:
This method will make it easier to test the parsing independently of the rendering.

Placeholder implementations are currently all stubs that will throw an exception, confirmed by tests.
They can be implemented as and when required, with the tests updated.

Part of #1440.

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -436,6 +436,16 @@ abstract class CSSList implements CSSElement, CSSListItem, Positionable
     }
 
     /**
+     * @internal
+     *
+     * @return array<string, bool|int|float|string|list<array<string, mixed>>>
+     */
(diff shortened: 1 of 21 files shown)


oliverklee:
I'd also be fine with doing that in a follow-up PR.

JakeQZ:
> We also should have the stub implementation (plus tests) in the concrete subclasses of the (usually) abstract classes as the real implementations usually will be different for each class:

I'm not sure that's necessary.  We can just add the real implementations as and when they are needed.  If a child class ends up inheriting a parent implementation, but should have a different implementation (probably extended with additional properties), it will be obvious when we come to use it in testing that the child class method needs to be implemented (and tested) first, since it will not return (all) the properties we want to check.

PR #1451, opened 2026-01-01
[TASK] Add trait for `getArrayRepresentation` to populate `class`

JakeQZ:
Part of #1440.


oliverklee:
Sorry, I missed that this is the PR to review and that #1452 is a draft PR. I'll copy my comments over.

src/ArrayRepresentationClassProvider.php
```diff
@@ -0,0 +1,31 @@
...
+ * @phpstan-require-implements Renderable
+ */
+trait ArrayRepresentationClassProvider
+{
+    /**
+     * @return array<string, bool|int|float|string|list<array<string, mixed>>>
+     *
+     * @internal
+     */
+    public function getArrayRepresentation(): array
```
oliverklee: I'd strongly prefer the trait to not have an implementation of `getArrayRepresentation` as this can lead to classes with an incomplete implementation if they implement the trait. (Also, I'd like to avoid `parent::` calls.)

Instead, I suggest having a private method `getShortClassName(): non-empty-string` in the trait. Then we can also build the array from scratch in `getArrayRepresentation` in the implementing classes, allowing for more concise code.

src/ArrayRepresentationClassProvider.php
```diff
@@ -0,0 +1,31 @@
...
+    {
+        $reflect = new \ReflectionClass($this);
+
+        return ['class' => $reflect->getShortName()];
```
oliverklee: We could also inline `$reflect`. (I'm fine either way and slightly prefer the inlined version.)

oliverklee (changes requested):

src/ArrayRepresentationClassProvider.php
JakeQZ: > Instead, I suggest having a private method `getShortClassName(): non-empty-string` in the trait.

Changed as suggested.  Also renamed the trait accordingly.

> (Also, I'd like to avoid `parent::` calls.)
> Then we can also build the array from scratch in `getArrayRepresentation` in the implementing classes, allowing for more concise code.

In cases where a subclass includes additional properties and the parent also has some properties, it would be more concise not to duplicate filling in the parent's properties in the subclass's implementation.

src/ArrayRepresentationClassProvider.php
JakeQZ: Inlined.

PR #1462, opened 2026-01-05
[TASK] Implement `URL::getArrayRepresentation()`

oliverklee:
Part of #1440.

### DIFF
--- a/src/Value/URL.php
+++ b/src/Value/URL.php
@@ -93,7 +93,7 @@ class URL extends PrimitiveValue
     {
         return [
             'class' => $this->getShortClassName(),
-            // We're using the term "uri" here to terminology the wording used in the specs:
+            // We're using the term "uri" here to match the wording used in the specs:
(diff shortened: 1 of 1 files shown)


src/Value/URL.php
```diff
@@ -80,4 +83,19 @@ public function render(OutputFormat $outputFormat): string
...
+    /**
+     * @return array<string, bool|int|float|string|array<mixed>|null>
+     *
+     * @internal
+     */
+    public function getArrayRepresentation(): array
+    {
+        return [
+            'class' => $this->getShortClassName(),
+            // We're using the term "uri" here to terminology the wording used in the specs:
```
JakeQZ: This needs something changing so it is readable.

```suggestion
            // We're using the term "uri" here to match the wording used in the specs:
```

oliverklee:
> Code changes are fine, but there's a comment that needs fixing.

Thanks! It looks like I changed the sentence and then forgot to read it again.

PR #1470, opened 2026-01-22
[TASK] Use `Selector::parse()`

JakeQZ:
This replaces `DeclarationBlock::parseSelector()`.

Part of #1325.

Issue #1325, opened 2025-07-10
Parse selectors into component classes


JakeQZ:
See https://github.com/MyIntervals/PHP-CSS-Parser/pull/138#issuecomment-418193262

Adding this to the wishlist.

JakeQZ:
Reference: https://www.w3.org/TR/selectors-4/#grammar

Some thoughts:
- `Selector` should have its own `parse()` method (rather than `DeclarationBlock::parseSelector()`) and generally take more responsibility for itself;
- `Selector` currently represents `complex-selector` according to the W3C grammar; I see no reason to change this;
- The selector should be internally represented as `CompoundSelector [Combinator, CompoundSelector]*` where `Combinator` also includes 'descendent combinator' (space - note the reference above is regarding syntax rather than representation) -
  - `Combinator` could simply be a string (`>`, `+`, `~` or ` `) and may not need its own class;
- We do not need to initially create any API to manipulate the internal representation; it would be a nice-to-have for the future;
- At a later date we can further break down `CompoundSelector` if desired.

sabberworm:
I have this very stale branch https://github.com/MyIntervals/PHP-CSS-Parser/tree/selector-parsing that I once used to start working on this, just FYI.

JakeQZ:
The first phase of this is now complete.  Selectors can now be rendered with the combinators minified, or more humanly readable (#1504).
  
> * `Selector` should have its own `parse()` method (rather than `DeclarationBlock::parseSelector()`) and generally take more responsibility for itself;

Done with #1475, #1470 (and #1471).

> * The selector should be internally represented as `CompoundSelector [Combinator, CompoundSelector]*` where `Combinator` also includes 'descendent combinator' (space - note the reference above is regarding syntax rather than representation) -
>       
>   * `Combinator` could simply be a string (`>`, `+`, `~` or ` `) and may not need its own class;

Done with #1478, #1487, #1488, #1494, #1496.  `Combinator` has a class, since that made it easier to represent the list, via an interface type.
 
> * We do not need to initially create any API to manipulate the internal representation; it would be a nice-to-have for the future;

Creating an API made sense: `Selector\Component` is an interface for either `CompoundSelector` or `Combinator` and includes a `setValue()` method.  `Selector` also has a `setComponents()` method to set them all at once.

> * At a later date we can further break down `CompoundSelector` if desired.

[...]

PR #1475, opened 2026-01-23
[TASK] Add `Selector::parse()`

JakeQZ:
The implementation is based on `DeclarationBlock::parseSelector()`, which will be updated to use this instead in #1470.

Part of #1325.

### DIFF
--- a/src/Property/Selector.php
+++ b/src/Property/Selector.php
@@ -133,6 +133,12 @@ class Selector implements Renderable
                         break 2;
                     }
                     break;
+                default:
+                    // This will never happen unless something gets broken in `ParserState`.
+                    throw new \UnexpectedValueException(
+                        'Unexpected character \'' . $nextCharacter
+                        . '\' returned from `ParserState::peek()` in `Selector::parse()`'
+                    );
             }
             $selectorParts[] = $parserState->consume(1);
         }
--- a/tests/Unit/Property/SelectorTest.php
+++ b/tests/Unit/Property/SelectorTest.php
@@ -226,7 +226,6 @@ final class SelectorTest extends TestCase
         $result = [];
         Selector::parse(new ParserState($selector, Settings::create()), $result);
 
-        self::assertArrayHasKey(0, $result);
         self::assertInstanceOf(Comment::class, $result[0]);
         self::assertSame('comment', $result[0]->getComment());
     }
@@ -250,10 +249,8 @@ final class SelectorTest extends TestCase
         $result = [];
         Selector::parse(new ParserState('/*comment1*/a/*comment2*/', Settings::create()), $result);
 
-        self::assertArrayHasKey(0, $result);
         self::assertInstanceOf(Comment::class, $result[0]);
         self::assertSame('comment1', $result[0]->getComment());
-        self::assertArrayHasKey(1, $result);
         self::assertInstanceOf(Comment::class, $result[1]);
         self::assertSame('comment2', $result[1]->getComment());
     }


tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,146 @@ public static function provideInvalidSelectors(): array
...
     }
 
+    /**
+     * @return array<non-empty-string, array{0: string}>
+     */
+    public static function provideInvalidSelectorsForParse(): array
+    {
+        return [
+            'empty string' => [''],
+            'whitespace only' => [" \t\n\r"],
```
oliverklee: Let's split this up into one case per whitespace type. If anything breaks, this will allow us to see which kind of whitespace parsing is acting up.

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,146 @@ public static function provideInvalidSelectors(): array
...
+            'not missing closing brace' => [':not(a'],
+            'not missing opening brace' => [':not a)'],
```
oliverklee: I got quite confused when I read these 2 descriptions the first time. :-)

Can we reword these two descriptions so the "not" is recognizable as a literal, e.g. like this?

'a "not" missing the opening/closing brace"

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,146 @@ public static function provideInvalidSelectors(): array
...
+            'attribute value missing closing single quote' => ['a[href=\'#top]'],
+            'attribute value missing closing double quote' => ['a[href="#top]'],
```
oliverklee, line 130: Let's also have two additional tests where the value has one type of quote for opening and the other for closing.

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,146 @@ public static function provideInvalidSelectors(): array
...
+        Selector::parse(new ParserState($selector, Settings::create()));
+    }
+
+    /**
+     * @return array<non-empty-string, array{0: non-empty-string}>
+     */
+    public static function provideStopCharacters(): array
+    {
+        return [
+            ',' => ['{'],
```
oliverklee: copy'n paste error?

```suggestion
            ',' => [','],
```

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,146 @@ public static function provideInvalidSelectors(): array
...
+     * @dataProvider provideSelectorsWithAndWithoutComment
+     */
+    public function parseExtractsCommentFromSelector(string $selector): void
+    {
+        $result = [];
+        Selector::parse(new ParserState($selector, Settings::create()), $result);
+
+        self::assertArrayHasKey(0, $result);
+        self::assertInstanceOf(Comment::class, $result[0]);
+        self::assertSame('comment', $result[0]->getComment());
```
oliverklee, line 230: This will really benefit from `getArrayRespresentation` (once we have it for that class).

tests/Unit/Property/SelectorTest.php
JakeQZ: Done.

tests/Unit/Property/SelectorTest.php
JakeQZ: Good point.  Done (with backticks, and adding the colon).

tests/Unit/Property/SelectorTest.php
JakeQZ: Good suggestion.  Done.

tests/Unit/Property/SelectorTest.php
JakeQZ: Oops.  My secret method has been found out.  Fixed.

tests/Unit/Property/SelectorTest.php
JakeQZ: `Comment` does already implement it.  But it doesn't seem to help in this case, as using it would lead to:

```php
self::assertArrayHasKey(0, $result);
self::assertInstanceOf(Renderable::class, $result[0]);
self::assertSame(['class' => 'Comment', 'contents' => 'comment'], $result[0]->getArrayRespresentation());
```

... which looks more clunky to me, and I can't think of any way it makes the test tighter.

It might be useful in cases where `$result` is not an array.  (And of course when `Comment` is a property of another `Renderable`.)

oliverklee: I've created #1476 to make these kinds of tests a bit less clunky.

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,151 @@ public static function provideInvalidSelectors(): array
...
+    /**
+     * @return array<non-empty-string, array{0: non-empty-string, 1: non-empty-string}>
+     */
+    public static function provideSelectorsWithAndWithoutComment(): array
+    {
+        return [
+            'comment before' => ['/*comment*/body', 'body'],
+            'comment after' => ['body/*comment*/', 'body'],
+            'comment within' => ['./*comment*/teapot', '.teapot'],
+            'comment within function' => [':not(#your-mug,/*comment*/.their-mug)', ':not(#your-mug,.their-mug)'],
```
oliverklee, line 197: Are you … mugging me? :wink:

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,151 @@ public static function provideInvalidSelectors(): array
...
+     * @param non-empty-string $selector
+     *
+     * @dataProvider provideSelectorsWithAndWithoutComment
+     */
+    public function parseExtractsCommentFromSelector(string $selector): void
+    {
+        $result = [];
+        Selector::parse(new ParserState($selector, Settings::create()), $result);
+
+        self::assertArrayHasKey(0, $result);
```
oliverklee: We can drop this line. PHPUnit will warn us if the key does not exist in the next line.

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,151 @@ public static function provideInvalidSelectors(): array
...
+
+    /**
+     * @test
+     */
+    public function parseExtractsTwoCommentsFromSelector(): void
+    {
+        $result = [];
+        Selector::parse(new ParserState('/*comment1*/a/*comment2*/', Settings::create()), $result);
+
+        self::assertArrayHasKey(0, $result);
```
oliverklee: We can drop this line.

tests/Unit/Property/SelectorTest.php
```diff
@@ -93,6 +113,151 @@ public static function provideInvalidSelectors(): array
...
+        self::assertInstanceOf(Comment::class, $result[0]);
+        self::assertSame('comment1', $result[0]->getComment());
+        self::assertArrayHasKey(1, $result);
```
oliverklee: We can drop this line.

src/Property/Selector.php
```diff
@@ -63,11 +66,105 @@ public static function isValid(string $selector): bool
...
+                        break 2;
+                    }
+                    break;
+                case '{':
+                    // The fallthrough is intentional.
+                case '}':
+                    if (!\is_string($stringWrapperCharacter)) {
+                        break 2;
+                    }
+                    break;
```
oliverklee, line 135: We should have a `default` case. If I read the code correctly, that case should never be reached, and reaching it would indicate a bug in our code. We should throw an exception in that case, maybe an `OutOfBoundsException`.

Or would a valid default case make sense here?

tests/Unit/Property/SelectorTest.php
JakeQZ: Done.

tests/Unit/Property/SelectorTest.php
JakeQZ: Done.

tests/Unit/Property/SelectorTest.php
JakeQZ: Done.  We can't, however (IIRC), drop the `assertInstanceOf` because Stan will complain.

IIRC, accessing a non-existent array element does not generate an `Error` exception, merely a PHP warning/notice.  Whereas attempting to call a non-existent method, or a method on a non-object causes an `Error`, which PHPUnit would nonetheless catch.

src/Property/Selector.php
JakeQZ: It should never happen, as long as the logic in `ParserState` is sound.  I've opted for `UnexpectedValueException`.

PR #1477, opened 2026-01-23
[TASK] Have `comsumeWhiteSpace()` return the consumed

JakeQZ:
Instead, have the comments array passed by reference to be appended to.  This avoids various methods having to do an `array_merge`.  Also, knowing if any whitespace was consumed will be needed for #1471.

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -188,7 +188,11 @@ class ParserState
     }
 
     /**
-     * @param list<Comment> $comments
+     * Consumes whitespace and/or comments until the next non-whitespace character that isn't a slash opening a comment.
+     *
+     * @param list<Comment> $comments Any comments consumed will be appended to this array.
+     *
+     * @return string the whitespace consumed, without the comments
      *
      * @throws UnexpectedEOFException
      * @throws UnexpectedTokenException
--- a/tests/Unit/Parsing/ParserStateTest.php
+++ b/tests/Unit/Parsing/ParserStateTest.php
@@ -8,6 +8,7 @@ use PHPUnit\Framework\TestCase;
 use Sabberworm\CSS\Comment\Comment;
 use Sabberworm\CSS\Parsing\ParserState;
 use Sabberworm\CSS\Settings;
+use TRegx\PhpUnit\DataProviders\DataProvider;
 
 /**
  * @covers \Sabberworm\CSS\Parsing\ParserState
@@ -245,6 +246,8 @@ final class ParserStateTest extends TestCase
     /**
      * @test
      *
+     * @param non-empty-string $whitespace
+     *
      * @dataProvider provideWhitespace
      */
     public function consumeWhiteSpaceExtractsCommentWithSurroundingWhitespace(string $whitespace): void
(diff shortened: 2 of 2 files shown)


src/Parsing/ParserState.php
```diff
@@ -197,12 +197,12 @@ public function parseCharacter(bool $isForIdentifier): ?string
      * This method may change the state of the object by advancing the internal position;
      * it does not simply 'get' a value.
```
JakeQZ, line 202: This is still required, despite the change.  I'm slightly disappointed that Stan can't work this out, as he is so good at everything else.

JakeQZ:
I hope [David Gilmour](https://en.wikipedia.org/wiki/David_Gilmour) will appreciate this.

src/Parsing/ParserState.php
oliverklee: It might at higher levels (with PHP-CSS-Parser, we're still only at level 5).

oliverklee:
> I hope [David Gilmour](https://en.wikipedia.org/wiki/David_Gilmour) will appreciate this.

I don't understand the connection to this PR. :cry: Can you help me out?

src/Parsing/ParserState.php
```diff
@@ -197,12 +197,12 @@ public function parseCharacter(bool $isForIdentifier): ?string
...
      */
```
oliverklee, line 203: Maybe we can add a sentence or two to the method description explaining what the method does and what the return value means.

tests/Unit/Parsing/ParserStateTest.php
```diff
@@ -157,4 +157,105 @@ public function consumeIfComesUpdatesLineNumber(): void
...
+    }
+
+    /**
+     * @test
+     *
+     * @dataProvider provideWhitespace
+     */
+    public function consumeWhiteSpaceExtractsCommentWithSurroundingWhitespace(string $whitespace): void
+    {
+        $commentText = 'A walk-on part in the war for a lead role in a cage?';
```
oliverklee, line 255: Ah, that's the reference. Got it. :-)

JakeQZ:
> Maybe we can add some human-readable description explaining how the method works.

I've added some description to the DocBlock for the method.  I've also added a test to confirm non-whitespace is not consumed, and tightened up a type annotation.

JakeQZ:
> > I hope [David Gilmour](https://en.wikipedia.org/wiki/David_Gilmour) will appreciate this.
> 
> I don't understand the connection to this PR. 😢 Can you help me out?

While writing the tests, I had Pink Floyd's [Wish You Were Hear](https://www.youtube.com/watch?v=K6qj09OHvjw) as an earworm, so when constructing some test comments, I put in some of the lyrics from it.  (It's akin to the [Hello Dolly example plugin](https://en-gb.wordpress.org/plugins/hello-dolly/) that comes preinstalled with WordPress.)

src/Parsing/ParserState.php
JakeQZ: I thought that higher levels just gave more warnings, rather than more deeply analyzing the code, though I may be wrong.  It's strange that it thinks that a method call which isn't a `get` won't modify the object.

PR #1478, opened 2026-01-23
[TASK] Add `SelectorComponent` interface

JakeQZ:
This will be implemented by `CompoundSelector` and `Combinator` to break up the components of a selector.

Part of #1325.

### DIFF
--- a/src/Property/Selector/SelectorComponent.php
+++ b/src/Property/Selector/SelectorComponent.php
@@ -5,24 +5,20 @@ declare(strict_types=1);
 namespace Sabberworm\CSS\Property\Selector;
 
 /**
- * @see https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Selectors/Selector_structure
+ * This interface is for a class that represents a part of a selector which is either a compound selector (or a simple
+ * selector, which is effectively a compound selector without any compounding) or a selector combinator.
  *
- * A complex selector is a sequence of one or more simple and/or compound selectors that are separated by combinators,
- * including the white space descendant combinator.
- *
- * This is essentially a 'selector', as a 'selector list' is a list of comma-separated selectors.
- *
- * A compound selector is a superset that includes a simple selector:
- * - `li` is a simple selector (but can be treated as a compound selector);
- * - `li:last-child` is a compound selector;
- * - `ul li:last-child` is a complex selector as it has a combinator (the whitespace descendent combinator).
- *
- * A (complex) selector can be decomposed thus:
+ * It allows a selector to be represented as an array of objects that implement this interface.
+ * This is the formal definition:
  * selector = compound-selector [combinator, compound-selector]*
  *
- * It's a list of alternating types.
+ * The selector is comprised of an array of alternating types that can't be easily represented in a type-safe manner
+ * without this.
  *
- * This interface covers both types, so they can be put into an array of the interface type.
+ * 'Selector component' is not a known grammar in the spec, but a convenience for the implementation.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Selectors/Selector_structure
+ * @see https://www.w3.org/TR/selectors-4/#structure
  */
 interface SelectorComponent
 {


JakeQZ:
I'd like feedback on method names (and also interface name), as once committed it will be harder to change later.

#1471 is an early example of implementation (with different names).

Refs:
- https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Selectors/Combinators
- https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Selectors/Selector_structure#compound_selector

src/Property/Selector/SelectorComponent.php
```diff
@@ -0,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Property\Selector;
+
+interface SelectorComponent
```
oliverklee, line 23: Maybe we can add the links to the specs to the interface description, and document that multiple instance will form a compound selector.

oliverklee: As for the naming, https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Selectors/Selector_structure#compound_selector uses the term "simple selector". Maybe we can use that instead: `SimpleSelector`

src/Property/Selector/SelectorComponent.php
```diff
@@ -0,0 +1,23 @@
...
+{
+    /**
+     * @return non-empty-string
+     */
+    public function getStringValue(): string;
```
oliverklee: I'm not completely sure about this naming …

If it's the value, we probably don't need the `String` as part of the method names, und just use `set/setValue` instead.

src/Property/Selector/SelectorComponent.php
JakeQZ: > A compound selector is a sequence of [simple selectors](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Selectors/Selector_structure#simple_selector) that are not separated by a [combinator](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Selectors/Selectors_and_combinators#combinators).

We can further break down the `CompoundSelector` later.  But the operands of the combinators are compound selectors.

src/Property/Selector/SelectorComponent.php
JakeQZ: It's an interface to get and set the content as a string representation.

Getting would return, e.g. `li.mug` for a compound selector, or `>` for a combinator.

Setting would be the only way to modify the object.

The specific implementation would need to handle 'setting' for its own internal representation (perhaps doing some parsing), but would be the only way to modify the object.  But selectors are generally seen and represented as strings, so it should just be a case of modifying the internal string-representation property (or when `CompoundSelector` is broken up into a list of `SimpleSelector`s, doing a bit of parsing).

I'm inclined to agree that just `setValue`/`getValue` would be sufficient.

src/Property/Selector/SelectorComponent.php
oliverklee: Hm, looks like I still don't fully grasp what the interface represents.

Could you please add a description to it? Then I can think about naming it in a more informed way. :-)

src/Property/Selector/SelectorComponent.php
oliverklee: Then let's go for `setValue`/`getValue`, I'd say.

JakeQZ: Done.

src/Property/Selector/SelectorComponent.php
JakeQZ: Added a DocBlock with all the info I can think of.

oliverklee: I'm still struggling to understand this … The description explains complex selectors, compound selectors, selectors and selector lists, but not what a selector component is or how it relates to the other things.

I'd expect the first sentence of the description to be something like this: "A selector component is …"

JakeQZ: I've changed the DocBlock.  Hopefully it is clearer now...

JakeQZ:
GitHub won't let me reply to your review comment.

I've changed the DocBlock.  Is it now clearer?

JakeQZ:
As this is already in the namespace `Selector`, I am now wondering if the interface should be simply `Component`, rather than `SelectorComponent`.  The extra `Selector` in the name seems redundant.

E.g.
```php
use Sabberworm\CSS\Property\Selector\SelectorComponent;
```

Might look better as
```php
use Sabberworm\CSS\Property\Selector\Component;
```

@oliverklee WDYT?

oliverklee:
@JakeQZ Totally makes sense! So I would be fine with dropping the duplicate `Selector` from the interface name.

PR #1482, opened 2026-01-26
Remove `thecodingmachine/safe` dependency

SjorsO:
This PR drops the dependency on [`thecodingmachine/safe`](https://github.com/thecodingmachine/safe) by absorbing the 7 functions that PHP-CSS-Parser actually uses (out of the 1,110+ functions that the dependency defines).

Removing this dependency significantly reduces the size of PHP-CSS-Parser:

| Metric | Before | After | Reduction |
|-------|--------|-------|-----------|
| Uncompressed size (composer --no-dev) | 13 MB | 5.7 MB | 7.3 MB (56% less) |
| Compressed size | 5.6 MB | 4.2 MB | 1.4 MB (25% less) |
| Lines of text | 274,141 | 44,676 | 84% less |
| Number of files | 801 | 333 | 58% less |

PHP-CSS-Parser is [installed 3.8 million times per month](https://packagist.org/packages/sabberworm/php-css-parser/stats), assuming one compressed download per install, then this PR would save ~5 TB of bandwidth monthly. Removing this many lines of text may also speed up IDE indexing, although that's hard to measure.

This PR is a breaking change, code like this will no longer work:
```php
try {
    $parser->parse($css);
} catch (Safe\Exceptions\PcreException $e) {
    // ...
}
```
Code like this must be updated to catch `Sabberworm\CSS\Safe\Exceptions\PcreException $e` instead (although it seems unlikely to me that anyone is catching these exceptions specifically instead of `Exception` or `Throwable`).

I haven't removed the `thecodingmachine/phpstan-safe-rule` PHPStan rule, it still works as before.

I've excluded `safe_functions.php` from PHPStan both because the phpdocs don't match the inspections, and because it would warn about the unsafe functions in this file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ Please also have a look at our
 
 ### Changed
 
+- Remove `thecodingmachine/safe` dependency (#1479)
 - Clean up extra whitespace in CSS selector (#1398)
 - The array keys passed to `DeclarationBlock::setSelectors()` are no longer
   preserved (#1407)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Please also have a look at our
 
 ### Changed
 
-- Remove `thecodingmachine/safe` dependency (#1479)
+- Remove `thecodingmachine/safe` dependency (#1482)
 - Clean up extra whitespace in CSS selector (#1398)
 - The array keys passed to `DeclarationBlock::setSelectors()` are no longer
   preserved (#1407)
--- a/bin/quickdump.php
+++ b/bin/quickdump.php
@@ -5,7 +5,7 @@ declare(strict_types=1);
 
 use Sabberworm\CSS\Parser;
 
-use function Safe\file_get_contents;
+use function Sabberworm\CSS\Safe\file_get_contents;
 
 /**
  * This script is used for generating the examples in the README.
--- a/composer.json
+++ b/composer.json
@@ -24,8 +24,7 @@
     "homepage": "https://www.sabberworm.com/blog/2010/6/10/php-css-parser",
     "require": {
         "php": "^7.2.0 || ~8.0.0 || ~8.1.0 || ~8.2.0 || ~8.3.0 || ~8.4.0 || ~8.5.0",
-        "ext-iconv": "*",
-        "thecodingmachine/safe": "^1.3 || ^2.5 || ^3.3"
+        "ext-iconv": "*"
     },
     "require-dev": {
         "php-parallel-lint/php-parallel-lint": "1.4.0",
@@ -44,6 +43,9 @@
         "ext-mbstring": "for parsing UTF-8 CSS"
     },
     "autoload": {
+        "files": [
+            "src/Safe/safe_functions.php"
+        ],
         "psr-4": {
             "Sabberworm\\CSS\\": "src/"
         }
--- a/config/phpstan.neon
+++ b/config/phpstan.neon
@@ -15,6 +15,9 @@ parameters:
     - %currentWorkingDirectory%/src/
     - %currentWorkingDirectory%/tests/
 
+  excludePaths:
+    - %currentWorkingDirectory%/src/Safe/safe_functions.php
+
   type_perfect:
     no_mixed_property: true
     no_mixed_caller: true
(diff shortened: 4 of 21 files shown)


JakeQZ:
Thanks @SjorsO.  I did not realize 'Safe' was such a large library.  I think that may because it is made by autogenerating the code from the PHP documenation (on , and puts the entire lot in the DocBlocks.  For example the DocBlock for `preg_match_all()` is  343 lines long.  Also, it generates a separate set of functions for each supported PHP version (e.g. 8.1, 8.2, 8.3, 8.4, 8.5).  Clearly the library could be improved to significantly reduce its size.

I've allowed the CI checks to run, which has picked up a few minor issues.  Would you be able to resolve those?

> This PR is a breaking change, code like this will no longer work:
> 
> ```
> try {
>     $parser->parse($css);
> } catch (Safe\Exceptions\PcreException $e) {
>     // ...
> }
> ```

We have only just started using 'Safe', and didn't consider that it may throw additional exceptions to be a breaking change.

> Code like this must be updated to catch `Sabberworm\CSS\Safe\Exceptions\PcreException $e` instead (although it seems unlikely to me that anyone is catching these exceptions specifically instead of `Exception` or `Throwable`).

I agree; I think it's highly unlikely, particularly as we have only just started using 'Safe'.

oliverklee:
If we copy those functions to our package, the added code becomes our responsibility and needs to follow the standards we have for new code: It needs proper test coverage, and we need to maintain it for the foreseeable future, including keeping it up to date with future PHP releases.

Someone needs to do this work. I’m inclined to not take this responsibility myself.

@SjorsO @JakeQZ Would any of you be willing to write the tests and shoulder maintaining the new functions?

JakeQZ:
> If we copy those functions to our package ...

It doesn't stop there.  We are also now using the 'Safe' library in [Emogrifier](https://github.com/MyIntervals/emogrifier) of which this package is a dependency.  Emogrifier is used by the [WooCommerce](https://woocommerce.com/) WordPress plugin which is installed on over 4 million websites.

I don't think this is a tenable solution, but I don't know what is.

Options might include:
- Apply fixes to the 'Safe' library so that
  - DocBlocks only include the necessary type annotations;
  - The set of generated functions is not duplicated for each PHP version.
- Revert all changes to use the 'Safe' library.

Obviously it is disappointing that the 'Safe' library turns out to be bloatware.

JakeQZ:
> Options might include:

Or create a separate package akin to 'Safe' that only includes the most commonly-used functions, and does not auto-generate the source.

JakeQZ:
FWIW, I checked the size of the package I have installed with PHP 8.3.  It's 6.31 MB, with 573 files.  That does not seem like a lot to me, given that a typical photo from a phone is around 3 MB, and also considering that some popular WordPress plugins are much heftier:

- UpdraftPlus (1.22.11):
  - 25.7 MB
  - 1,638 files
- JetPack (10.3):
  - 25.6 MB
  - 1,822 files

Though I agree it's much more than it should be.  I think the way forward is to get the issues fixed in the 'Safe' libraray:
- Do not include redundant DocBlock content; only include the type annotations;
- Avoid having a separate directory of functions for each PHP version;
- Have the fixes backported to the 1.x and 2.x branches.

JakeQZ:
> Though I agree it's much more than it should be. I think the way forward is to get the issues fixed in the 'Safe' libraray:
> 
> * Do not include redundant DocBlock content; only include the type annotations;
> * Avoid having a separate directory of functions for each PHP version;
>  * Have the fixes backported to the 1.x and 2.x branches.

https://github.com/thecodingmachine/safe/issues/706

JakeQZ:
> * Revert all changes to use the 'Safe' library.

That might be the best option for the time being, although a PITA.

SjorsO:
> > * Revert all changes to use the 'Safe' library.
> 
> That might be the best option for the time being, although a PITA.

Personally I'd go with explicitly checking for `false` in places where it actually matters. I would avoid adding any dependency to this library, even a smaller or custom version of "thecodingmachine/safe".

I'm honestly not exactly sure why these "safe" functions are necessary. All other libraries I use don't seem to use them (including all of Laravel).

oliverklee:
> I'm honestly not exactly sure why these "safe" functions are necessary.

It helps us keep our code simple as we don't need to check for `false` when calling these functions, but can use the exceptions instead.

SjorsO:
For reference, I've opened up a new PR with a different approach: https://github.com/MyIntervals/PHP-CSS-Parser/pull/1484

oliverklee:
Closing in favor of #1484.

PR #1484, opened 2026-01-27
Remove `thecodingmachine/safe` dependency

SjorsO:
This PR drops the dependency on [`thecodingmachine/safe`](https://github.com/thecodingmachine/safe).

Removing this dependency significantly reduces the size of PHP-CSS-Parser:

| Metric | Before | After | Reduction |
|-------|--------|-------|-----------|
| Uncompressed size (composer --no-dev) | 2.6 MB | 0.9 MB | 1.7 MB (65% less) |
| Compressed size | 0.58 MB | 0.25 MB | 0.33 MB (56% less) |
| Lines of text | 91,111 | 29,309 | 68% less |
| Number of files | 534 | 216 | 60% less |

Removing this dependency also:
- Reduces supply chain attack risk (worth taking seriously after the recent Axios incident)
- Saves ~2TB of bandwidth per month (assuming 1 compressed download for [6.8 million installs this month](https://packagist.org/packages/sabberworm/php-css-parser/stats))
-  Speeds up IDE indexing by removing ~61,000 LoC and 318 files


src/Parsing/ParserState.php
```diff
@@ -319,7 +339,8 @@ public function consumeExpression(string $expression, ?int $maximumLength = null
     {
         $matches = null;
         $input = ($maximumLength !== null) ? $this->peek($maximumLength) : $this->inputLeft();
-        if (preg_match($expression, $input, $matches, PREG_OFFSET_CAPTURE) !== 1) {
+        /** @phpstan-ignore theCodingMachineSafe.function */
+        if (\preg_match($expression, $input, $matches, PREG_OFFSET_CAPTURE) !== 1) {
             throw new UnexpectedTokenException($expression, $this->peek(5), 'expression', $this->lineNumber);
```
SjorsO, line 347: No need to check for `false` here. If this `preg_match` returns `false` then we'll already throw an exception

src/Property/Selector.php
```diff
@@ -183,7 +184,8 @@ public function setSelector(string $selector): void
         $hasAttribute = \strpos($selector, '[') !== false;
 
         // Whitespace can't be adjusted within an attribute selector, as it would change its meaning
-        $this->selector = !$hasAttribute ? preg_replace('/\\s++/', ' ', $selector) : $selector;
+        /** @phpstan-ignore theCodingMachineSafe.function */
+        $this->selector = !$hasAttribute ? \preg_replace('/\\s++/', ' ', $selector) : $selector;
```
SjorsO: `preg_replace` cannot return `false`. The dependency has `preg_replace` as a [special case](https://github.com/thecodingmachine/safe/blob/98d020e21833599d4fc5fb94a398790de6101562/lib/special_cases.php#L143).

I think it should be fine to use the normal version of `preg_replace`

src/Value/CalcFunction.php
```diff
@@ -62,7 +60,9 @@ public static function parse(ParserState $parserState, bool $ignoreCase = false)
                 if (\in_array($parserState->peek(), $operators, true)) {
                     if (($parserState->comes('-') || $parserState->comes('+'))) {
                         if (
+                            /** @phpstan-ignore theCodingMachineSafe.function */
                             preg_match('/\\s/', $parserState->peek(1, -1)) !== 1
+                            /** @phpstan-ignore theCodingMachineSafe.function */
                             || preg_match('/\\s/', $parserState->peek(1, 1)) !== 1
```
SjorsO: if either of these returns `false` we'll already throw an exception

src/Property/Selector.php
oliverklee: According to https://www.php.net/manual/en/function.preg-replace.php#refsect1-function.preg-replace-returnvalues, it can return a string, an array or `null`.

SjorsO: You're right, I missed that. I've pushed a commit that checks for `null`

JakeQZ:
Appreciate your efforts, @SjorsO (code changes look solid), but I honestly don't think this is the best way forward.  I think we need to resolve the bloat in the in the 'Safe' package, for which I created https://github.com/thecodingmachine/safe/issues/706, and have had a response.

SjorsO:
> ... but I honestly don't think this is the best way forward. I think we need to resolve the bloat in the in the 'Safe' package

Given how widely used PHP-CSS-Parser and emogrifier are, I consider getting rid of any dependency a huge win.

Even if `thecodingmachine/safe` manages to slim down and remove all the docblocks and other PHP versions, it still [autoloads 79 files that contain 1100+ functions](https://github.com/thecodingmachine/safe/blob/98d020e21833599d4fc5fb94a398790de6101562/composer.json#L12) (!!), I find this hard to justify, especially considering the alternative is simply checking for `false`.

I think this PR is a nice middle ground: no dependency that thousands of people have to install, but still a PHPStan rule that warns about unsafe usage during development.

JakeQZ:
https://github.com/thecodingmachine/safe/issues/660 is another reason to avoid using this dependency.  In WordPress, each plugin (or theme) brings in its own versions of each dependency, which leads to conflicts, though, in fairness, that is an issue with WordPress.

oliverklee:
Marking this as draft as we're currently working to drastically shrink that package in https://github.com/thecodingmachine/safe/issues/706.

We should revisit this PR here once a new version of Safe has been released.

JakeQZ:
> Marking this as draft as we're currently working to drastically shrink that package in [thecodingmachine/safe#706](https://github.com/thecodingmachine/safe/issues/706).
> 
> We should revisit this PR here once a new version of Safe has been released.

With [version 3.4](https://github.com/thecodingmachine/safe/releases/tag/v3.4.0), the size is reduced from 6.31Mb to 1.58Mb.

To put it in context, this is comparable to the size of one or two photos in HD quality.

Another factor that impacts website managers is the number of individual files.  Hosting providers using cPanel usually impose a limit on the number of file handles.  The new version reduces the number of files from 573 to 318.

The new version is only availble for PHP versions >=8.1, and I don't expect will be backported - the older PHP versions are already EOL.

Where do we go from here?

I have no strong opinion, but don't see the download size and number of files as a significant impact, and would prefer to use the library to keep the code robust and clean, without having to implement additional checks and tests.

oliverklee:
> but don't see the download size and number of files as a significant impact, and would prefer to use the library to keep the code robust and clean, without having to implement additional checks and tests.

I fully agree.

SjorsO:
For the record, I still strongly feel that adding 1.58mb and 318 files to thousands (millions?) of applications is way too much. 

Removing this dependency is a good step to prevent everyone's `vendor` directory from turning into a second `node_modules`.

SjorsO:
I still hope you'll reconsider and remove this dependency. `thecodingmachine/safe` still adds 318 extra files and 2.5MB to every project that uses DomPDF. Removing it would also reduce supply chain attack risk (something worth taking seriously after the recent Axios incident).

I'm happy to resolve any merge conflicts and re-review the changes if you decide to move forward with this PR.

oliverklee (commented):
I'm still hesitant about this.

For this PR to possibly get in, it still needs quite a lot of work:

- We should also remove the `thecodingmachine/phpstan-safe-rule` dependency (as it's only relevant for projects using `thecodingmachine/safe`).
- We need to drop the added `@phpstan-ignore` annotations.
- So far, we've relied on Safe throwing exceptions when things go wrong. If we do that on our own, we'll need to fully cover the added exceptions with tests (or `assert` if we know some code will not fail).

@SjorsO Would you be willing to do the work? (Particularly, the test coverage will take some time.)

If so, we should split this up into smaller PRs, e.g., one PR per file.

@JakeQZ Would you be okay with this (assuming taht @SjorsO will do the heavy lifting)?

SjorsO:
> We should also remove the thecodingmachine/phpstan-safe-rule dependency (as it's only relevant for projects using thecodingmachine/safe).
>
> We need to drop the added @phpstan-ignore annotations.

Keeping the `thecodingmachine/phpstan-safe-rule` gives us the best of both worlds: as a dev dependency, it doesn't affect end users, and it alerts us whenever we use an unsafe function. When the rule flags a new unsafe call, we handle the `false` return explicitly and add a `@phpstan-ignore` comment to signal "this unsafe call is being handled properly."

Keeping this PHPStan rule guarantees we won't miss any unsafe function calls in the future. Are you sure you want to remove it as well?

oliverklee:
> Keeping this PHPStan rule guarantees we won't miss any unsafe function calls in the future. Are you sure you want to remove it as well?

Ah, I hadn't though of that. I'll need to think about that.

JakeQZ:
> @JakeQZ Would you be okay with this (assuming taht @SjorsO will do the heavy lifting)?

The main areas of concern are `preg_replace` etc., where failure is unpredictable, and very much based on the `$subject`.

I added a class to Emogrifier to cater for this, which was removed in https://github.com/MyIntervals/emogrifier/pull/1514 (I can't find when it was added because GitHub is a bit rubbish),

Perhaps that class can be reinstated for both projects,  And somehow shared.  Or an alternative solution be worked out.

I don't expect any of the other inbuilt functions to fail at all.  Some `assert()`s will surely sort those out.

SjorsO:
> > Keeping this PHPStan rule guarantees we won't miss any unsafe function calls in the future. Are you sure you want to remove it as well?
> 
> Ah, I hadn't though of that. I'll need to think about that.

Let me know once you've decided, then I'll get started

oliverklee:
> Let me know once you've decided, then I'll get started

Let's drop the PHPStan ruleset.

oliverklee:
> Let's drop the PHPStan ruleset.

But only after we've dropped the Safe package, i.e., after the last PR from a series of PRs (to make them faster to review) resulting from this PR here.

tests/Unit/Value/CalcFunctionTest.php
```diff
@@ -159,6 +159,10 @@ public function provideInvalidSyntax(): array
         return [
             'missing space around -' => ['calc(100%-20px)'],
             'missing space around +' => ['calc(100%+20px)'],
+            'missing space before -' => ['calc(100%- 20px)'],
+            'missing space before +' => ['calc(100%+ 20px)'],
+            'missing space after -' => ['calc(100% -20px)'],
+            'missing space after +' => ['calc(100% +20px)'],
```
SjorsO, line 165: These add test coverage for the two `preg_match` calls in `CalcFunction` (the second `preg_match` previously had no test coverage)

SjorsO:
@oliverklee This is ready for review. I've updated the PR description with the impact removing this dependency will have.

I hope it's alright that I've left this as a single PR. The changes should be straightforward enough to review together.

23 call sites are changed:
- 17 are simple "grab result + `false` check"
- 3 already threw an exception if the result was `false`, so no changes were needed
- 3 involve slightly more work on a loop condition:
   - [ParserState](https://github.com/MyIntervals/PHP-CSS-Parser/pull/1484/changes#diff-6871f481dbdbb2fb79629f40f8525404085da6984c5fc4b8c1981b44c0e34657L207)'s while loop
   - [CssString](https://github.com/MyIntervals/PHP-CSS-Parser/pull/1484/changes#diff-df24e9f3754cf6c586b68c01cf88753c935981250b3354baeaaf5041f8f2f4dbL61)'s while loop
   - [Value](https://github.com/MyIntervals/PHP-CSS-Parser/pull/1484/changes#diff-2e0a239d4b376b82b5bfaa90a74960eacd3e349e049dde49c9b271c9264c5daaL223)'s do...while loop

I wasn't able to add tests for most of the changed `preg_match` calls because it only returns `false` if:
- The pattern doesn't compile (impossible here, patterns are hardcoded)
- The backtrack/recursion limit is hit (impossible here, none of the patterns do backtracking)
- The `/u` flag is set and the subject is malformed UTF-8 (only two cases, I've added test coverage for those)

A follow-up PR should remove the `thecodingmachine/phpstan-safe-rule` dev dependency and the PHPStan annotations

tests/Unit/Parsing/ParserStateTest.php
```diff
@@ -303,4 +303,28 @@ public function consumeWhiteSpaceStopsAtNonWhitespace(string $nonWhitespace, str
...
+    /**
+     * @test
+     */
+    public function constructorThrowsForTextThatIsNotValidUtf8(): void
+    {
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionMessage('Unexpected error');
+
+        new ParserState("\xFF body{}", Settings::create());
+    }
```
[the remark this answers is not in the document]
SjorsO: Not an actual problem

src/Value/CalcFunction.php
```diff
@@ -62,8 +60,10 @@ public static function parse(ParserState $parserState, bool $ignoreCase = false)
                 if (\in_array($parserState->peek(), $operators, true)) {
                     if (($parserState->comes('-') || $parserState->comes('+'))) {
                         if (
-                            preg_match('/\\s/', $parserState->peek(1, -1)) !== 1
-                            || preg_match('/\\s/', $parserState->peek(1, 1)) !== 1
+                            /** @phpstan-ignore theCodingMachineSafe.function */
+                            \preg_match('/\\s/', $parserState->peek(1, -1)) !== 1
+                            /** @phpstan-ignore theCodingMachineSafe.function */
+                            || \preg_match('/\\s/', $parserState->peek(1, 1)) !== 1
```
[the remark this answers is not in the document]
SjorsO: `preg_match` returning `false` here should be impossible. It masking the error as an `UnexpectedTokenException` is fine

src/Parsing/ParserState.php
```diff
@@ -318,7 +342,8 @@ public function consumeExpression(string $expression, ?int $maximumLength = null
...
         }
```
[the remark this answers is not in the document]
SjorsO: `preg_match` can return `false` here if `$expression` is invalid, but that should never happen, and if it does then a `UnexpectedTokenException` is fine

SjorsO:
@oliverklee if there's anything I can do to help get this merged, then please let me know

oliverklee:
> if there's anything I can do to help get this merged, then please let me know

@SjorsO Thanks for the ping! Yes, indeed, there is.

Please let's split this up into multiple PRs. This will make reviewing and merging a lot easier and faster for us.

1. one for each changed class
2. then one last PR that drop the dependency and removes the PHPStan baseline entries

I propose you start with a single PR first and get that merged so you know what is necessary to get the PRs merged.

Some thoughts in advance:

- Please let's not introduce new `@phpstan-ignore` comments for the PREG functions. Instead, we should add the entries to the PHPStan baseline `composer phpstan:baselin` as part of the PRs.
- For each changed function, we need to decide whether the failure case can actually occur. If so, we need to cover it with a test. And if not, we should use an `assert` after the call to the PREG function.

Would this work for you?

oliverklee:
Marking as draft as this PR now is moved to smaller, separate PRs.

PR #1485, opened 2026-01-30
Fix Tailwind escaped quotes selector in Selector.php

samuil-banti-wpenigne:
Tailwind allows selectors that contain quotes like this one: `.before\:content-\[\'\'\]:before`

The goal of this PR is to detect this escaped quotes.

A way to test both the issue and the fix:
```
$content = "
    .before\:content-\[\'\'\]:before {
        content: \"→ \";
        color: blue;
    }
";
try {
    $settings = \Sabberworm\CSS\Settings::create()->beStrict();
    $parser = new \Sabberworm\CSS\Parser($content, $settings);
    $tree = $parser->parse();
    echo $tree->render(\Sabberworm\CSS\OutputFormat::createPretty());
} catch (Exception $e) {
    echo "Error: " . $e->getMessage() . "\n";
}
```

Proof that the selectors are valid:

```
<!DOCTYPE html>
<html lang="en">
<head>
    <meta charset="UTF-8">
    <meta name="viewport" content="width=device-width, initial-scale=1.0">
    <title>CSS Selector Example</title>
    <style>
        .before\:content-\[\'\'\]:before {
            content: "→ ";
            color: blue;
        }
        .before\:content-\[\"\"\]:before {
            content: "→ ";
            color: blue;
        }
    </style>
</head>
<body>
    <div class='before:content-[""]'>Double Quotes</div>
	<div class="before:content-['']">Single Quotes</div>
</body>
</html>
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@ Please also have a look at our
 
 - Add support for escaped quotes in the selectors (#1485)
 - Provide line number in exception message for mismatched parentheses in
-selector (#1435)
+  selector (#1435)
 - Add support for CSS container queries (#1400)
 
 ### Changed


oliverklee:
Hi @samuil-banti-wpenigne, thanks a lot!

Could you please add an automated test that fails without the fix and that passes with the fix (i.e., a regression test)?

src/Property/Selector.php
```diff
@@ -95,7 +95,10 @@ public static function parse(ParserState $parserState, array &$comments = []): s
...
                     // The fallthrough is intentional.
                 case '"':
-                    if (!\is_string($stringWrapperCharacter)) {
+                    $lastPart = \end($selectorParts);
+                    $backslashCount = \strspn(\strrev($lastPart), '\\');
+                    $quoteIsEscaped = ($backslashCount % 2 === 1);
+                    if (!\is_string($stringWrapperCharacter) && !$quoteIsEscaped) {
                         $stringWrapperCharacter = $nextCharacter;
                     } elseif ($stringWrapperCharacter === $nextCharacter) {
                         if (\substr(\end($selectorParts), -1) !== '\\') {
```
JakeQZ: Since `$lastPart` is introduced, it can be reused here:

```suggestion
                        if (\substr($lastPart, -1) !== '\\') {
```

JakeQZ (commented):
The code change looks fine to me, though I've added a further suggestion.

Re the unit tests, you should be able to add a couple of datasets to this data provider to cover it: https://github.com/MyIntervals/PHP-CSS-Parser/blob/f78fca507b88cd2049f039020c5cf4ef11d212b4/tests/Unit/Property/SelectorTest.php#L58

Could you also add an entry to `CHANGELOG.md` (in the root directory), newest first (with wording along the lines of "Support ..." rather than "Fix ...")?

Thanks <3

src/Property/Selector.php
samuil-banti-wpenigne: Good catch but I think we should change that as well so that we can handle cases where escaped backslash is followed by quote.

src/Property/Selector.php
```diff
@@ -95,12 +95,13 @@ public static function parse(ParserState $parserState, array &$comments = []): s
...
+                    $quoteIsEscaped = ($backslashCount % 2 === 1);
+                    if (!\is_string($stringWrapperCharacter) && !$quoteIsEscaped) {
                         $stringWrapperCharacter = $nextCharacter;
-                    } elseif ($stringWrapperCharacter === $nextCharacter) {
-                        if (\substr(\end($selectorParts), -1) !== '\\') {
-                            $stringWrapperCharacter = null;
-                        }
+                    } elseif ($stringWrapperCharacter === $nextCharacter && !$quoteIsEscaped) {
+                        $stringWrapperCharacter = null;
                     }
```
JakeQZ: We could probably further change the logic to check `$quoteIsEscaped` first, then have a simpler if/elseif.  I think that makes it easier to understand the code, and may save one runtime test:

```suggestion
                    if (!$quoteIsEscaped) {
                        if (!\is_string($stringWrapperCharacter)) {
                            $stringWrapperCharacter = $nextCharacter;
                        } elseif ($stringWrapperCharacter === $nextCharacter) {
                            $stringWrapperCharacter = null;
                        }
                    }
```

(Note: commiting the suggestion using GitHub might not work properly - the indentation was initially wrong after I used the 'Add a suggestion' button.)

tests/Unit/Property/SelectorTest.php
```diff
@@ -369,4 +369,89 @@ public function getArrayRepresentationThrowsException(): void
...
+     */
+    public function isValidForSelectorsWithEscapedQuotesReturnsTrue(string $selector): void
+    {
+        self::assertTrue(Selector::isValid($selector));
+    }
+
+    /**
+     * @test
+     */
+    public function parsesAttributeWithEscapedQuoteNotPrematurelyClosingString(): void
```
JakeQZ: Can I suggest a slightly more human-readable test method name:


```suggestion
    public function parsingAttributeWithEscapedQuoteDoesNotPrematurelyCloseString(): void
```

tests/Unit/Property/SelectorTest.php
```diff
@@ -369,4 +369,89 @@ public function getArrayRepresentationThrowsException(): void
...
+    {
+        // This tests that an escaped quote doesn't close the string prematurely
```
JakeQZ: With the modified method name, I think this comment becomes redundant, as it just repeats what the method name says.

```suggestion
```

tests/Unit/Property/SelectorTest.php
```diff
@@ -369,4 +369,89 @@ public function getArrayRepresentationThrowsException(): void
...
+        $result = Selector::parse(new ParserState($selector, Settings::create()));
+
+        self::assertInstanceOf(Selector::class, $result);
+        self::assertSame($selector, $result->getSelector());
+    }
+
+    /**
+     * @test
+     */
+    public function distinguishesEscapedFromUnescapedQuotes(): void
```
JakeQZ: I think the test method name should indicate that it's focusing on the `parse()` method:

```suggestion
    public function parseDistinguishesEscapedFromUnescapedQuotes(): void
```

tests/Unit/Property/SelectorTest.php
```diff
@@ -369,4 +369,89 @@ public function getArrayRepresentationThrowsException(): void
...
+
+        $result = Selector::parse(new ParserState($selector, Settings::create()));
+
+        self::assertSame($selector, $result->getSelector());
+    }
+
+    /**
+     * @test
+     */
+    public function handlesEvenNumberOfBackslashesBeforeQuote(): void
```
JakeQZ: Also here:

```suggestion
    public function parseHandlesEvenNumberOfBackslashesBeforeQuote(): void
```

CHANGELOG.md
```diff
@@ -13,6 +13,7 @@ Please also have a look at our
 - Provide line number in exception message for mismatched parentheses in
   selector (#1435)
 - Add support for CSS container queries (#1400)
+- Add support for escaped quotes in the selectors (#1485)
```
JakeQZ: I'll move the new entry to the top of the list...

```suggestion
- Add support for escaped quotes in the selectors (#1485)
- Provide line number in exception message for mismatched parentheses in
selector (#1435)
- Add support for CSS container queries (#1400)
```

CHANGELOG.md
```diff
@@ -10,8 +10,9 @@ Please also have a look at our
 
 ### Added
 
+- Add support for escaped quotes in the selectors (#1485)
 - Provide line number in exception message for mismatched parentheses in
-  selector (#1435)
+selector (#1435)
```
JakeQZ:
```suggestion
  selector (#1435)
```

JakeQZ:
I just realized that the tests are only covering escaped quotes within strings.

I've submitted #1489 to cover escaped quotes outside of strings, like in the example in the OP.  This just adds additional tests.  The code change remains solid and stands up to them :)

PR #1488, opened 2026-02-01
[TASK] Add `CompoundSelector` class

JakeQZ:
Part of #1325.

### DIFF
--- a/src/Property/Selector/CompoundSelector.php
+++ b/src/Property/Selector/CompoundSelector.php
@@ -21,6 +21,25 @@ class CompoundSelector implements Renderable, Component
 {
     use ShortClassNameProvider;
 
+    private const PARSER_STOP_CHARACTERS = [
+        '{',
+        '}',
+        '\'',
+        '"',
+        '(',
+        ')',
+        ',',
+        ' ',
+        "\t",
+        "\n",
+        "\r",
+        '>',
+        '+',
+        '~',
+        ParserState::EOF,
+        '', // `ParserState::peek()` returns empty string rather than `ParserState::EOF` when end of string is reached
+    ];
+
     private const SELECTOR_VALIDATION_RX = '/
         ^
             # not starting with whitespace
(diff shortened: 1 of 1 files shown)


JakeQZ:
Marking as 'draft' for now, as will need to include changes from #1485.

JakeQZ:
This now includes the changes from #1485 (parsing functionality is duplicated, but the duplication will eventually be removed from `Selector` when that is changed to use the new classes instead of a string), and is ready for review.

src/Property/Selector/CompoundSelector.php
```diff
@@ -0,0 +1,255 @@
...
+     * @throws UnexpectedTokenException
+     *
+     * @internal
+     */
+    public static function parse(ParserState $parserState, array &$comments = []): self
+    {
+        $selectorParts = [];
+        $stringWrapperCharacter = null;
+        $functionNestingLevel = 0;
+        static $stopCharacters = [
```
oliverklee: Would it make sense to move this to a class constant instead? (In general, I'm not a big fan of static local variables as these are not constant and hence might introduce hidden state.)

tests/Unit/Property/Selector/CompoundSelectorTest.php
```diff
@@ -0,0 +1,494 @@
...
+    }
+
+    /**
+     * @return array<non-empty-string, array{0: non-empty-string}>
+     */
+    public static function provideCompoundSelectorWithEscapedQuotes(): array
+    {
+        return [
+            'escaped double quote in double-quoted attribute' => ['a[href="test\\"value"]'],
+            'escaped single quote in single-quoted attribute' => ['a[href=\'test\\\'value\']'],
```
oliverklee: For the cases with escaped single quotes, I propose we use double quotes for the string itself, which would reduce additional quoting and make the data easier to read (for humans).

src/Property/Selector/CompoundSelector.php
JakeQZ: My thinking was that having them in the method makes it easier to check that the cases in the switch statement cover them all.  There are arguments either way; I've moved them to a constant.

tests/Unit/Property/Selector/CompoundSelectorTest.php
JakeQZ: I've changed the represntation of the strings that contain only single quotes.

I read somewhere that using double quotes to enclose strings no longer has any impact on performance.  If that's true, it may even be the case that using double quotes to enclose strings containing single quotes improves performance, since the interpreter doesn't have to deal with unescaping.

JakeQZ:
> Also, this should get a changelog entry (and maybe we can add the changelog entry for #1487 as well while we're at it).

I was planning to do that when the new classes are actually used in the `Selector` representation (and can reference these PRs in the changelog entry at that time).  At the moment they are just spare parts.

oliverklee (approved):

PR #1493, opened 2026-02-05
[TASK] Extract method `Selector::parseComponents`

JakeQZ:
This will allow the conversion of a selector string into an array of `Selector\Component` objects to also be used by the `setSelector()` method.

Part of #1325.


JakeQZ:
The new private method can be moved within the class definition as a separate PR - it's where it is now to make diffing easier for review.

PR #1495, opened 2026-02-05
[TASK] Render selector combinators with surrounding spaces by default

JakeQZ:
Part of #1325.

### DIFF
--- a/src/Property/Selector/Combinator.php
+++ b/src/Property/Selector/Combinator.php
@@ -96,7 +96,7 @@ class Combinator implements Component, Renderable
     public function render(OutputFormat $outputFormat): string
     {
         // TODO: allow optional spacing controlled via OutputFormat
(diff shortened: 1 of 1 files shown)


JakeQZ:
Eventually there will be an `OutputFormat` option to control this.

src/Property/Selector/Combinator.php
```diff
@@ -96,7 +96,7 @@ public function getSpecificity(): int
     public function render(OutputFormat $outputFormat): string
     {
         // TODO: allow optional spacing controlled via OutputFormat
-        return $this->getValue();
+        return $this->value !== ' ' ? ' ' . $this->value . ' ' : ' ';
```
oliverklee: Let‘s make this a bit quicker to read:

```suggestion
        return $this->value !== ' ' ? (' ' . $this->value . ' ') : ' ';
```

JakeQZ: Done.

PR #1496, opened 2026-02-05
[FEATURE] Represent `Selector` as `Component` objects

JakeQZ:
Part of #1325.

### DIFF
--- a/src/Property/Selector.php
+++ b/src/Property/Selector.php
@@ -77,7 +77,7 @@ class Selector implements Renderable
      * @param non-empty-string|non-empty-list<Component> $selector
      *        Providing a string is deprecated in version 9.2 and will not work from v10.0
      *
-     * @throws \UnexpectedValueException if the selector is not valid
+     * @throws UnexpectedTokenException if the selector is not valid
      */
     final public function __construct($selector)
     {
@@ -184,7 +184,7 @@ class Selector implements Renderable
     /**
      * @param non-empty-string $selector
      *
-     * @throws \UnexpectedValueException if the selector is not valid
+     * @throws UnexpectedTokenException if the selector is not valid
      *
      * @deprecated in version 9.2, will be removed in v10.0.  Use `setComponents()` instead.
      */


JakeQZ:
Would like initial feedback on this...

There's a bit more to do:
- A few more tests;
- Changelog entries.

It might be possible to break this down into seperate PR, but I can't think how - everything is interdependent.

JakeQZ:
> It might be possible to break this down into seperate PR, but I can't think how - everything is interdependent.

Perhaps an initial change would be to have `setSelector()` (and the constructor) throw an exception if the selector is invalid, using the existing `isValid()` method.  I wouldn't consider this to be a breaking change, more of a fix, though something that would probably go in the 'Changed' section of the changelog.

JakeQZ:
> Perhaps an initial change would be to have `setSelector()` (and the constructor) throw an exception if the selector is invalid, using the existing `isValid()` method. I wouldn't consider this to be a breaking change, more of a fix, though something that would probably go in the 'Changed' section of the changelog.

#1498 covers exceptions possibly being thrown by `setSelector()` and the constructor.

src/Property/Selector.php
```diff
@@ -137,52 +143,76 @@ public static function parse(ParserState $parserState, array &$comments = []): s
...
 
     /**
      * @throws \UnexpectedValueException if the selector is not valid
      */
     public function setSelector(string $selector): void
     {
-        if (!self::isValid($selector)) {
-            throw new \UnexpectedValueException("Selector `$selector` is not valid.");
-        }
+        $parserState = new ParserState($selector, Settings::create());
```
oliverklee, line 193: Working with the parser doesn't feel right in the setter. I'd expect the setter to get fully parsed things, and the responsibility of providing correct data falls on the caller.

Is there any other way we can validate `$selector`? Or maybe create a new setter?

JakeQZ: It's a means of supporting the legacy method of setting the selector as a string.  To decompose it into the components requires parsing.  We could deprecate setting the selector as a string.

JakeQZ:
> There's a bit more to do:
> * A few more tests;
> * Changelog entries.

Have added the additional tests and changelog entry.

src/Property/Selector.php
JakeQZ: There is a new setter: `setComponents()`.

oliverklee: > We could deprecate setting the selector as a string.

Yes, please let's do so.

JakeQZ: Perhaps as a separate further PR.

JakeQZ:
> Have added the additional tests and changelog entry.

Now actually committed that update.

JakeQZ:
#1499 partially implements `getArrayRepresentation()` to reduce the payload of this PR.

JakeQZ:
#1500 applies the change to the `Selector` constructor, to further reduce the payload of this PR.

Ideally we would like to change the constructor to accept an array of `Selector\Component` instead of a string.  But that would be a breaking change, which I'd like to avoid for now.  An alternative would be to allow either (for the time being), but that would mean the constructor parameter would have to be a `mixed` type, which I'm not keen on either.

src/Property/Selector.php
JakeQZ: I've added deprecation notices to this PR for both `getSelector()` and `setSelector()`.

oliverklee: I've added #1501 for removing the two methods.

src/Property/Selector.php
```diff
@@ -137,52 +143,81 @@ public static function parse(ParserState $parserState, array &$comments = []): s
...
+        return $this;
     }
 
+    /**
+     * @deprecated in version 9.2, will be removed in v10.0.  Use either `getComponents()` or `render()` instead.
+     */
     public function getSelector(): string
     {
-        return $this->selector;
+        return $this->render(new OutputFormat());
```
oliverklee, line 181: I getter actually rendering the selector (and using an output format) feels wrong to me - the data returned by the getter should be independent of how things actually render. Can we do it without calling `render(`)`?

tests/Unit/Property/SelectorTest.php
```diff
@@ -320,7 +323,7 @@ public function parseExtractsTwoCommentsFromSelector(): void
      */
     public function canConstructObjectWithEmptyState(): void
     {
-        $subject = new Selector('');
+        $subject = new Selector();
```
oliverklee: We should cover both cases: empty string, and with the state not being provided.

tests/Unit/Property/SelectorTest.php
```diff
@@ -329,10 +332,17 @@ public function canConstructObjectWithEmptyState(): void
...
      *
      * @dataProvider provideInvalidSelectors
+     * @dataProvider provideInvalidSelectorsForParse
      */
     public function constructorThrowsExceptionWithInvalidSelector(string $selector): void
     {
-        $this->expectException(\UnexpectedValueException::class);
+        // An empty string is allowed to construct an empty object
+        if ($selector === '') {
+            self::expectNotToPerformAssertions();
```
oliverklee: Please let's have a separate testcase for this as this case directly contradicts the behavior stated in the test method name. (And in general, a conditional in a test method is a code smell that tends to mark tests that mix different behavior, and where there's at least a second test method hidden, begging to be extracted.)

tests/Unit/Property/SelectorTest.php
```diff
@@ -341,16 +351,91 @@ public function constructorThrowsExceptionWithInvalidSelector(string $selector):
...
+    }
+
+    /**
+     * @test
+     *
+     * @param list<Component> $components
+     *
+     * @dataProvider provideComponentsAndArrayRepresentation
+     */
+    public function getComponentsReturnsComponentsSet(array $components): void
```
oliverklee, line 459: We also need a test for `getComponents` for cases where either nothing or am empty string is provided to the constructor (i.e., the default value for the components).

oliverklee:
> Ideally we would like to change the constructor to accept an array of `Selector\Component` instead of a string. But that would be a breaking change, which I'd like to avoid for now. An alternative would be to allow either (for the time being), but that would mean the constructor parameter would have to be a `mixed` type, which I'm not keen on either.

Or we use a union type (as PHPDoc type annotation): `string|array<Component>` and mark the string as deprecated.

src/Property/Selector.php
JakeQZ: I suppose that's possible.  But we are going to deprecate and remove this method.  So I don't consider it worthwhile to find a technically correct temporary alternative.

tests/Unit/Property/SelectorTest.php
JakeQZ: Done.

tests/Unit/Property/SelectorTest.php
JakeQZ: This is just eliminating one value from the data provider that is not valid in this situation.  Though adding another data provider may work...

JakeQZ: Done (hopefully).

tests/Unit/Property/SelectorTest.php
JakeQZ: Done.  But as per https://github.com/MyIntervals/PHP-CSS-Parser/pull/1496#issuecomment-3864024639 there may be a better way forward.

JakeQZ:
> Or we use a union type (as PHPDoc type annotation): `string|array<Component>` and mark the string as deprecated.

I was trying to avoid introducing a mixed type, but doing so has made things slightly messy.

I think your suggestion is the way forward.

JakeQZ:
> Or we use a union type (as PHPDoc type annotation): `string|array<Component>` and mark the string as deprecated.

That makes the change a bit cleaner.  Have done that, and discarded the idea of allowing `Selector` to be constructed in an empty state.

src/Property/Selector.php
oliverklee: Sounds fair.

src/Property/Selector.php
```diff
@@ -73,11 +72,18 @@ public static function isValid(string $selector): bool
     }
 
     /**
+     * @param string|list<Component> $selector
```
oliverklee: May any of these be empty, or should we annotate them with `non-empty-string` and `non-empty-list` (and also the getter/setter/property)?

src/Property/Selector.php
```diff
@@ -57,9 +56,9 @@ class Selector implements Renderable
         /ux';
 
     /**
-     * @var string
+     * @var list<Component>
      */
-    private $selector;
+    private $components = [];
```
oliverklee: As this is initialized in the constructor, the property doesn't need to have a default value anymore (and shouldn't have so that it communicates that there is no default value).

src/Property/Selector.php
```diff
@@ -140,52 +146,81 @@ public static function parse(ParserState $parserState, array &$comments = []): s
...
-        $selectorString = '';
-        foreach ($selectorParts as $selectorPart) {
-            $selectorPartValue = $selectorPart->getValue();
-            if (\in_array($selectorPartValue, ['>', '+', '~'], true)) {
-                $selectorString .= ' ' . $selectorPartValue . ' ';
-            } else {
-                $selectorString .= $selectorPartValue;
-            }
-        }
+        return new static($selectorParts);
```
oliverklee, line 151: Do we need to check `selectorParts` to be non-empty here?

src/Property/Selector.php
```diff
@@ -140,52 +146,81 @@ public static function parse(ParserState $parserState, array &$comments = []): s
...
-        $this->selector = !$hasAttribute ? preg_replace('/\\s++/', ' ', $selector) : $selector;
+        $this->components = $components;
     }
 
     /**
      * @return int<0, max>
      */
     public function getSpecificity(): int
     {
-        return SpecificityCalculator::calculate($this->selector);
```
oliverklee, line 192: This now replaces the only usage of the `SpecificityCalculator`.

The `SpecificityCalculator` does more than only creating a sum of specificities. So I have the suspicion that this simplification here actually is breaking, and that `::getSpecificity()` is not covered enough with tests to notice this.

I'd really feel safer if we created more comprehensive tests for `Selector::getSpecificity()` first.

src/Property/Selector.php
JakeQZ: Neither of the type options should be empty for construction of a valid `Selector`.  Have tightened the annotations.

src/Property/Selector.php
JakeQZ: Removed the initialization outside of the constructor.

src/Property/Selector.php
JakeQZ: `parseComponents()` is now declared to return a `non-empty-list` (or throw an exception), and the constructor defined to only accept a `non-empty-list` (or `non-empty-string`).  That should cover it.

src/Property/Selector.php
JakeQZ: We already have the tests:
- https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/tests/Unit/Property/SelectorTest.php#L362
- https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/tests/Unit/Property/SelectorTest.php#L379

With a comprehensive data set:
- https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/tests/Unit/Property/SelectorTest.php#L58

`SpecificityCalculator` is now used by `CompoundSelector`:
- https://github.com/MyIntervals/PHP-CSS-Parser/blob/main/src/Property/Selector/CompoundSelector.php#L226

src/Property/Selector.php
JakeQZ: I've added #1502 as a pre-PR for this.

src/Property/Selector.php
JakeQZ: Also now changed type annoation for `get/setComponents()` to `non-empty-list`.

JakeQZ:
I've tidied up a few things, and reduced the diff with a pre-PR.  I can't see any other possible pre-PRs, so this is ready for re-review.

src/Property/Selector.php
oliverklee: Ah, I seemed to have missed those. Thanks!

src/Property/Selector.php
```diff
@@ -144,57 +148,85 @@ public static function parse(ParserState $parserState, array &$comments = []): s
...
+    {
+        return $this->components;
+    }
+
+    /**
+     * @param non-empty-list<Component> $components
+     *        This should be an alternating sequence of `CompoundSelector` and `Combinator`, starting and ending with a
+     *        `CompoundSelector`, and may be a single `CompoundSelector`.
+     */
+    public function setComponents(array $components): self
```
oliverklee, line 167: Either this methods should return `void` (for consistency), or we should have a test for the fluent interface.

JakeQZ: I've added a test for `setComponents()`.  I didn't make `setSelector()` method fluent: since it's now deprecated, there didn't seem much point.

JakeQZ:
I spotted a few type annotations in `SelectorTest` that could be tightened to use `non-empty-list` or `non-empty-string`, so have done so, and added a fluent interface test.

JakeQZ:
Also now corrected a `@throws` annotation.

oliverklee (approved):
Nice!

PR #1500, opened 2026-02-07
[TASK] Allow construction of 'empty' `Selector`

JakeQZ:
This was already allowed.  This change merely clarifies the behaviour, and makes the constructor parameter optional.

The `parse()` method is static, so it is more convenient if it can create an empty object, then set the value via setter methods.

### DIFF
--- a/tests/Unit/Property/SelectorTest.php
+++ b/tests/Unit/Property/SelectorTest.php
@@ -318,7 +318,17 @@ final class SelectorTest extends TestCase
     /**
      * @test
      */
-    public function canConstructObjectWithEmptyState(): void
+    public function constructsObjectWithEmptyStateWithNoArgumetProvided(): void
+    {
+        $subject = new Selector();
+
+        self::assertSame('', $subject->getSelector());
+    }
+
+    /**
+     * @test
+     */
+    public function constructsObjectWithEmptyStateWithEmptyStringProvided(): void
     {
         $subject = new Selector('');
 


src/Property/Selector.php
```diff
@@ -72,9 +72,12 @@ public static function isValid(string $selector): bool
     /**
      * @throws \UnexpectedValueException if the selector is not valid
      */
-    final public function __construct(string $selector)
+    final public function __construct(string $selector = '')
```
oliverklee, line 75: We should cover creating a selector (and thus having an empty string for a selector) without passing the `$selector` with a test.

oliverklee (changes requested):

src/Property/Selector.php
JakeQZ: That's already been done, in #1498.

oliverklee: I don't see a test that uses `new Selector()` in #1498 - but I'm still at my first coffee. Am I missing something?

JakeQZ: Test method added was `canConstructObjectWithEmptyState()`:

https://github.com/MyIntervals/PHP-CSS-Parser/pull/1498/changes#diff-9faf61b21720e298175bb4d7499b57e7cbf5ff21f2064fa990f2577f4a07510dR318-R327

JakeQZ: I've added an additional test for construction without a parameter.

JakeQZ:
As per https://github.com/MyIntervals/PHP-CSS-Parser/pull/1496#issuecomment-3864024639, this isn't needed.

PR #1446, opened 2025-12-26
[FEATURE] Render selector

8ctopus:
This is a PR draft for selector tidying to transform this:

```css
a>p {
    color: red;
}

a>p[title="hello>world"]>small {
    outline: 0;
}
```

into this:

```css
a > p {
        color: red;
}

a > p[title="hello>world"] > small {
        outline: 0;
}
```

There are no specific tests yet, and there are improvements that can be made, I just want to get your approval before I polish things up.

PR #1504, opened 2026-02-10
[FEATURE] Add `OutputFormat::setSpaceAroundSelectorCombinator()`

JakeQZ:
Resolves #1446

### DIFF
--- a/tests/Unit/OutputFormatTest.php
+++ b/tests/Unit/OutputFormatTest.php
@@ -417,6 +417,33 @@ final class OutputFormatTest extends TestCase
     /**
      * @test
      */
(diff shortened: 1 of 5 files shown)


tests/Unit/OutputFormatTest.php
```diff
@@ -414,6 +414,33 @@ public function setSpaceAfterSelectorSeparatorProvidesFluentInterface(): void
         self::assertSame($this->subject, $this->subject->setSpaceAfterSelectorSeparator(' '));
     }
 
+    /**
+     * @test
+     */
+    public function getSpaceAroundSelectorCombinatorInitiallyReturnsSpace(): void
```
oliverklee: In addition to the default output format, we also should have tests for the compact preset and the pretty preset.

Issue #1445, opened 2025-12-25
inconsistent handling of commented CSS rules


8ctopus:
```php
$input = <<<CSS
a {
    /* margin: 10px 10px 10px 0; */
    outline: none;
}

CSS;

$document = (new \Sabberworm\CSS\Parser($input))
    ->parse();

$format = OutputFormat::createPretty();

echo $document->render($format);
```

becomes:

```css
a {
        /* margin: 10px 10px 10px 0; */
        outline: none;
}
```

while if you update the input to remove the outline property,

```css
a {
    /* margin: 10px 10px 10px 0; */
}
```

 it becomes:

```css
a {}
```

so any commented rule(s) not followed by a CSS rule are removed from the resulting rule set.

JakeQZ:
The current behaviour is sort-of by design, but I agree should be improved.  It assumes that a comment preceding a `Rule` is commenting that `Rule` (and is not a commented-out `Rule`), so it becomes associated with that `Rule`.  Thus, when the `Rule` is removed, the comment that goes with it is also removed.

Looking at the code, I think a comment in the following would also be dropped at the parsing stage:

```css
a {
    /* This has no properties yet. */
}
```

To resolve these issues, I think we would need to modify `RuleSet` so that the list of `Rule`s can be a list of `Rule|Comment`s.  And give the `removeRule()` method and option to keep the comment (which will effectively replace the `Rule` with the `Comment`(s) it had, if any).

oliverklee:
> To resolve these issues, I think we would need to modify `RuleSet` so that the list of `Rule`s can be a list of `Rule|Comment`s. 

I'd prefer we replace the union type with a shared interface. That would allow us to use native type declarations and make it clear in the interface which methods are relevant.

JakeQZ:
> I'd prefer we replace the union type with a shared interface.

I was thinking that.  I just wrote the union as shorthand.

8ctopus:
Honestly, I reported it but don't need it fixed, especially if it adds a lot of complexity to the code.

JakeQZ:
Hmm...

We also need to support this kind of construct:

```css
p {
    /* This is a comment for the declaration */
    font-size: 14px; /* smaller than usual */
    /* This is a comment for the next declaration */
    color: green;
}
```

[...]

PR #1505, opened 2026-02-11
[TASK] Allow whitepace parsing to be stopped at EOL

JakeQZ:
The whitespace parsing includes extraction of comments to be associated with the content.

There are cases where the comment is after the content, but should be associated with the preceding rather than the following. This will help achieve that.

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -201,12 +201,16 @@ class ParserState
      * This method may change the state of the object by advancing the internal position;
      * it does not simply 'get' a value.
      */
-    public function consumeWhiteSpace(array &$comments = []): string
+    public function consumeWhiteSpace(array &$comments = [], bool $stopAfterLineFeed = false): string
     {
         $consumed = '';
         do {
             while (preg_match('/\\s/isSu', $this->peek()) === 1) {
-                $consumed .= $this->consume(1);
+                $characterComsumed = $this->consume(1);
+                $consumed .= $characterComsumed;
+                if ($stopAfterLineFeed && $characterComsumed === "\n") {
+                    break 2;
+                }
             }
             if ($this->parserSettings->usesLenientParsing()) {
                 try {
--- a/tests/Unit/Parsing/ParserStateTest.php
+++ b/tests/Unit/Parsing/ParserStateTest.php
@@ -303,4 +303,67 @@ final class ParserStateTest extends TestCase
 
         self::assertTrue($subject->comes($nonWhitespace));
     }
+
+    /**
+     * @return array<non-empty-string, array{0: string}>
+     */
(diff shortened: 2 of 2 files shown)


oliverklee:
I don't quite understand this yet: Why should consuming whitespace stop at a linefeed, i.e., what makes the linefeed whitespace character special?

JakeQZ:
> I don't quite understand this yet: Why should consuming whitespace stop at a linefeed, i.e., what makes the linefeed whitespace character special?

It's so we would be able to correctly attribute comments to the relevant declarations as 'before' and 'after' (or 'leading' and 'trailing'), e.g.

```css
/* This is a comment about the font size. */
font-size: 14px;
color: green; /* This is a trailing comment about the colour.  It has nothing to do with the width. */
/* This is a comment about the width. */
width: 50%;
```

So after parsing a property declaration, we can consume whitespace and any trailing comments up to the next line feed.  Then we parse the next property declaration, consuming possibly more whitespace and any leading comments.

It's a building block towards solving #663 and #1445.

oliverklee:
> It's so we would be able to correctly attribute comments to the relevant declarations as 'before' and 'after' (or 'leading' and 'trailing'),

I'm not entirely happy with this approach. We should (or could) be able to have line number information for comments. Comments should not "belong" to CSS elements, though - they could be within a declaration, after a declaration, before a declaration, or between declarations.

Particularly, I'd like to avoid introducing of end-of-line comments belong to the declaration before it. This will avoid a can of complexity worms for us, and I currently don't see a usecase for that.

And in general, I'd strongly prefer we drop the concept of comments belonging to a declaration (maybe, just maybe with the exception of a comment that is _within_ a declaration - but even that can get hairy when a declaration is represented as more than one object).

Would you be okay with dropping this concept, @JakeQZ ?

JakeQZ:
> Would you be okay with dropping this concept, @JakeQZ ?

That's fine by me.  It was just an idea I had for solving issues like #663 and #1445, on the basis that currently each declaration owns comments that come before it, but trailing comments become associated with the next declaration, or dropped if there are no more declarations.

So we could instead have `RuleSet` (as it's currently named) represented as something like `array<DeclarationOrComment>` as suggested and discussed in #1445, though there will be some awkwardness as IIRC the declarations are keyed by their property name.

> And in general, I'd strongly prefer we drop the concept of comments belonging to a declaration

The current design has all the various entities implement the `Commentable` interface, so changing this will be much wider-reaching.

> maybe, just maybe with the exception of a comment that is _within_ a declaration

I suspect something like

```css
color: /* My favourite colour */ purple;
```

currently gets rendered as

```css
/* My favourite colour */ 
color: purple;
```

because the placement of the comment is not recorded.

PR #1506, opened 2026-02-12
[TASK] Rename `Rule::$rule` to `Rule::$propertyName`

JakeQZ:
This has become a misnomer.  Rules are the outer enclosing blocks.

Deprecate the existing setters and getters and add new ones.

(The name of the `Rule` class itself is also a misnomer.)

Ref: https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/CSS_Declaration

### DIFF
--- a/src/Rule/Rule.php
+++ b/src/Rule/Rule.php
@@ -75,7 +75,7 @@ class Rule implements Commentable, CSSElement, Positionable
         $parserState->consumeWhiteSpace($comments);
         $rule->setComments($comments);
         $parserState->consume(':');
-        $value = Value::parseValue($parserState, self::listDelimiterForRule($rule->getRule()));
+        $value = Value::parseValue($parserState, self::getDelimitersForPropertyValue($rule->getRule()));
         $rule->setValue($value);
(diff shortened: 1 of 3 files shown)


src/Rule/Rule.php
```diff
@@ -97,17 +97,17 @@ public static function parse(ParserState $parserState, array $commentsBeforeRule
      * The first item is the innermost separator (or, put another way, the highest-precedence operator).
      * The sequence continues to the outermost separator (or lowest-precedence operator).
      *
-     * @param non-empty-string $rule
+     * @param non-empty-string $propertyName
      *
      * @return list<non-empty-string>
      */
-    private static function listDelimiterForRule(string $rule): array
+    private static function listDelimiterForRule(string $propertyName): array
```
oliverklee: Should we also rename the method to `listDelimiterForPropertyName` and update the PHPDoc (if necessary)?

JakeQZ: Good point.  I've renamed the method to hopefully make it clearer what it does.

PR #1508, opened 2026-02-12
[TASK] Rename `Rule\Rule` to `Property\Declaration`

JakeQZ:
A property declaration is not a rule.

Ref: https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/CSS_Declaration

Add a `class_alias` to avoid a breaking change, and deprecate the original class name.

### DIFF
--- a/tests/UnitDeprecated/Rule/RuleTest.php
+++ b/tests/UnitDeprecated/Rule/RuleTest.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Rule;
+
+use PHPUnit\Framework\TestCase;
+use Sabberworm\CSS\Property\Declaration;
+use Sabberworm\CSS\Rule\Rule;
+
+/**
+ * @covers \Sabberworm\CSS\Rule\Rule
+ */
+final class RuleTest extends TestCase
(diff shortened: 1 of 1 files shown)


JakeQZ:
I did `git mv` for the main class file, but had to replace it for the `class_alias`, so GitHub is not showing the diffs as it should :/

src/Rule/Rule.php
```diff
@@ -4,228 +4,11 @@
...
-     * @return array<string, bool|int|float|string|array<mixed>|null>
-     *
-     * @internal
-     */
-    public function getArrayRepresentation(): array
-    {
-        throw new \BadMethodCallException('`getArrayRepresentation` is not yet implemented for `' . self::class . '`');
-    }
-}
+class_alias(Declaration::class, Rule::class);
```
oliverklee, line 14: We also should have a unit test for this alias (in `tests/UnitDeprecated/`).

JakeQZ: Added a test.

JakeQZ:
> … and we also should rename the tests change change them to use the new class. (I'd also be fine with doing that in a follow-up PR).

I was planning to do that in follow-up PRs - we won't be able to remove the deprecated class until that's done.

PR #1514, opened 2026-02-13
[TASK] Use `Property\Declaration` in `DeclarationBlockTest`

JakeQZ:
This replaces `Rule\Rule`, following #1508.

Also use `getPropertyName()` rather than `getRule()` which was deprecated in #1506.

This covers both the 'functional' and 'unit' flavours of the `TestCase`, and the trait that does most of the heavy lifting for the 'unit' flavour.


JakeQZ:
> One line in the commit message got lost, though, as it probably started with `#`, marking it as a comment.

Ah, it was referencing a previous PR, #1506.  Will correct in the merge by including an extra word from the previous line.

PR #1518, opened 2026-02-13
[TASK] Use `Property\Declaration` in generic `DeclarationBlockTest`

JakeQZ:
This replaces `Rule\Rule`, following #1508.


JakeQZ:
Note that this class is neither in `Unit` nor `Functional` test namespaces, and was missed in #1514.

PR #1521, opened 2026-02-14
[TASK] Rename methods in `RuleContainer`

JakeQZ:
E.g. `addRule()` is changed to `addDeclaration()`, with `addRule()` deprecated and changed to a wrapper method to call on to `addDeclaration()`.

`RuleContainer` itself should be renamed to `DeclarationContainer` as a separate change.

The tests will also be updated as a separate change.

### DIFF
--- a/src/RuleSet/LegacyRuleContainerMethods.php
+++ b/src/RuleSet/LegacyRuleContainerMethods.php
@@ -7,7 +7,7 @@ namespace Sabberworm\CSS\RuleSet;
 use Sabberworm\CSS\Property\Declaration;
 
 /**
- * @internal
+ * Provides a mapping of the deprecated methods in a `RuleContainer` to their renamed replacements.
  */
 trait LegacyRuleContainerMethods
 {


JakeQZ:
This could probably be split into a separate PR for each method rename if desired...

Also there are a couple of comments I updated with alterations not entirely related to this PR, so I'll commit those first.

src/RuleSet/LegacyRuleContainerMethods.php
```diff
@@ -0,0 +1,57 @@
...
+namespace Sabberworm\CSS\RuleSet;
+
+use Sabberworm\CSS\Property\Declaration;
+
+/**
+ * @internal
+ */
+trait LegacyRuleContainerMethods
+{
+    public function addRule(Declaration $declarationToAdd, ?Declaration $sibling = null): void
```
oliverklee, line 17: All these methods should get an `@deprecated` annotation stating since which version they are deprecated and in which version they will be removed (like in `RuleContainer`).

JakeQZ: I wondered about this, as I had marked the trait as `@internal`, but the methods are public and implemented in the classes using the trait, and this would be the only place to see the deprecation notice other than in the interface.

I've added the deprecation notices into the trait.

Should the trait nonetheless be marked as `@internal`?

oliverklee: No, the trait should not be `@internal`. The reason is that IDEs mark usages of `@internal` methods and classes (so that the devs know that they should switch to something else), but probably not if only the trait it `@internal`.

JakeQZ: I've removed `@internal` and also added a description for the trait.

PR #1530, opened 2026-02-16
[TASK] Rename `RuleContainer` to `DeclarationList`

JakeQZ:
It contains property declarations not rules, and describing it as a 'list' rather than a 'container' seems more apt, as the declarations are ordered.

Changes to use the new interface name in the source code and tests will be submitted as follow-up PRs.

### DIFF
--- a/config/phpstan.neon
+++ b/config/phpstan.neon
@@ -17,6 +17,7 @@ parameters:
 
   bootstrapFiles:
     - %currentWorkingDirectory%/src/Rule/Rule.php
+    - %currentWorkingDirectory%/src/RuleSet/RuleContainer.php
 
   type_perfect:
     no_mixed_property: true
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@ Please also have a look at our
 
 ### Changed
 
+- `RuleSet\RuleContainer` is renamed to `RuleSet\DeclarationList` (#1530)
 - Methods like `setRule()` in `RuleSet` and `DeclarationBlock` have been renamed
   to `setDeclaration()`, etc. (#1521)
 - `Rule\Rule` class is renamed to `Property\Declaration`
@@ -35,6 +36,8 @@ Please also have a look at our
 
 ### Deprecated
 
+- `RuleSet\RuleContainer` is deprecated; use `RuleSet\DeclarationList` instead
+  (#1530)
 - Methods like `setRule()` in `RuleSet` and `DeclarationBlock` are deprecated;
   there are direct replacements such as `setDeclaration()` (#1521)
 - `Rule\Rule` class is deprecated; `Property\Declaration` is a direct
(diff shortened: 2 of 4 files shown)


JakeQZ:
I forgot to add the `phpstan.neon` config file.  Now added.

If I'd pushed the changes as two separate commits (first rename, then add replacement with `class_alias`) diffing might have been easier.  But I'm not sure about creating a PR from a branch with multiple commits.  Let me know if I should try that next time.

config/phpstan.neon
```diff
@@ -17,6 +17,7 @@ parameters:
 
   bootstrapFiles:
     - %currentWorkingDirectory%/src/Rule/Rule.php
+    - %currentWorkingDirectory%/src/RuleSet/RuleContainer.php
```
oliverklee, line 20: We'll need to replace these additional bootstrap file as soon as possible (in a dedicated PR) with something that also works for consumers of our library. At the moment, [PHPStan fails to find these aliases within Emogrifier](https://github.com/MyIntervals/emogrifier/actions/runs/22039554885/job/63678114608), and other consumers will run into the same problem.

JakeQZ: > We'll need to replace these additional bootstrap file

Oh.  I don't know of a way of doing that.

The aliases allow us to make a minor release using the new class names, with the old class names deprecated.  Then they will be dropped in the next major release.

I expected Stan to find the alias files; when I found it didn't, I discoved this means of telling it.

Is there a way of providing some PHPStan config for consumers?  Or maybe a 'dev' Composer autoload setting could work?

oliverklee: I think we can try these approaches (in this order until one solves the problem):

1. Add a file `autoloading/class-aliases.php` (or something similar), move all class aliases there, and [add it to `autoload.files` in `composer.json`](https://getcomposer.org/doc/04-schema.md#files).
2. Try the [TYPO3 class alias loader](https://github.com/TYPO3/class-alias-loader) (in PHP-CSS-Parser, not in Emogrifier, so we know this works for consumers)

JakeQZ: 🤞

I see there is a branch in Emogrifier that uses the `dev` branch of PHP-CSS-Parser, but not sure how to experimentally hack Composer's `vendor` directory.  Though maybe if there is a way via Composer to avoid the PHPStan bootstrap settings within PHP-CSS-Parser itself, that may solve the problem for consumers.

oliverklee: We can create a PR (or a branch) here in PHP-CSS-Parser and then create a branch in Emogrifier that uses that branch (similar to the PR that uses PHP-CSS-Parser main) and then either run the tests locally or on CI for a PR.

oliverklee: I've just entered #1532 for this and will give it a try.

oliverklee: @JakeQZ This is how I tested this in Emogrifier without having to push to GitHub first: https://github.com/MyIntervals/emogrifier/pull/1570

PR #1534, opened 2026-02-17
[BUGFIX] Make class aliases available for consumers

### LINKED ISSUE #1532: PHPStan does not find the class aliases
For our own CI jobs, PHPStan only finds the class aliases if we define additional bootstrap file.

And for consumers (like Emogrifer), [PHPStan cannot find the class aliases at all](https://github.com/MyIntervals/emogrifier/actions/runs/22039554885/job/63678114608).

I'll give it a go.

oliverklee:
Fixes #1532


oliverklee:
Marking as draft as I'd like to get the cleanup PR #1535 merged first.

class-aliases/aliases.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+use Sabberworm\CSS\Rule\Rule;
+use Sabberworm\CSS\RuleSet\RuleContainer;
+use Sabberworm\CSS\Property\Declaration;
+use Sabberworm\CSS\RuleSet\DeclarationList;
```
JakeQZ: These should be in alphabetical order.

oliverklee:
> Could we not keep the class alias files in their original locations, and just have the two files explicitly loaded via Composer?

I'll give it a try …

oliverklee:
Okay, done and repushed.

oliverklee:
> I assume it still works.

It does indeed (at least on my machine).

JakeQZ:
Relevant PHPStan issues that were not resolved which means we need the Composer workaround:
- https://github.com/phpstan/phpstan/issues/7296
- https://github.com/phpstan/phpstan/issues/8414

JakeQZ:
> > I assume it still works.
> 
> It does indeed (at least on my machine).

I'm now getting errors from Stan saying it can't find the classes, within the PHP-CSS-Parser project alone.  But maybe Composer hasn't updated the class loaders.  When I run `composer update` it says there's nothing to update.

oliverklee:
> I'm now getting errors from Stan saying it can't find the classes,

I got those as well. It turned out I needed to delete the PHPStan caches (which on my Linux machine are located in `/tmp/phstan/`).

JakeQZ:
> > I'm now getting errors from Stan saying it can't find the classes,
> 
> I got those as well. It turned out I needed to delete the PHPStan caches (which on my Linux machine are located in `/tmp/phstan/`).

```cmd
"./vendor/bin/phpstan" clear-result-cache
```

does the job and should work on both Windows and Linux.  I think we have some other cache-clearing commands in `composer.json` that this could be added to...

JakeQZ:
> I think we have some other cache-clearing commands in `composer.json` that this could be added to...

We don't seem to have any cache-clearing commands any more.  Perhaps they were for tools that we stopped using.

PR #1535, opened 2026-02-17
[TASK] Streamline the code checker and fixer configuration

oliverklee:
- always check and fix the `bin/`, `config/`, `src/` and `test/` directories
- sort the directories when provided to a tool
- make the rector command explicit

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -76,10 +76,10 @@
         "check:dynamic": [
             "@check:tests"
         ],
-        "check:php:codesniffer": "phpcs --standard=config/phpcs.xml config src tests",
-        "check:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests config",
-        "check:php:lint": "parallel-lint src tests config bin",
-        "check:php:rector": "rector --no-progress-bar --dry-run --config=config/rector.php",
+        "check:php:codesniffer": "phpcs --standard=config/phpcs.xml bin config src tests",
+        "check:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin config src tests",
+        "check:php:lint": "parallel-lint bin config src tests",
+        "check:php:rector": "rector process --no-progress-bar --dry-run --config=config/rector.php",
         "check:php:stan": "phpstan --no-progress --configuration=config/phpstan.neon",
         "check:static": [
             "@check:composer:normalize",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -76,10 +76,10 @@
...
             "@check:tests"
         ],
-        "check:php:codesniffer": "phpcs --standard=config/phpcs.xml config src tests",
-        "check:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin src tests config",
-        "check:php:lint": "parallel-lint src tests config bin",
-        "check:php:rector": "rector --no-progress-bar --dry-run --config=config/rector.php",
+        "check:php:codesniffer": "phpcs --standard=config/phpcs.xml bin config src tests",
+        "check:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix --dry-run -v --show-progress=dots --diff bin config src tests",
+        "check:php:lint": "parallel-lint bin config src tests",
+        "check:php:rector": "rector process --no-progress-bar --dry-run --config=config/rector.php",
```
JakeQZ, line 82: I'd like to know why `process` is now needed on the command line, when it wasn't before.

I've looked at the [documentation](https://getrector.com/documentation/), which alludes to it with "run `process` command with...", but provides no links, and does not explain what the commands are, how they work, or why they might be needed.

composer.json
```diff
@@ -105,9 +105,9 @@
             "@fix:php:codesniffer",
             "@fix:php:fixer"
         ],
-        "fix:php:codesniffer": "phpcbf --standard=config/phpcs.xml config src tests",
-        "fix:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix bin src tests",
-        "fix:php:rector": "rector --config=config/rector.php",
+        "fix:php:codesniffer": "phpcbf --standard=config/phpcs.xml bin config src tests",
+        "fix:php:fixer": "\"./.phive/php-cs-fixer\" --config=config/php-cs-fixer.php fix bin config src tests",
+        "fix:php:rector": "rector process --config=config/rector.php",
```
JakeQZ, line 110: Ditto.

oliverklee:
> But I need to know why `process` is added to the rector command, and can't find the info.

This is the command for Rector to run. It's the default command (so omitting it will still get Rector to run it), but explicitly providing it makes it easier for the human reader to know what will happen.

oliverklee:
> Where are you finding the documentation? I can't find anything saying that.

That's from my experience: "Rector processes files even when I omit the `process` command." :-)

JakeQZ:
> "Rector processes files even when I omit the `process` command." :-)

Where's that from?

May as well include the `process` command for clarity.

PR #1541, opened 2026-02-20
[TASK] Use `DeclarationList` in the tests

JakeQZ:
The interface was renamed from `RuleContainer` in #1530.

Also accordingly rename the trait that provides test methods for implementing classes.

### DIFF
--- a/tests/Unit/RuleSet/DeclarationBlockTest.php
+++ b/tests/Unit/RuleSet/DeclarationBlockTest.php
@@ -23,7 +23,7 @@ use TRegx\PhpUnit\DataProviders\DataProvider;
  */
 final class DeclarationBlockTest extends TestCase
 {
-    use DeclarationListTest;
+    use DeclarationListTests;
 
     /**
      * @var DeclarationBlock
--- a/tests/Unit/RuleSet/DeclarationListTests.php
+++ b/tests/Unit/RuleSet/DeclarationListTests.php
@@ -16,7 +16,7 @@ use TRegx\PhpUnit\DataProviders\DataProvider;
  *
  * @phpstan-require-extends TestCase
  */
-trait DeclarationListTest
+trait DeclarationListTests
 {
     /**
      * @test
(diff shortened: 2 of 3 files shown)


JakeQZ:
For some reason the CI build can't find the renamed class, but it works fine on my machine.

oliverklee:
While we're renaming the trait anyway, I'd prefer we don't use the `Test` prefix for it. This prefix should only be used for testcase classes, not for traits. (Also, this will allow us to have a `DeclarationListTest` in the future if we want it.)

oliverklee:
> For some reason the CI build can't find the renamed class, but it works fine on my machine.

… which is a Windows machine that uses a case-insensitive file system. :smile:

JakeQZ:
> The casing of the directory is incorrect (`unit` instead of `Unit`). I guess that's hard to fix on Windows. I'll take care of it.

Ah.  Directory names aren't case sensitive in Windows.  I guess lowercase `unit` somehow ended up in the `git mv` part of the commit.

> While we're renaming the trait anyway, I'd prefer we don't use the `Test` prefix for it. This prefix should only be used for testcase classes, not for traits. (Also, this will allow us to have a `DeclarationListTest` in the future if we want it.)

I think you mean suffix.  I'm open to alternative suggestions - `DeclarationListTester`?

oliverklee:
> I think you mean suffix. I'm open to alternative suggestions - `DeclarationListTester`?

Ah, yes, of course suffix.

I‘d be equally fine with `DeclarationListTester` as well as `SharedDeclarationListTests`.

JakeQZ:
> I‘d be equally fine with `DeclarationListTester` as well as `SharedDeclarationListTests`.

Maybe `DeclarationListTests` - it says what it provides via `use` (it's not a tester itself, and that it might be shared between `TestCase`s is perpahps not relevant to the trait itself)...

oliverklee:
Then let‘s go with `DeclarationListTests`.

oliverklee (approved):

PR #1544, opened 2026-02-22
[TASK] Implement `CSSNamespace::getArrayRepresentation()`

oliverklee:
Part of #1440.

### DIFF
--- a/src/Property/CSSNamespace.php
+++ b/src/Property/CSSNamespace.php
@@ -106,7 +106,8 @@ class CSSNamespace implements AtRule, Positionable
     {
         return [
             'class' => $this->getShortClassName(),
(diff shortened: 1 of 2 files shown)


src/Property/CSSNamespace.php
```diff
@@ -102,6 +104,10 @@ public function atRuleArgs(): array
      */
     public function getArrayRepresentation(): array
     {
-        throw new \BadMethodCallException('`getArrayRepresentation` is not yet implemented for `' . self::class . '`');
+        return [
+            'class' => $this->getShortClassName(),
+            'url' => $this->url->getArrayRepresentation(),
```
JakeQZ: Should this be `uri` to match the spec?

PR #1548, opened 2026-02-25
updated README.md with a preloading helper note

dinobot71:


sabberworm:
IMHO this takes up too much space of the readme. I would rather see this as a wiki page on GitHub or in a dedicated FAQ.

JakeQZ:
> IMHO this takes up too much space of the readme. I would rather see this as a wiki page on GitHub or in a dedicated FAQ.

I'm inclined to agree and think that [wiki pages](https://docs.github.com/en/communities/documenting-your-project-with-wikis/adding-or-editing-wiki-pages) are the way forward.  We can also use them for documenting other more detailed information that is beyond the scope of the README.  The README can link to the wiki pages, which presumably will have an index.

I have no experience of setting these up, though the link above provides some information.  I would expect an initial set-up to be required, but can't find any info on that.  It's also not clear if wiki page changes can be submitted as PRs.

oliverklee:
Let's put this on hold until we know whether #1552 could fix the root cause of the issue.

oliverklee:
Not needed anymore now that we have a fix for the root cause.

Still, thanks a bunch for writing this, @dinobot71! :heart:

PR #1549, opened 2026-02-26
Add support for modern CSS at-rules @layer, @scope, and @starting-style

samuil-banti-wpenigne:


samuil-banti-wpenigne:
@oliverklee it seems like there's an issue accessing [coveralls.io](https://coveralls.io/)

oliverklee:
Yes, [Coveralls currently is down](https://status.coveralls.io/).

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -94,4 +109,131 @@ public function parsesSyntacticallyCorrectAtRuleInStrictMode(string $css): void
...
+        $contents = (new Parser($css))->parse()->getContents();
+        $atRuleBlockList = $contents[0];
+
+        self::assertInstanceOf(AtRuleBlockList::class, $atRuleBlockList);
+        self::assertSame('scope', $atRuleBlockList->atRuleName());
+        self::assertSame('(.card)', $atRuleBlockList->atRuleArgs());
+
+        $nestedContents = $atRuleBlockList->getContents();
+        self::assertCount(1, $nestedContents, 'Scope should contain one declaration block');
+    }
```
JakeQZ: Does parsing with a scope limit also work following this change?   And parsing with a selector list?  If so, we should perhaps add tests to confirm this.

We could use a data provider to provide various `@scope` arguments, perhaps lifted from the [MDN examples](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@scope), so we'd only need one `@test` method, which would insert the argument into `$css`, and check it in `assertSame( ... atRuleArgs())`.  This may also work for `@scope` without an argument.

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -94,4 +109,131 @@ public function parsesSyntacticallyCorrectAtRuleInStrictMode(string $css): void
...
+        $contents = (new Parser($css))->parse()->getContents();
+        $atRuleBlockList = $contents[0];
+
+        self::assertInstanceOf(AtRuleBlockList::class, $atRuleBlockList);
+        self::assertSame('layer', $atRuleBlockList->atRuleName());
+        self::assertSame('theme', $atRuleBlockList->atRuleArgs());
+
+        $nestedContents = $atRuleBlockList->getContents();
+        self::assertCount(1, $nestedContents, 'Layer should contain one declaration block');
+    }
```
JakeQZ: Does parsing a statement [`@layer` rule](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@layer) like

```css
@layer theme, layout, utilities;
```

also work following this change?

If it does, we should add tests to confirm.  If not, that will be for a future PR.

src/Property/AtRule.php
```diff
@@ -15,10 +15,12 @@ interface AtRule extends CSSListItem
     /**
      * Since there are more set rules than block rules,
      * we’re whitelisting the block rules and have anything else be treated as a set rule.
+     * NOTE: The new line when assigning the value is added to avoid the line length limit of codesniffer.
```
JakeQZ: This comment belongs in the PR discussion or commit message, not the code.
```suggestion
```

src/Property/AtRule.php
```diff
@@ -15,10 +15,12 @@ interface AtRule extends CSSListItem
...
      *
      * @internal since 8.5.2
      */
-    public const BLOCK_RULES = 'media/document/supports/region-style/font-feature-values/container';
+    public const BLOCK_RULES =
```
JakeQZ, line 21: We need to change this to an array, but that's for another PR.

JakeQZ (changes requested):
This looks excellent, thanks <3

The tests are fairly comprehensive, but I'd like a few more to confirm what does and doesn't work now.

I think the tests for `@scope` parsing would be better handled by using a data provider for the argument, so that we can throw whatever we like at it (I suggest the various examples on the [MDN documentation page](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/At-rules/@scope)).  If some constructs don't work, they can be commented out with a comment in the data provider array indicating that a further fix is needed.

It would also be good to know if `@layer` statement rules are now also parsed and rendered correctly following this change (as well as `@layer` block rules, which are covered in the tests added).

It might be helpful to split this PR into three manageable chunks for each at-rule.  But if it's a hassle to do that, it's nae bother.

If you rebase, the Coveralls build failure will vanish, following #1551.

tests/CSSList/AtRuleBlockListTest.php
samuil-banti-wpenigne: Nope, it doesn't work currently :(
The parser discards `@layer theme, layout, utilities;` and returns an empty contents array and renders nothing.

src/Property/AtRule.php
samuil-banti-wpenigne: This seems like a quick win. Are you sure you want it in a separate PR?

tests/CSSList/AtRuleBlockListTest.php
```diff
@@ -94,4 +109,100 @@ public function parsesSyntacticallyCorrectAtRuleInStrictMode(string $css): void
...
+                'layer',
+                '',
+                1,
+            ],
+            'scope with selector' => [
+                '@scope (.card) { .title { font-size: 2rem; } }',
+                'scope',
+                '(.card)',
+                1,
+            ],
```
JakeQZ, line 139: Would it be possible to have a test for scope with a limt, e.g. `@scope (.article-body) to (figure)`?  Or does that not currently parse correctly?

src/Property/AtRule.php
JakeQZ: This still needs to be removed.

src/Property/AtRule.php
JakeQZ: > This seems like a quick win. Are you sure you want it in a separate PR?

Having PRs only addressing one thing makes it easier to backport them or temporarily undo them if a problem is found.  So, yes.

tests/CSSList/AtRuleBlockListTest.php
JakeQZ: > Nope, it doesn't work currently :(

I feared that would be the case.  There are statement at-rules, and block at-rules, and `@layer` can be either.  Some congnitive effort and refactoring will likely be required to support both flavours of `@layer` (as a future PR).

tests/CSSList/AtRuleBlockListTest.php
samuil-banti-wpenigne: It parses correctly.

src/Property/AtRule.php
samuil-banti-wpenigne: I was just waiting for confirmation that BLOCK_RULES will not be converted to array in this PR.

JakeQZ (approved):
Super.  Thanks for contributing again <3

src/Property/AtRule.php
JakeQZ: #1556 for this.

PR #1550, opened 2026-02-26
[TASK] Allow PHPStan to use unlimited processes

oliverklee:
This helps speed up local development and CI.


oliverklee:
> I had noticed the PHPStan step taking a short while after making code changes, but does this really speed it up?

Yes, on my machine it does. (I have a fat desktop machine with 16 cores, though.) :muscle:

PR #1552, opened 2026-02-27
[BUGFIX] Avoid double autoloading of class aliases

### LINKED ISSUE #1546: PHP-CSS-Parser Incompatible with OPCache / Composer Autoload
Hi, after several hours of trying to resolve this issue by cleaning/re-installing various parts of my system (and obviously reinstalling vendor/) I am unable to get this package working within a PHP-FPM/Apache backend.  It consistently generates a fatal exception reproducible in the same place every time.  It appears to be not be working propertly with autoload and OPCache.  

When I use the package in a CLI environment (no OPCache involved)...there is no issue.

The exception I'm seeing:

```
[BOOT] Starting PHP-FPM (Background)...


[Preloader] init...
[Preloader] loading...
[Preloader] Preloaded 2151 classes
[Preloader] complete.
[25-Feb-2026 04:50:23] NOTICE: fpm is running, pid 42
[25-Feb-2026 04:50:23] NOTICE: ready to handle connections

[BOOT] php-fpm is RUNNING.

[BOOT] Starting Apache (Foreground)...
...
...
- -  25/Feb/2026:04:50:27 +0000 "GET /backend.php" 200 //var/www/html/backend.php 481.192 2048 4.16%

NOTICE: PHP message: Exception [/var/www/html/vendor/thecodingmachine/safe/generated/Exceptions/ClassobjException.php: 9]: Cannot redeclare class Sabberworm\CSS\Rule\Rule (previously declared in /var/www/html/vendor/sabberworm/php-css-parser/src/Property/Declaration.php:27)

Traceback: 
#0 /var/www/html/vendor/thecodingmachine/safe/generated/8.1/classobj.php(19): Safe\Exceptions\ClassobjException::createFromPhpError()
#1 /var/www/html/vendor/sabberworm/php-css-parser/src/Rule/Rule.php(14): Safe\class_alias('Sabberworm\\CSS\\...', 'Sabberworm\\CSS\\...')
#2 /var/www/html/vendor/composer/autoload_real.php(41): require('/var/www/html/v...')
#3 /var/www/html/vendor/composer/autoload_real.php(45): {closure:ComposerAutoloaderInita513f45a21088ff88ff9a0ececb2b3ac::getLoader():37}('0174385c3be07e8...', '/var/www/html/v...')
#4 /var/www/html/vendor/autoload.php(22): ComposerAutoloaderInita513f45a21088ff88ff9a0ececb2b3ac::getLoader()
#5 /var/www/html/backend.php(21): require('/var/www/html/v...')

#6 {main}

```
The problem seems to be this class and its declarations...the way its working with thecodingmachine/safe package.

```php-css-parser/src/Rule```

Unfortunately I can not simply disable OPCache, I have to continue using it.  So my only option would seem to be to wrap the usage of this package (via dompdf...I'm generated pdf files), into a CLI script that I would literally exec() from the PHP backend endpoint.

That seems to me to be a terrible solution...this package should not break and "poison" everything else.  Its not breaking when I use it, it breaking when my backend loads up the standard autoload file.  

In fact, if I remove a few lines from the autoload files (fgrep sabb in autoload/composer):

```
autoload_files.php:    '0174385c3be07e86008907d06ee66531' => $vendorDir . '/sabberworm/php-css-parser/src/Rule/Rule.php',
autoload_files.php:    '98aea6e41b9cb79b379b10f37ba1f0b7' => $vendorDir . '/sabberworm/php-css-parser/src/RuleSet/RuleContainer.php',
autoload_psr4.php:    'Sabberworm\\CSS\\' => array($vendorDir . '/sabberworm/php-css-parser/src'),
autoload_static.php:        '0174385c3be07e86008907d06ee66531' => __DIR__ . '/..' . '/sabberworm/php-css-parser/src/Rule/Rule.php',
autoload_static.php:        '98aea6e41b9cb79b379b10f37ba1f0b7' => __DIR__ . '/..' . '/sabberworm/php-css-parser/src/RuleSet/RuleContainer.php',
autoload_static.php:            0 => __DIR__ . '/..' . '/sabberworm/php-css-parser/src',
installed.json:                "sabberworm/php-css-parser": "^8.4 || ^9.0"
installed.json:            "name": "sabberworm/php-css-parser",
installed.json:            "homepage": "https://www.sabberworm.com/blog/2010/6/10/php-css-parser",
installed.json:            "install-path": "../sabberworm/php-css-parser"
installed.php:        'sabberworm/php-css-parser' => array(
installed.php:            'install_path' => __DIR__ . '/../sabberworm/php-css-parser',
```
the problem goes away...If I remove those 4 Rule and RuleSet files from the composer autoload files...the problem goes away.  
```
autoload_files.php:    '0174385c3be07e86008907d06ee66531' => $vendorDir . '/sabberworm/php-css-parser/src/Rule/Rule.php',
autoload_files.php:    '98aea6e41b9cb79b379b10f37ba1f0b7' => $vendorDir . '/sabberworm/php-css-parser/src/RuleSet/RuleContainer.php',
autoload_static.php:        '0174385c3be07e86008907d06ee66531' => __DIR__ . '/..' . '/sabberworm/php-css-parser/src/Rule/Rule.php',
autoload_static.php:        '98aea6e41b9cb79b379b10f37ba1f0b7' => __DIR__ . '/..' . '/sabberworm/php-css-parser/src/RuleSet/RuleContainer.php',
```

I didn't do any kind of weird installation here, just standard composer setup, and then "composer install" with an empty vendor folder.

Somehow, this package and/or its interaction with thecodingmachine/safe stuff...is injecting those 4 files into the autoload files...which in turn is causing "double loads" of the files, and then PHP barfs with a critical exception...which kills my backend :( 

I would really like to continue using dompdf, but having to walk on eggshells and wrap it into external CLI script is terrible hack in my view. Not to mention all the performance I'll be losing everytime is runs externally and has to reload every single file, making the endpoint for my PDFs that much slower. :(

The other hack here is to just edit the autoload files per above, but again, this seems like a terrible hack.  Having to hack files out of an automated step of composer...for one package is not really sustainable. How do I know what other classes will be an issue? But, at least performance is still reasonably good.

Is there any better work around available?  I'm using the latest stable stuff, I don't have weird versions of anything installed.  As I say, when I run my PDF generation outside of the Apache/PHP-FPM/OPCache environment...everything works complete fine...and also of course, when I edit the composer autoload files.

Note also that my preload file does NOT preload either this packge or the thecodingmachine stuff.  Its being loaded directly by the composer autoload files.    There is some kind of interaction here that is poisoning the autoload system.   All you need to reproduce this is:

```require dirname(__FILE__) . '/vendor/autoload.php';```

For my composer file, the requirements I have:

```
  "require": {
    "bshaffer/oauth2-server-php":           "~1.8",
    "bshaffer/oauth2-server-httpfoundation-bridge": "v1.2",
    "lcobucci/jwt":                         "*",
    "firebase/php-jwt":                     "*",
    "monolog/monolog":                      "*",
    "nesbot/carbon":                        "*",
    "sabre/xml":                            "*",
    "guzzlehttp/guzzle":                    "*",
    "symfony/yaml":                         "*",
    "symfony/psr-http-message-bridge":      "*",
    "symfony/process":                      "*",
    "symfony/mailer":                       "*",
    "symfony/mailgun-mailer":               "*",
    "symfony/sendgrid-mailer":              "*",
    "symfony/google-mailer":                "*",
    "league/route":                         ">=6.2.0",
    "laminas/laminas-diactoros":            "*",
    "laminas/laminas-httphandlerrunner":    "*",
    "league/container":                     "*",
    "peppeocchi/php-cron-scheduler":        "*",
    "php":                                  ">=8.4",
    "opis/closure":                         ">=4.3.1",
    "erusev/parsedown":                     "*",
    "illuminate/database":                  "^12.10",
    "illuminate/events":                    "^12.10",
    "illuminate/bus":                       "^12.10",
    "illuminate/collections":               "^12.10",
    "illuminate/broadcasting":              "^12.10",
    "illuminate/support":                   "^12.10",
    "illuminate/contracts":                 "^12.10",
    "illuminate/queue":                     "^12.10",
    "illuminate/container":                 "^12.10",
    "ramsey/uuid":                          "4.7.5",
    "studio24/rotate":                      "*",
    "ext-json":                             "*",
    "ext-zlib":                             "*",
    "ext-pdo":                              "*",
    "ext-sqlite3":                          "*",
    "phpfastcache/phpfastcache":            "*",
    "ext-curl":                             "*",
    "salsify/json-streaming-parser":        "*",
    "dompdf/dompdf":                        "*",
    "dompdf/php-svg-lib":                   "*",
    "dompdf/php-font-lib":                  "*",
    "twig/twig":                            "*",
    "psr/log":                              "*",
    "ext-sockets":                          "*",
    "ext-zip":                              "*",
    "ext-gd":                               "*",
    "ext-imagick":                          "*",
    "ext-pcntl":                            "*",
    "ext-pgsql":                            "*",
    "ext-odbc":                             "*",
    "ext-calendar":                         "*",
    "ext-iconv":                            "*",
    "ext-fileinfo":                         "*",
    "ext-xml":                              "*",
    "ext-dom":                              "*"
  },
```
The other main clauses that maybe I have wrong in the composer file:

```
 "config": {
    "allow-plugins": {
      "kylekatarnls/update-helper": true
    }
  },
  "require-dev": {
    "phpunit/phpunit": "^12"
  }
```

Please let me know if a better work around is available.  Dompdf seems to work really well, but I need it to work both for CLI and in a web server backend.

### LINKED ISSUE #1557: Rule class is undefined, which does not let my symfony applicatino start
https://github.com/MyIntervals/PHP-CSS-Parser/blob/dfdf1834ca91d549206697220caf82b88c6bf338/src/Rule/Rule.php#L14

https://github.com/MyIntervals/PHP-CSS-Parser/blob/dfdf1834ca91d549206697220caf82b88c6bf338/src/Rule/Rule.php#L14

Hi, I installed the sabberworm/php-css-parser as required package of dompdf, but after the upgrade from 9.1.0 to 9.2.0 my symfony application does not start, because the Rule class is nowhere declared. Can you please fix it?

Thank you in advance, greetings,
Giulio

### PR DESCRIPTION
**Pull request by oliverklee**
Fixes #1546
Fixes #1557

### DIFF
--- a/src/Rule/Rule.php
+++ b/src/Rule/Rule.php
@@ -8,7 +8,7 @@ use Sabberworm\CSS\Property\Declaration;
 
 use function Safe\class_alias;
 
-if (!\class_exists(Rule::class) && !\interface_exists(Rule::class)) {
+if (!\class_exists(Rule::class, false) && !\interface_exists(Rule::class, false)) {
     /**
      * @deprecated in v9.2, will be removed in v10.0.  Use `Property\Declaration` instead, which is a direct
      *             replacement.
--- a/src/RuleSet/RuleContainer.php
+++ b/src/RuleSet/RuleContainer.php
@@ -6,7 +6,7 @@ namespace Sabberworm\CSS\RuleSet;
 
 use function Safe\class_alias;
 
-if (!\class_exists(RuleContainer::class) && !\interface_exists(RuleContainer::class)) {
+if (!\class_exists(RuleContainer::class, false) && !\interface_exists(RuleContainer::class, false)) {
     /**
      * @deprecated in v9.2, will be removed in v10.0.  Use `DeclarationList` instead, which is a direct replacement.
      */

### CODE REVIEW

**Comment by JakeQZ** [comment]
If this doesn't work, an alternative would be to wrap the `class_alias()` calls in a `class_exists()` check, e.g.

```php
if (!\class_exists(Rule::class, false)) {
    /**
     * @deprecated in v9.2, will be removed in v10.0.  Use `Property\Declaration` instead, which is a direct replacement.
     */
    class_alias(Declaration::class, Rule::class);
}
```

I think that should work, and I'd actually prefer it:
- It avoids introducing an additional root-level directory along with the associated configuration/command changes;
- We can keep the aliasing files in the location of the old classes -
  - If PHPStan is updated to support recognizing class aliases without the files being specifically preloaded, we can then simply remove the `autoload.files` section from `composer.json` (the only reason the files are being preloaded is to avoid PHPStan warnings in consumer projects - #1534).

oliverklee:
I've added the wrappers now.

JakeQZ:
> I've added the wrappers now.

With the wrappers added, would it be possible to keep the `class_alias` files in their original location, so that we don't have to introduce an additional root-level directory and change the commands in `composer.json`?

class-aliases/Rule/Rule.php
```diff
@@ -0,0 +1,17 @@
...
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Rule;
+
+use Sabberworm\CSS\Property\Declaration;
+
+use function Safe\class_alias;
+
+if (!\class_exists(Rule::class) && !\interface_exists(Rule::class)) {
```
JakeQZ: I think we should pass `false` as the second argument to `class_exists()` and `interface_exists()` to avoid any unwanted autoloading.

class-aliases/RuleSet/RuleContainer.php
```diff
@@ -0,0 +1,14 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\RuleSet;
+
+use function Safe\class_alias;
+
+if (!\class_exists(RuleContainer::class) && !\interface_exists(RuleContainer::class)) {
```
JakeQZ: I think we should pass `false` as the second argument to `class_exists()` and `interface_exists()` to avoid any unwanted autoloading.

JakeQZ (changes requested):
Moving the files to a `class-aliases` directory does not help, as reported in https://github.com/MyIntervals/PHP-CSS-Parser/issues/1546#issuecomment-3978677467:

> Moving the references from 'src' to 'class-aliases' is just changing where the problem happens.

I think we should keep the files in their original location, to avoid introducing a new root-level directory and potential confusion.  I'm hopeful that the `class_exists()` test will solve the problem.

JakeQZ:
We should probably make a 9.2.1 release with this fix, since there have been several reports of the issue since 9.2.0 which was only released a few days ago.

oliverklee:
As we've added a new feature, this would then be a 9.3.0 release.

PR #1559, opened 2026-03-03
[TASK] Make aliased class deprecation notices visible to tools

### LINKED ISSUE #1553: Mark class aliases as deprecated in a different way
Our current way of marking the class aliases as deprecated [doesn't get registered with PHPStan](https://github.com/MyIntervals/emogrifier/pull/1579)  (and also not with my IDE). So we'll need a different approach.

https://www.reddit.com/r/PHP/comments/51riss/using_class_alias_to_maintain_bc_while/

JakeQZ:
Partly resolves #1553.

### DIFF
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -25,22 +25,10 @@ parameters:
 			path: ../src/Parsing/ParserState.php
 
 		-
-			message: '#^Negated boolean expression is always true\.$#'
-			identifier: booleanNot.alwaysTrue
-			count: 1
-			path: ../src/Rule/Rule.php
-
-		-
 			message: '#^Parameters should have "string" types as the only types passed to this method$#'
 			identifier: typePerfect.narrowPublicClassMethodParamType
 			count: 1
 			path: ../src/RuleSet/DeclarationBlock.php
-
-		-
-			message: '#^Negated boolean expression is always true\.$#'
-			identifier: booleanNot.alwaysTrue
-			count: 1
-			path: ../src/RuleSet/RuleContainer.php
 
 		-
 			message: '#^Parameter \#2 \$arguments of class Sabberworm\\CSS\\Value\\CSSFunction constructor expects array\<Sabberworm\\CSS\\Value\\Value\|string\>\|Sabberworm\\CSS\\Value\\RuleValueList, Sabberworm\\CSS\\Value\\Value\|string given\.$#'
--- a/config/phpstan-baseline.neon
+++ b/config/phpstan-baseline.neon
@@ -25,10 +25,22 @@ parameters:
 			path: ../src/Parsing/ParserState.php
 
 		-
+			message: '#^Negated boolean expression is always true\.$#'
+			identifier: booleanNot.alwaysTrue
+			count: 1
+			path: ../src/Rule/Rule.php
+
+		-
 			message: '#^Parameters should have "string" types as the only types passed to this method$#'
 			identifier: typePerfect.narrowPublicClassMethodParamType
 			count: 1
 			path: ../src/RuleSet/DeclarationBlock.php
+
+		-
+			message: '#^Negated boolean expression is always true\.$#'
+			identifier: booleanNot.alwaysTrue
+			count: 1
+			path: ../src/RuleSet/RuleContainer.php
 
 		-
 			message: '#^Parameter \#2 \$arguments of class Sabberworm\\CSS\\Value\\CSSFunction constructor expects array\<Sabberworm\\CSS\\Value\\Value\|string\>\|Sabberworm\\CSS\\Value\\RuleValueList, Sabberworm\\CSS\\Value\\Value\|string given\.$#'
--- a/src/Rule/Rule.php
+++ b/src/Rule/Rule.php
@@ -9,9 +9,12 @@ use Sabberworm\CSS\Property\Declaration;
 use function Safe\class_alias;
 
 if (!\class_exists(Rule::class, false) && !\interface_exists(Rule::class, false)) {
-    /**
-     * @deprecated in v9.2, will be removed in v10.0.  Use `Property\Declaration` instead, which is a direct
-     *             replacement.
-     */
     class_alias(Declaration::class, Rule::class);
+    if (!\class_exists(Rule::class, false)) {
+        /**
+         * @deprecated in v9.2, will be removed in v10.0.  Use `Property\Declaration` instead, which is a direct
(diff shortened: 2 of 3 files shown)


src/Rule/Rule.php
```diff
@@ -9,9 +9,12 @@
 use function Safe\class_alias;
 
 if (!\class_exists(Rule::class, false) && !\interface_exists(Rule::class, false)) {
-    /**
-     * @deprecated in v9.2, will be removed in v10.0.  Use `Property\Declaration` instead, which is a direct
-     *             replacement.
-     */
     class_alias(Declaration::class, Rule::class);
+    if (!\class_exists(Rule::class, false)) {
```
oliverklee: Just to be safe, we should do the full class & interface check as well.

config/phpstan-baseline.neon
```diff
@@ -24,12 +24,24 @@ parameters:
...
+			path: ../src/Rule/Rule.php
+
 		-
 			message: '#^Parameters should have "string" types as the only types passed to this method$#'
 			identifier: typePerfect.narrowPublicClassMethodParamType
 			count: 1
 			path: ../src/RuleSet/DeclarationBlock.php
 
+		-
+			message: '#^Negated boolean expression is always true\.$#'
```
oliverklee: These should be [PHPStan annotations](https://phpstan.org/user-guide/ignoring-errors) in the code with the error code and a short explanation why we do it this way. This will allow us to keep the baseline as a to-do list of thing we'd like to fix.

oliverklee:
I've given this a go: PhpStorm correctly marks the `Rule` class as deprecated both in PHP-CSS-Parser as well as Emogrifier.

PHPStan doesn't pick it up in my case either. So this seems to be a dead end.

oliverklee:
… or we keep this for the IDE support.

JakeQZ:
> … or we keep this for the IDE support.

If it fixes the issue for IDEs, it at least gets us some of the way (and probably the most important part of the way).  We can try to figure out how to fix it also for Stan as a future PR.

So on that basis, I'd say we go ahead with this change.  I'll put in some `@phstan` annotations...

src/Rule/Rule.php
JakeQZ: Done.

config/phpstan-baseline.neon
JakeQZ: I've added the annotations, but they only seem to work to silence the `booleanAnd.alwaysTrue` warnings (introduced by having the `class_exists` and `interface_exists` combined), and seem to have no effect on the `booleanNot.alwaysTrue` warning.

It looks like this may be a bug in PHPStan, and we may have to use `@phpstan-ignore-next-line` for the time being.

JakeQZ: Changed to use `@phpstan-ignore-next-line`, since `@phpstan-ignore` is not working.  If you can figure out why, it would be nice to only exclude certain things, but I can't, and I think it's a bug in PHPStan.

src/Rule/Rule.php
```diff
@@ -9,9 +9,15 @@
 use function Safe\class_alias;
 
 if (!\class_exists(Rule::class, false) && !\interface_exists(Rule::class, false)) {
-    /**
-     * @deprecated in v9.2, will be removed in v10.0.  Use `Property\Declaration` instead, which is a direct
-     *             replacement.
-     */
     class_alias(Declaration::class, Rule::class);
+    // The test is expected to evuluate to false,
```
oliverklee: Thur sooms to boo a typoo: 😉 

```suggestion
    // The test is expected to evaluate to false,
```

src/RuleSet/RuleContainer.php
```diff
@@ -7,8 +7,14 @@
 use function Safe\class_alias;
 
 if (!\class_exists(RuleContainer::class, false) && !\interface_exists(RuleContainer::class, false)) {
-    /**
-     * @deprecated in v9.2, will be removed in v10.0.  Use `DeclarationList` instead, which is a direct replacement.
-     */
     class_alias(DeclarationList::class, RuleContainer::class);
+    // The test is expected to evuluate to false,
```
oliverklee:
```suggestion
    // The test is expected to evaluate to false,
```

src/Rule/Rule.php
```diff
@@ -9,9 +9,15 @@
...
+    // but allows for the deprecation notice to be picked up by IDEs like PHPStorm.
+    // @phpstan-ignore-next-line
```
oliverklee: We can narrow this down to only ignore that one specific type of warning so PHPStan can warn us if anything else is amiss:

```suggestion
    // @phpstan-ignore-next-line booleanAnd.alwaysTrue
```

src/RuleSet/RuleContainer.php
```diff
@@ -7,8 +7,14 @@
...
+    // but allows for the deprecation notice to be picked up by IDEs like PHPStorm.
+    // @phpstan-ignore-next-line
```
oliverklee:
```suggestion
    // @phpstan-ignore-next-line booleanAnd.alwaysTrue
```

oliverklee (changes requested):

src/Rule/Rule.php
JakeQZ: Fixed.

src/RuleSet/RuleContainer.php
JakeQZ: Fixed.

src/Rule/Rule.php
JakeQZ: Added.

src/RuleSet/RuleContainer.php
JakeQZ: Added.

oliverklee (approved):

src/Rule/Rule.php
JakeQZ: Actually this doesn't do anything; `@phpstan-ignore-next-line` does not have parameters.  #1560 tidies this up.

config/phpstan-baseline.neon
JakeQZ: > I've added the annotations, but they only seem to work to silence the `booleanAnd.alwaysTrue` warnings (introduced by having the `class_exists` and `interface_exists` combined), and seem to have no effect on the `booleanNot.alwaysTrue` warning.

I finally realized that putting `booleanNot.alwaysTrue` was silencing one of the warnings, but not both of them.  Having it repeated silences both warnings.

PR #1564, opened 2026-04-11
[TASK] Implement `Declaration::getArrayRepresentation()`

oliverklee:
Part of #1440.

### DIFF
--- a/src/Property/Declaration.php
+++ b/src/Property/Declaration.php
@@ -14,6 +14,7 @@ use Sabberworm\CSS\Parsing\UnexpectedEOFException;
 use Sabberworm\CSS\Parsing\UnexpectedTokenException;
 use Sabberworm\CSS\Position\Position;
 use Sabberworm\CSS\Position\Positionable;
+use Sabberworm\CSS\ShortClassNameProvider;
 use Sabberworm\CSS\Value\RuleValueList;
 use Sabberworm\CSS\Value\Value;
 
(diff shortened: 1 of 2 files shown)


src/Property/Declaration.php
```diff
@@ -226,6 +228,13 @@ public function render(OutputFormat $outputFormat): string
...
     public function getArrayRepresentation(): array
     {
-        throw new \BadMethodCallException('`getArrayRepresentation` is not yet implemented for `' . self::class . '`');
+        return [
+            'class' => $this->getShortClassName(),
+            'propertyName' => $this->propertyName,
+            // We're using the term "property value" here to match the wording used in the specs:
+            // https://www.w3.org/TR/CSS22/syndata.html#declaration
+            'propertyValue' => $this->value,
+            'important' => $this->isImportant,
```
JakeQZ, line 237: I take all the Git Blame for this possible slight misnomer in the class.  I lifted the terminology from [MDN](https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/CSS_Declaration) which currently still refers to the CSSOM properties for a 'CSS Declaration' as "property name" and "value".  Though it does refer to `isImportant` as "important flag".

I think we are close enough to the mark to avoid confusion, and the official nomenclature may change in future W3C revisions anyway.

JakeQZ (approved):

PR #1568, opened 2026-04-16
[CLEANUP] Move PHPUnit configuration to `Build/`

AnReiter:


.gitattributes
oliverklee: We need to drop this line.

oliverklee (changes requested):

oliverklee:
This is failing as the CI job runs `./vendor/bin/phpunit` instead of the Composer script. I propose we get a PR in first that changes the CI job to use the Composer script.

oliverklee (approved):

PR #1575, opened 2026-04-19
[TASK] Disable some rules from Rector Type Perfect

oliverklee:
The two type-narrowing rules tend to have too many false positives at the moment.

### DIFF
--- a/Build/phpstan/phpstan-baseline.neon
+++ b/Build/phpstan/phpstan-baseline.neon
@@ -7,28 +7,10 @@ parameters:
 			path: ../../src/CSSList/CSSList.php
 
 		-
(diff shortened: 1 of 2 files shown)


JakeQZ (approved):
We should revisit this in future and reenable type narrowing.  But in the meantime, there are issues with it:
- It does not seem to play well with union types -
  - We should redesign the API to avoid union types, perhaps introducing classes to wrap the type;
- The warning messages are rather cryptic, identifying the callee as the source of the problem, rather than the caller, making them sometimes almost impossible to diagnose -
  - This needs improvement in TypePerfect or PHPStan or whatever is obfuscating the details.

PR #1576, opened 2026-04-19
[CLEANUP] Drop leftover `.gitkeep`

oliverklee:


JakeQZ (approved):
This was put in place when setting up the directory structure.  Now there is some content, it is no longer required.  And when that content is migrated to non-deprecated unit tests, the directory will now magically disappear.

PR #1577, opened 2026-04-19
[TASK] Add tests for `RuleValueList::getArrayRepresentation()`

oliverklee:
Part of #1440.

### DIFF
--- a/tests/Unit/Value/Fixtures/ConcreteRuleValueList.php
+++ b/tests/Unit/Value/Fixtures/ConcreteRuleValueList.php
@@ -1,9 +1,0 @@
-<?php
-
-declare(strict_types=1);
-
-namespace Sabberworm\CSS\Tests\Unit\Value\Fixtures;
-
-use Sabberworm\CSS\Value\RuleValueList;
-
-final class ConcreteRuleValueList extends RuleValueList {}
--- a/tests/Unit/Value/Fixtures/ConcreteRuleValueList.php
+++ b/tests/Unit/Value/Fixtures/ConcreteRuleValueList.php
@@ -1,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit\Value\Fixtures;
+
+use Sabberworm\CSS\Value\RuleValueList;
+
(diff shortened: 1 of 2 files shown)


tests/Unit/Value/Fixtures/ConcreteRuleValueList.php
```diff
@@ -0,0 +1,9 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\Unit\Value\Fixtures;
+
+use Sabberworm\CSS\Value\RuleValueList;
+
+final class ConcreteRuleValueList extends RuleValueList {}
```
JakeQZ: `RuleValueList` is already a concrete class.  This is not necessary.

JakeQZ (changes requested):
We don't need `ConcreteRuleValueList` as `RuleValueList` is already a concrete class.

tests/Unit/Value/Fixtures/ConcreteRuleValueList.php
oliverklee: Oh, indeed. It seems I got confused by the class also getting subclassed. Fixed and repushed.

tests/Unit/Value/RuleValueListTest.php
```diff
@@ -25,4 +26,60 @@ public function getArrayRepresentationIncludesClassName(): void
...
+        self::assertSame('+', $result['components'][1]['value']);
+        self::assertSame('Size', $result['components'][2]['class']);
+    }
+
+    /**
+     * @test
+     */
+    public function getArrayRepresentationIncludesSeparator(): void
+    {
+        $separator = ', ';
```
JakeQZ, line 78: We should probably add a test that the default separator is returned if none is provided to the constructor.  And in `ValueListTest` also check that a non-default separator is returned.  But that can be for another PR.

JakeQZ (approved):
The implementation was put in to the abstract class `ValueList`, and tests were added for that class.

These new tests are for the concrete subclass `RuleValueList`, and pretty much the same, to confirm that nothing is broken by the subclassing.

PR #1580, opened 2026-04-26
[TASK] Implement `(At)RuleSet::getArrayRepresentation()`

oliverklee:
Part of #1440.

### DIFF
--- a/src/RuleSet/RuleSet.php
+++ b/src/RuleSet/RuleSet.php
@@ -14,6 +14,7 @@ use Sabberworm\CSS\Parsing\UnexpectedTokenException;
 use Sabberworm\CSS\Position\Position;
 use Sabberworm\CSS\Position\Positionable;
 use Sabberworm\CSS\Property\Declaration;
+use Sabberworm\CSS\ShortClassNameProvider;
 
 /**
  * This class is a container for individual `Declaration`s.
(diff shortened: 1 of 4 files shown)


src/RuleSet/RuleSet.php
```diff
@@ -344,7 +346,20 @@ static function () use ($declaration, $nextLevelFormat): string {
...
-        throw new \BadMethodCallException('`getArrayRepresentation` is not yet implemented for `' . self::class . '`');
+        $declarationsArrayRepresentation = [];
+        foreach ($this->declarations as $propertyName => $declarationForOneProperty) {
+            $declarationsArrayRepresentation[$propertyName] = \array_map(
+                function (Declaration $declaration): array {
+                    return $declaration->getArrayRepresentation();
+                },
+                $declarationForOneProperty
+            );
+        }
```
JakeQZ: I think the loop iterator value variable should be plural for clarity:

```suggestion
        foreach ($this->declarations as $propertyName => $declarationsForOneProperty) {
            $declarationsArrayRepresentation[$propertyName] = \array_map(
                function (Declaration $declaration): array {
                    return $declaration->getArrayRepresentation();
                },
                $declarationsForOneProperty
            );
        }
```

JakeQZ (changes requested):
I got a bit confused because there can be multiple values defined for a single property, e.g.

```css
font-size: 16px;
font-size: 1rem; /* for browsers that support `rem` units */
```

So I think the loop variable for the property value should be pluralized.  Otherwise looks fine.  (We already have tests for `Declaration::getArrayRepresentation()`, so don't need to provide property values in the tests for `(At)RuleSet`.)

I assume there will be a future PR to add the at-rule name to the array representation.

oliverklee:
> I assume there will be a future PR to add the at-rule name to the array representation.

I've added that now, together with the arguments.

JakeQZ (approved):

PR #1582, opened 2026-05-03
[TASK] Test `LineName::getArrayRepresentation()`

oliverklee:
The tests are basically the same as those for the parent class `ValueList`. The main difference is that the default separator now is a space, not a comma.

Part of #1440.

### DIFF
--- a/tests/Unit/Value/LineNameTest.php
+++ b/tests/Unit/Value/LineNameTest.php
@@ -6,6 +6,7 @@ namespace Sabberworm\CSS\Tests\Unit\Value;
 
 use PHPUnit\Framework\TestCase;
 use Sabberworm\CSS\Value\LineName;
+use Sabberworm\CSS\Value\Size;
 
 /**
  * @covers \Sabberworm\CSS\Value\LineName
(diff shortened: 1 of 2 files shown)


tests/Unit/Value/LineNameTest.php
```diff
@@ -17,12 +18,62 @@ final class LineNameTest extends TestCase
...
+
+        self::assertSame('LineName', $result['class']);
+    }
+
+    /**
+     * @test
+     */
+    public function getArrayRepresentationIncludesStringComponent(): void
+    {
+        $subject = new LineName(['Helvetica']);
```
JakeQZ: Would prefer not to have a font name, which is confusing.  It should be some made-up idendifier like `main-start`.

tests/Unit/Value/LineNameTest.php
```diff
@@ -17,12 +18,62 @@ final class LineNameTest extends TestCase
...
+    public function getArrayRepresentationIncludesMultipleMixedComponents(): void
+    {
+        $subject = new LineName([new Size(1), '+', new Size(2)]);
+
+        $result = $subject->getArrayRepresentation();
+
+        self::assertSame('Size', $result['components'][0]['class']);
+        self::assertSame('+', $result['components'][1]['value']);
+        self::assertSame('Size', $result['components'][2]['class']);
+    }
```
JakeQZ: These tests are invalid, because `LineName` should not accept `Size` in the argument list.  Instead we should have a test that two string identifiers are represented (as well as that for a single string identifier which isn't a font name).

JakeQZ (changes requested):
I've never used line names (aka [named grid lines](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Grid_layout/Named_grid_lines)).

It seems that they are user-defined identifiers (i.e. strings), and that the `LineName` constructor should actually only allow strings in the `$components` parameter, not `Value`s (for another PR).

In the meantime, we should be consistent with what the class is representing: a non-empty list of user-defined string identifiers, which are whitespace-separated in the rendered CSS.

oliverklee:
> I've never used line names (aka [named grid lines](https://developer.mozilla.org/en-US/docs/Web/CSS/Guides/Grid_layout/Named_grid_lines)).

I've added this to the Class PHPDoc in #1588.

> It seems that they are user-defined identifiers (i.e. strings), and that the `LineName` constructor should actually only allow strings in the `$components` parameter, not `Value`s (for another PR).

I've entered #1589 for this.

tests/Unit/Value/LineNameTest.php
```diff
@@ -17,12 +17,53 @@ final class LineNameTest extends TestCase
...
+
+        $result = $subject->getArrayRepresentation();
+
+        self::assertSame($name, $result['components'][0]['value']);
+    }
+
+    /**
+     * @test
+     */
+    public function getArrayRepresentationCanIncludesMultipleStringComponents(): void
```
JakeQZ:
```suggestion
    public function getArrayRepresentationCanIncludeMultipleStringComponents(): void
```

JakeQZ (changes requested):
There's a grammar mistake in a method name - otherwise looks good.

JakeQZ (approved):

PR #1583, opened 2026-05-13
[TASK] Do not treat PHPDoc type information as certain

oliverklee:
This is a library. This means that we have no control over how other software calls our methods, and whether they adher to the types required via `@param` annotations.

Hence, we shouldn't rely on those types being certain.

### DIFF
--- a/Build/phpstan/phpstan-baseline.neon
+++ b/Build/phpstan/phpstan-baseline.neon
@@ -49,12 +49,6 @@ parameters:
 			path: ../../src/Value/Value.php
 
 		-
-			message: '#^Call to static method PHPUnit\\Framework\\Assert\:\:assertSame\(\) with ''red'' and Sabberworm\\CSS\\Value\\Value will always evaluate to false\.$#'
-			identifier: staticMethod.impossibleType
-			count: 1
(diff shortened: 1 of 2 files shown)


JakeQZ (approved):

Build/phpstan/phpstan-baseline.neon
```diff
@@ -48,12 +48,6 @@ parameters:
 			count: 1
 			path: ../../src/Value/Value.php
 
-		-
-			message: '#^Call to static method PHPUnit\\Framework\\Assert\:\:assertSame\(\) with ''red'' and Sabberworm\\CSS\\Value\\Value will always evaluate to false\.$#'
-			identifier: staticMethod.impossibleType
-			count: 1
-			path: ../../tests/ParserTest.php
-
```
JakeQZ, line 56: That this warning existed indicates something is returning a string but has DocBlock type indicating a different return type.

JakeQZ: #1584 resolves this concern.  Turns out the test code had a loop with no iterations, within which was the `assertSame()` which would always fail, but was never called.

PR #1591, opened 2026-05-20
[BUGFIX] Use `@phpstan-ignore` not `@phpstan-ignore-next-line`

JakeQZ:
`@phpstan-ignore` has the parameter which is the specific warning to ignore. `@phpstan-ignore-next-line` does not, and will result in suppression of all warnings for the next line.
`@phpstan-ignore` also only applies to the next line (or whatever is relevent in context).


oliverklee (approved):
🧹

PR #1592, opened 2026-06-15
[TASK] Move `ParserTest` to `Unit` namespace

JakeQZ:
The test methods in this class are unit tests, testing only the `Parser` class.

### DIFF
--- a/tests/Unit/ParserTest.php
+++ b/tests/Unit/ParserTest.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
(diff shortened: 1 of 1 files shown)


JakeQZ:
Actually, `Parser::parse()` just calls on to `Document::parse()`, so I don't think this change is valid.  Aborting.

oliverklee:
Yeah, I vaguely remember discussing this a long time ago and then concluding that the parser tests need to be functional or integration as the parser class makes heavy use of almost all other classes.

PR #1595, opened 2026-06-17
[TASK] Temporarily drop the `composer normalize` CI job

oliverklee:
Currently, we keep getting errors about problems with GitHub authentication, which breaks this job. We'll need to sort this out first before we re-activate the job.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -103,7 +103,6 @@ jobs:
             fail-fast: false
             matrix:
                 command:
-                    - composer:normalize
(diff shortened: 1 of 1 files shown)


oliverklee:
Putting on hold for now - maybe the update in #1566 already fixes the issue.

JakeQZ:
> Putting on hold for now - maybe the update in #1566 already fixes the issue.

Doesn't seem to be an issue with the CI build for #1566.  Though the problem appeared to be intermittent.

PR #1609, opened 2026-08-12
[TASK] Update to Rector 2.6.2

oliverklee:
https://github.com/rectorphp/rector/releases/tag/2.6.2


JakeQZ (commented):
Can the version increment be done as a separate pre-PR, then the config changes?

JakeQZ (approved):
The `PHPUnitSetList` class is removed in 2.6.2, so the config changes have to be tied in with the version increment.

PR #1610, opened 2026-08-12
[TASK] Update PHPStan

oliverklee:


JakeQZ (approved):
Following #1626, this can be merged.  Since there are no controversial changes, I've proceeded to do so.

PR #1612, opened 2026-08-14
[TASK] Remove `thecodingmachine/safe` dependency (part 1)

SjorsO:
As discussed here: https://github.com/MyIntervals/PHP-CSS-Parser/pull/1484#issuecomment-5241402914

`preg_match` can return `false` if:
- The pattern is invalid
- the `u` flag is used and the subject or pattern aren't valid UTF-8
- Backtrack/recursion limits are hit

This `preg_match` can't fail because:
- The pattern is always valid (`$match` is always a fixed value)
- There is no `u` flag
- There is no backtracking or recursion

### DIFF
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -251,7 +251,7 @@ abstract class CSSList implements CSSElement, CSSListItem, Positionable
         }
 
         $matchResult = \preg_match("/^(-\\w+-)?$match$/i", $identifier);
-        \assert($matchResult !== false);
+        \assert(\is_int($matchResult));
         return $matchResult === 1;
     }
 
--- a/src/CSSList/CSSList.php
+++ b/src/CSSList/CSSList.php
@@ -25,8 +25,6 @@ use Sabberworm\CSS\Value\CSSString;
 use Sabberworm\CSS\Value\URL;
 use Sabberworm\CSS\Value\Value;
 
-use function Safe\preg_match;
-
 /**
  * This is the most generic container available. It can contain `DeclarationBlock`s (rule sets with a selector),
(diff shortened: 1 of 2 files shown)


src/CSSList/CSSList.php
```diff
@@ -252,7 +250,9 @@ private static function identifierIs(string $identifier, string $match): bool
             return true;
         }
 
-        return preg_match("/^(-\\w+-)?$match$/i", $identifier) === 1;
+        $matchResult = \preg_match("/^(-\\w+-)?$match$/i", $identifier);
+        \assert($matchResult !== false);
```
oliverklee: For make this easier to read, let's stat what it is instead of what it isn't:


```suggestion
        \assert(\is_int($matchResult));
```

oliverklee (changes requested):

oliverklee (approved):

JakeQZ:
> This `preg_match` can't fail because:
> 
> * There is no backtracking or recursion

Actually there is backtracking, or at least the storage of backtracking information.  The engine is simple and does not know what it might be matching next.  So when matching the optional vendor-prefix, it needs to be able to rewind in case the following part will match without it but not with.  We can see that's not possible, but it can't.  So it will build up a rewind stack.  If presented with something like `-veryverylongvendorprefix-translate` (shortened for readability), it could bail.

Use of the [possessive quantifier](https://www.regular-expressions.info/possessive.html) will mitigate this.  I think that's beyond the scope of these PRs, so have added #1614.

The library as it is should bail out somehow reasonably in such unrealistic situations, as well as it did before the introduction of 'Safe', and at least there is now an `assert()`.

PR #1615, opened 2026-08-15
[TASK] Remove thecodingmachine/safe dependency (part 2)

SjorsO:
- Giving `false` to `Parser` already throws an exception (string typehint + strict types)
- The `opendir()` call opens the fixtures directory, that should always exist. If it somehow fails, the tests should already fail too

### DIFF
--- a/bin/quickdump.php
+++ b/bin/quickdump.php
@@ -5,15 +5,16 @@ declare(strict_types=1);
 
 use Sabberworm\CSS\Parser;
 
-use function Safe\file_get_contents;
-
 /**
  * This script is used for generating the examples in the README.
  */
 
 require_once(__DIR__ . '/../vendor/autoload.php');
 
-$source = file_get_contents('php://stdin');
+$source = \file_get_contents('php://stdin');
+if ($source === false) {
+    throw new \RuntimeException('Failed to read from stdin.');
+}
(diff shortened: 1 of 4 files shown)


bin/quickdump.php
```diff
@@ -5,15 +5,16 @@
...
-
 /**
  * This script is used for generating the examples in the README.
  */
 
 require_once(__DIR__ . '/../vendor/autoload.php');
 
-$source = file_get_contents('php://stdin');
+$source = \file_get_contents('php://stdin');
+if ($source === false) {
```
oliverklee: Let's check for the expected data type instead:

```suggestion
if (!\is_string($source)) {
```

bin/quickdump.php
```diff
@@ -5,15 +5,16 @@
...
+    throw new \RuntimeException('Failed to read from stdin.');
```
oliverklee: All thrown exceptions should get an exception code (usually the current timestamp). This will allow searching the source code for the exception code to quickly find the line that throws the exception:

```suggestion
    throw new \RuntimeException('Failed to read from stdin.', 1786954465);
```

oliverklee (changes requested):

oliverklee (approved):

PR #1616, opened 2026-08-19
[TASK] Remove thecodingmachine/safe dependency (part 3)

SjorsO:
All calls in `ParserState` that can return `false` now throw an exception instead, each with its own unit test that checks for the unique exception code.

### DIFF
--- a/src/Parsing/ParserState.php
+++ b/src/Parsing/ParserState.php
@@ -7,10 +7,6 @@ namespace Sabberworm\CSS\Parsing;
 use Sabberworm\CSS\Comment\Comment;
 use Sabberworm\CSS\Settings;
 
-use function Safe\iconv;
-use function Safe\preg_match;
-use function Safe\preg_split;
-
 /**
  * @internal since 8.7.0
  */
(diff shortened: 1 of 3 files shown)


src/Parsing/ParserState.php
```diff
@@ -164,7 +172,15 @@ public function parseCharacter(bool $isForIdentifier): ?string
                 $utf32EncodedCharacter .= \chr($codePoint & 0xff);
                 $codePoint = $codePoint >> 8;
             }
-            return iconv('utf-32le', $this->charset, $utf32EncodedCharacter);
+            // The suppression is needed because `\iconv` emits a notice, as well as returning `false`, on failure.
+            $convertedCharacter = @\iconv('utf-32le', $this->charset, $utf32EncodedCharacter);
```
SjorsO, line 176: Suppressing this notice seems like a smart move. We're already handling the error, no reason to emit a notice too

PHPUnit fails if a test emits a notice, so we either have to handle this with a `@` or add a small error handling hack in the unit test

tests/Unit/Parsing/ParserStateTest.php
```diff
@@ -15,6 +15,77 @@
...
+     */
+    public function constructorThrowsForTextThatIsNotValidUtf8(): void
+    {
+        if (\PHP_VERSION_ID < 70304) {
+            // https://bugs.php.net/76127
+            self::markTestSkipped('Before PHP 7.3.4 preg_split did not return false for invalid UTF-8');
+        }
+
+        $this->expectException(\RuntimeException::class);
+        $this->expectExceptionCode(1787112623);
```
oliverklee, line 30: Please also always test for the exception message as well using `$this->expectExceptionMessage()`.

oliverklee (changes requested):
Great tests! I particularly like the test names, and I also love that we now have tests for the error cases.

I've added a comment on making the tests a bit tighter.

oliverklee (approved):
Thanks!

PR #1617, opened 2026-08-21
[TASK] Remove thecodingmachine/safe dependency (part 4)

SjorsO:
- This PR tackles all four files in the `Property` directory
- Two of them can fail on invalid UTF-8, both of these have tests
- The others don't use the `/u` modifier, and don't do backtracking, so they can't fail. These got an `assert()`

### DIFF
--- a/src/Property/Selector.php
+++ b/src/Property/Selector.php
@@ -15,8 +15,6 @@ use Sabberworm\CSS\Renderable;
 use Sabberworm\CSS\Settings;
 use Sabberworm\CSS\ShortClassNameProvider;
 
-use function Safe\preg_match;
-
 /**
  * Class representing a single CSS selector. Selectors have to be split by the comma prior to being passed into this
  * class.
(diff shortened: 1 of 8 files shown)


src/Property/Selector.php
```diff
@@ -68,7 +66,10 @@ class Selector implements Renderable
     public static function isValid(string $selector): bool
     {
         // Note: We need to use `static::` here as the constant is overridden in the `KeyframeSelector` class.
-        $numberOfMatches = preg_match(static::SELECTOR_VALIDATION_RX, $selector);
+        $numberOfMatches = \preg_match(static::SELECTOR_VALIDATION_RX, $selector);
```
SjorsO, line 69: This pattern is overridden in `KeyframeSelector`, so I've added a test for this in both `Selector` and `KeyframeSelector`. Both patterns fail the same way on invalid UTF-8

oliverklee (approved):

PR #1619, opened 2026-08-22
[FEATURE] Add support for PHP 8.6

oliverklee:


oliverklee:
This probably requires the PHPStan update in #1610.

JakeQZ:
> This probably requires the PHPStan update in #1610.

That's now resolved.

Seems we're both trying to rebase this at the same time.  Hope I haven't trampled on your branch.

The unit tests are still failing with PHP 8.6...

JakeQZ:
> The unit tests are still failing with PHP 8.6...

The error is from https://github.com/enojaroff/HOA-Consistency/blob/master/Source/Prelude.php#L54, in a dependency of PHPStan that appears to be no longer maintained.

I don't know how soon it is likely to be fixed and then make its way into PHPStan, if ever.  I've logged https://github.com/phpstan/phpstan/issues/15127.

If we want to move forward with PHP 8.6 sooner, we could temporarily put a `try`...`catch` block in the test.  Though I don't know what it should be `catch`ing, and don't have any plans to install PHP 8.6 locally to find out.  The test would still run as intended on PHP<=8.5.

JakeQZ:
#1627 should help, following review, and then rebase of this.

JakeQZ (approved):

PR #1621, opened 2026-08-24
[TASK] Remove thecodingmachine/safe dependency (part 7)

SjorsO:
This is the last code change. Next PR can remove the package.

The preg_match in `CSSString` uses `/u` and can fail, this has a unit test. All others can't fail.


oliverklee (approved):
Thanks!

PR #1623, opened 2026-08-25

oliverklee:


JakeQZ (approved):
The `validate-config` command line was added in [Rector 2.6.7](https://github.com/rectorphp/rector/releases/tag/2.6.7).

The older versions of this and some of the other tools won't actually work, but are permitted so that `composer install` will complete with older PHP versions so that the unit tests can be run.

PR #1624, opened 2026-08-27
Add support for layer declarations ending with semicolon

samuil-banti-wpenigne:
Semicolon-terminated `@layer` is valid CSS but was parsed as a block, so later rules were swallowed.

`@layer` has two forms ([CSS Cascade 5](https://www.w3.org/TR/css-cascade-5/#layer-empty)):

```
@layer reset;
@layer base, components;
@layer theme { .button { color: blue; } }
```

Generic at-rule parsing always searched for {. For `@layer reset;`, that meant consuming through the next { (e.g. `@property`), so the rest of the stylesheet was treated as nested contents of a fake layer block.

Fix: stop at `{` or `;`. A `;` becomes `AtRuleStatement;` a `{` still becomes `AtRuleBlockList` or `AtRuleSet`. layer stays in `BLOCK_RULES`, so `@layer theme { … }` is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ Please also have a look at our
 
 ### Added
 
+- Add tests for semicolon-terminated `@layer` statements // TODO: add PR ID
+
 ### Changed
 
 ### Deprecated
@@ -19,6 +21,9 @@ Please also have a look at our
 - Remove the `thecodingmachine/safe` dependency (#1622)
 
 ### Fixed
+
+- Parse semicolon-terminated `@layer` statements (and other statement at-rules)
+  instead of consuming until the next `{` // TODO: add PR ID
 
 ### Documentation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Please also have a look at our
 
 ### Added
 
-- Add tests for semicolon-terminated `@layer` statements // TODO: add PR ID
+- Add tests for semicolon-terminated `@layer` statements (#1624)
 
 ### Changed
 
@@ -23,7 +23,7 @@ Please also have a look at our
 ### Fixed
 
 - Parse semicolon-terminated `@layer` statements (and other statement at-rules)
-  instead of consuming until the next `{` // TODO: add PR ID
+  instead of consuming until the next `{` (#1624)
 
 ### Documentation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,6 @@ Please also have a look at our
 
 ### Added
 
-- Add tests for semicolon-terminated `@layer` statements (#1624)
-
 ### Changed
 
 ### Deprecated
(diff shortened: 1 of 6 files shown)


tests/CSSList/AtRuleStatementTest.php
```diff
@@ -0,0 +1,133 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Sabberworm\CSS\Tests\CSSList;
```
JakeQZ: I think this should be in the namespace `Sabberworm\CSS\Tests\Functional\Property`, with the file location changed accordingly.

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@ Please also have a look at our
 
 ### Added
 
+- Add tests for semicolon-terminated `@layer` statements (#1624)
+
```
JakeQZ:
```suggestion
```

This isn't needed as it's developer-specific.  The main entry in the 'Fixed' section is sufficient.

JakeQZ (changes requested):
Wow.  Excellent contribution @samuil-banti-wpenigne.  Hard to pick any holes in that.

The functional tests for the new class should be in a different location (and namespace), and I don't think we need a changelog entry that tests have been added.

Otherwise, looks perfect.  Thank you so much, particularly for taking the time to understand the codebase.

@oliverklee, do you have any other nitpicks?

src/CSSList/CSSList.php
```diff
@@ -210,8 +211,13 @@ private static function parseAtRule(ParserState $parserState): ?CSSListItem
             }
             return new CSSNamespace($url, $prefix, $identifierLineNumber);
         } else {
-            // Unknown other at rule (font-face or such)
-            $arguments = \trim($parserState->consumeUntil('{', false, true));
+            // Unknown other at rule (font-face, @layer, or such)
+            $arguments = \trim($parserState->consumeUntil(['{', ';'], false, false));
```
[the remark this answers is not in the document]
JakeQZ: It's the same with `{`.  Fixing it is beyond the scope of this PR.  #1625 added.

JakeQZ (approved):
LGTM now.  One of the best PR submissions we've ever had.  Thank you.

@oliverklee, anything else to pick up on?

`phar.io` server is currently down, causing failing checks...

oliverklee:
@JakeQZ I most probably won't have time to look at this today or tomorrow, and I don't want to block this. So feel free to merge in the meantime.

JakeQZ:
> @JakeQZ I most probably won't have time to look at this today or tomorrow, and I don't want to block this. So feel free to merge in the meantime.

OK.  Happy to do so so that people can pick up the `-dev` branch.  The code change looks solid, the new class located with its siblings (which may all be relocated at some point), and the tests good.

The only thing I can think of that might be falling short is whether the tests cover every eventuality and format of at-rules.  But they never will - there's always an edge case that has been missed.  AI was surprisingly smart at picking up the pre-existing lack of string parsing, which has been logged as #1625.  Any other issues found can be resolved as a follow-up.

Thanks again @samuil-banti-wpenigne for your excellent contribution.

Issue #1628, opened 2026-09-04
Harden our GitHub workflows


oliverklee:
https://phpunit.expert/articles/hardening-github-actions-workflows.html
https://blog.packagist.com/securing-our-github-actions-workflows-with-zizmor/
https://docs.zizmor.sh/
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#permissions

PR #1631, opened 2026-09-05
[TASK] Pin the GH action releases

oliverklee:
https://docs.zizmor.sh/audits/#unpinned-uses

Part of #1628


JakeQZ (commented):
As for the sister PR, can the `uses` not be pinned by a specific version tag (e.g. `@v7.0.1`)?  This would make it clearer and less error-prone than using the SHA, simpler to maintain, and allow Dependabot to update the tools (unless it can handle maintaining pins by SHA).

JakeQZ (approved):

PR #1633, opened 2026-09-05
[TASK] Set concurrency limits for the GH Action workflows

oliverklee:
https://docs.zizmor.sh/audits/#concurrency-limits

Part of #1628


JakeQZ (commented):
I think we might need to include the [`github.workflow`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) property in the `group`, otherwise won't the CI workflow cancel the Code Coverage workflow or vice versa?

oliverklee:
> I think we might need to include the [`github.workflow`](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/control-workflow-concurrency#example-only-cancel-in-progress-jobs-or-runs-for-the-current-workflow) property in the `group`, otherwise won't the CI workflow cancel the Code Coverage workflow or vice versa?

No, due to the different prefixes: https://github.com/MyIntervals/emogrifier/pull/1655#issuecomment-5557834524

JakeQZ (approved):

